PR #104, opened 2022-08-12
add scatter gather configs

lmbollen:
This PR goes on top of #66
 
This PR adds GADTs to configure the scatter- and gather units. 
These GADTs hide the component specific type variables from the rest of the implementation.

### DIFF
--- a/bittide/src/Bittide/ScatterGather.hs
+++ b/bittide/src/Bittide/ScatterGather.hs
@@ -22,12 +22,16 @@ import Bittide.Calendar
 import Bittide.DoubleBufferedRam
 import Bittide.SharedTypes
 
+-- | GADT to explicitly differentiate between a configuration for the 'scatterUnitWb' and
+-- 'gatherUnitWb' at type level.
 data ScatterConfig nBytes addrW where
   ScatterConfig ::
     (KnownNat memDepth, 1 <= memDepth) =>
     (CalendarConfig nBytes addrW (Index memDepth)) ->
     ScatterConfig nBytes addrW
 
+-- | GADT to explicitly differentiate between a configuration for the 'scatterUnitWb' and
+-- 'gatherUnitWb' at type level.
 data GatherConfig nBytes addrW where
   GatherConfig ::
     (KnownNat memDepth, 1 <= memDepth) =>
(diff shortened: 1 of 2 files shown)


bittide/src/Bittide/ScatterGather.hs
```diff
@@ -31,7 +47,7 @@ scatterUnit ::
   , KnownNat memDepth, 1 <= memDepth
   , KnownNat frameWidth) =>
   -- | Configuration for the 'calendar'.
-  CalendarConfig nBytes addrW (CalendarEntry memDepth) ->
+  CalendarConfig nBytes addrW (Index memDepth) ->
```
leonschoorl, line 50: Shouldn't this take a `ScatterConfig`?

bittide/src/Bittide/ScatterGather.hs
```diff
@@ -59,7 +75,7 @@ gatherUnit ::
   , KnownNat frameWidth, 1 <= frameWidth
   , KnownNat (DivRU frameWidth 8), 1 <= (DivRU frameWidth 8)) =>
   -- | Configuration for the 'calendar'.
-  CalendarConfig nBytes addrW (CalendarEntry memDepth) ->
+  CalendarConfig nBytes addrW (Index memDepth) ->
```
leonschoorl, line 78: Shouldn't this take the `GatherConfig` instead?

leonschoorl (approved):
Otherwise LGTM

bittide/src/Bittide/ScatterGather.hs
lmbollen: The `GADT`s are used to hide implementations details where they are not necessary and be able to differentiate  between the S/G unit calendars.
The differentiation would be nice to have, but `memDepth` is present in the API of the `scatterUnit` and gatherUnit`. I could change it, but I'd have to add a run-time check that verifies whether the tyVars are the same.

bittide/src/Bittide/ScatterGather.hs
```diff
@@ -22,12 +22,16 @@ import Bittide.Calendar
 import Bittide.DoubleBufferedRam
 import Bittide.SharedTypes
 
+-- | GADT to explicitly differentiate between a configuration for the 'scatterUnitWb' and
```
leonschoorl, line 25: This is actually a GADT, but an existential type, that just happens to be defined using [GADTSyntax](https://downloads.haskell.org/ghc/9.2.4/docs/html/users_guide/exts/gadt_syntax.html#extension-GADTSyntax)

leonschoorl: But you can probably leave those details out, if what you care about documenting here is that they allow you differentiate between scatter and gather config.
The existential part is what allows you to hide the `memDepth`.

bittide/src/Bittide/ScatterGather.hs
leonschoorl: But you're right, not having access to the `memDepth` inside the `ScatterConfig` would be problematic here

PR #139, opened 2022-09-20
Optionally use the Xilinx FIFO in simulation

vmchale:
This adds a little wrapper, since the proper FIFO starts at 0 - we wait until it is half-full before clock control does anything. 

This closes https://github.com/google-research/bittide/issues/124


martijnbastiaan:
Could you rebase on the proper branches? I don't think this reflects your current proposal :)

vmchale:
Yes! Will do.

Issue #125, opened 2022-09-12
Add boot stage 1: fill elastic buffers, stabilize elastic buffers


martijnbastiaan:
Related, now closed issue: https://github.com/google-research/bittide/issues/42.

As we've seen in simulations, buffer occupancies can decrease/increase arbitrarily while waiting for a system to become stable. For example, a system with "many" nodes has produced the following plot in the past:

![Screenshot from 2022-09-12 21-22-10](https://user-images.githubusercontent.com/5658854/189738122-31b49c99-f63b-4990-85d4-f57bc12e9921.png)

The problem is that larger systems would need larger FIFOs in order to function. This is unacceptable, as Bittide systems can grow arbitrarily large.

To solve this, a single node (with many elastic buffers) should follow roughly to following steps:

* Track mode:
  1. Collect all "stable" and "{under,over}flow" signals from the elastic buffers.
  2. If under/overflow: reset all elastic buffers and **clock control state**. Return to (1).
  3. If not all stable: return to (1).
 
* Stable mode:
  1. Monitor {under,over}flow. Export sticky version of these signals.

This algorithm works, as long as the individual steps always move closer to the ideal situation. Note that, while in boot mode any restrictions to the clock modifier board do not apply.

lmbollen:
I created #217 as concept solution for this issue

martijnbastiaan:
This has been fixed by [Sanjay's reframing approach](https://github.com/bittide/bittide-hardware/pull/279) combined with ["infinite" counters](https://github.com/bittide/bittide-hardware/pull/289).

PR #141, opened 2022-09-21
Add boot mode to clock control, enabling use of smaller FIFOs

vmchale:
Part of https://github.com/bittide/bittide-hardware/issues/125

It still works! See:

[clockscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9616871/clockscomplete3.pdf)
[elasticbufferscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9616872/elasticbufferscomplete3.pdf)

### DIFF
--- a/elastic-buffer-sim/src/Bittide/Simulate.hs
+++ b/elastic-buffer-sim/src/Bittide/Simulate.hs
@@ -99,7 +99,9 @@ ebController size clkRead clkWrite = go <$> rdEna <*> outRd
   go True (dc, False) = Just dc
   go _ _ = Nothing
 
-  rdEna = register clkRead resetGen enableGen True $ mealy clkRead resetGen enableGen f False outRd
+  rdEna =
+    register clkRead resetGen enableGen True
+      $ mealy clkRead resetGen enableGen f False outRd
    where
     f :: Bool -> (DataCount, Underflow) -> (Bool, Bool)
     f False (_, False) = (False, True)
@@ -107,7 +109,9 @@ ebController size clkRead clkWrite = go <$> rdEna <*> outRd
     f True (d, _) | d == targetDataCount size = (False, True)
     f _ _ = (False, False)
 
-  wrEna = register clkWrite resetGen enableGen True $ mealy clkWrite resetGen enableGen g False outWr
+  wrEna =
+    register clkWrite resetGen enableGen True
+      $ mealy clkWrite resetGen enableGen g False outWr
    where
     g :: Bool -> (DataCount, Overflow) -> (Bool, Bool)
     g False (_, False) = (False, True)
@@ -144,24 +148,26 @@ elasticBuffer size clkRead clkWrite readEna writeEna
       then goRead relativeTime fillLevel rps wps
       else goWrite relativeTime fillLevel rps wps
 
-  goWrite relativeTime fillLevel rps (writePeriod :- wps) rdEna (wrEna :- wrEnas) | wrEna =
+  goWrite relativeTime fillLevel rps (writePeriod :- wps) rdEna (wrEna :- wrEnas)
+    | wrEna =
     second (next :-) $
       go (relativeTime - toInteger writePeriod) newFillLevel rps wps rdEna wrEnas
    where
     next@(newFillLevel, _)
-      | fillLevel >= size = (size, True)
+      | fillLevel >= size = (targetDataCount size, True)
       | otherwise = (fillLevel + 1, False)
 
   goWrite relativeTime fillLevel rps (writePeriod :- wps) rdEna (_ :- wrEnas) =
     second ((fillLevel, False) :-) $
       go (relativeTime - toInteger writePeriod) fillLevel rps wps rdEna wrEnas
 
-  goRead relativeTime fillLevel (readPeriod :- rps) wps (rdEna :- rdEnas) wrEnas | rdEna =
+  goRead relativeTime fillLevel (readPeriod :- rps) wps (rdEna :- rdEnas) wrEnas
+    | rdEna =
     first (next :-) $
       go (relativeTime + toInteger readPeriod) newFillLevel rps wps rdEnas wrEnas
    where
     next@(newFillLevel, _)
-      | fillLevel <= 0 = (0, True)
+      | fillLevel <= 0 = (targetDataCount size, True)
       | otherwise = (fillLevel - 1, False)
 
   goRead relativeTime fillLevel (readPeriod :- rps) wps (_ :- rdEnas) wrEnas =
--- a/elastic-buffer-sim/src/Bittide/Simulate.hs
+++ b/elastic-buffer-sim/src/Bittide/Simulate.hs
@@ -85,6 +85,12 @@ tunableClockGen settlePeriod periodOffset stepSize _reset speedChange =
     in
       newPeriod :- go newSettleCounter newPeriod scs
 
+-- | This wrapper disables reads or writes when the elastic buffer
+-- over-/underflows, as appropriate.
+--
+-- It also censors 'DataCount' after an over-/underflow, so that we do not take
+-- measurements from an elastic buffer until it has settled back to its
+-- midpoint.
 ebController ::
   forall readDom writeDom.
   (KnownDomain readDom, KnownDomain writeDom) =>
@@ -92,14 +98,18 @@ ebController ::
   Clock readDom ->
   Clock writeDom ->
   Signal readDom (Maybe DataCount)
-ebController size clkRead clkWrite = go <$> rdEna <*> outRd
+ebController size clkRead clkWrite =
+  go <$> rdToggle <*> outRd <*> overflowRd
  where
-  (outRd, outWr) = elasticBuffer size clkRead clkWrite rdEna wrEna
+  (outRd, outWr) = elasticBuffer size clkRead clkWrite rdToggle wrToggle
 
-  go True (dc, False) = Just dc
-  go _ _ = Nothing
+  go True (dc, False) False = Just dc
+  go _ _ _ = Nothing
 
-  rdEna =
+  overflowRd =
+    dualFlipFlopSynchronizer clkWrite clkRead resetGen enableGen False (snd <$> outWr)
+
+  rdToggle =
     register clkRead resetGen enableGen True
       $ mealy clkRead resetGen enableGen f False outRd
    where
@@ -109,7 +119,7 @@ ebController size clkRead clkWrite = go <$> rdEna <*> outRd
     f True (d, _) | d == targetDataCount size = (False, True)
     f _ _ = (False, False)
 
-  wrEna =
+  wrToggle =
     register clkWrite resetGen enableGen True
       $ mealy clkWrite resetGen enableGen g False outWr
    where
--- a/elastic-buffer-sim/src/Bittide/Simulate.hs
+++ b/elastic-buffer-sim/src/Bittide/Simulate.hs
@@ -9,6 +9,7 @@ Provides a rudimentary simulation of elastic buffers.
 -- SPDX-License-Identifier: Apache-2.0
 
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 
 module Bittide.Simulate where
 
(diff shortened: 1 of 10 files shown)


martijnbastiaan (changes requested):
This doesn't match my expectations: the clock periods shouldn't radically jump for reasons mentioned [here](https://github.com/bittide/bittide-hardware/issues/125#issuecomment-1251246067). I see two very suspicious trends:

* If there are three nodes A B C and A resets its buffers, it shouldn't have any effect on B C. We do see this happening in the plots you posted: even though only two nodes have their FIFOs overflow, all three display erratic clock frequency jumps.

* If we _do_ reset buffers of a node A, its clock frequency shouldn't jump radically either. It should stay where it is, until the clock control kicks in - and then it should move closer to the settle frequency.

I don't have time to find the cause of these issues, but there's at least two things you should check:

* Nodes should continue to pump data to other nodes, even though their FIFOs are being reset.
* The clock multiplier boards should _not_ have their state reset. Just the clock control.

------------------------------------

In more general terms: I don't expect `callisto`'s type signature to change. It should be held in reset as long as you're still resetting the FIFOs. Make sure that its output is forced to `NoChange` when in reset.

Could you also add documentation to `ebController`? It should also have tests testing basic behavior.

elastic-buffer-sim/src/Bittide/Simulate.hs
```diff
@@ -85,21 +85,31 @@ tunableClockGen settlePeriod periodOffset stepSize _reset speedChange =
...
-  (outRd, outWr) = elasticBuffer size clkRead clkWrite rdEna wrEna
+  (outRd, outWr) = elasticBuffer size clkRead clkWrite rdToggle wrToggle
+
+  go True (dc, False) False = Just dc
+  go _ _ _ = Nothing
 
-  go True (dc, False) = Just dc
-  go _ _ = Nothing
+  overflowRd =
+    dualFlipFlopSynchronizer clkWrite clkRead resetGen enableGen False (snd <$> outWr)
```
martijnbastiaan, line 110: This is something I hadn't considered yet. You probably want to drain the FIFO completely and then start reading again at its midpoint, because it's impossible to predict the latency write-to-read...

vmchale:
[clockscomplete5.pdf](https://github.com/bittide/bittide-hardware/files/9633714/clockscomplete5.pdf)
[elasticbufferscomplete5.pdf](https://github.com/bittide/bittide-hardware/files/9633715/elasticbufferscomplete5.pdf)

Here are the data with the `dualFlipFlopSynchronizer` fix. I will try to offer more explanation (or a fix) shortly, but these look different.

elastic-buffer-sim/src/Bittide/Simulate.hs
vmchale: Sure.

elastic-buffer-sim/src/Bittide/Simulate.hs
```diff
@@ -84,65 +85,118 @@ tunableClockGen settlePeriod periodOffset stepSize _reset speedChange =
...
+   where
+    f :: Bool -> (DataCount, Underflow) -> (Bool, Bool)
+    f False (_, False) = (False, True)
+    f _ (_, True) = (True, False)
+    f True (d, _) | d == targetDataCount size = (False, True)
+    f _ _ = (False, False)
+
+  wrToggle =
+    register clkWrite resetGen enableGen True
+      $ mealy clkWrite resetGen enableGen g False outWr
```
martijnbastiaan, line 127:
```suggestion
    register clkWrite resetGen enableGen True
      $ mealy clkWrite resetGen enableGen g False outWr
```

`resetGen` is not synthesizable. Add a reset line to `ebController` and route it here. If you need a register without a reset, try [delay](https://hackage.haskell.org/package/clash-prelude-1.6.4/docs/Clash-Explicit-Signal.html#v:delay).

elastic-buffer-sim/src/Bittide/Simulate.hs
```diff
@@ -84,65 +85,118 @@ tunableClockGen settlePeriod periodOffset stepSize _reset speedChange =
...
+  go _ _ _ = (deepErrorX "Data count censored", True)
+
+  overflowRd =
+    dualFlipFlopSynchronizer clkWrite clkRead resetGen enableGen False (snd <$> outWr)
+
+  rdToggle =
+    register clkRead resetGen enableGen True
+      $ mealy clkRead resetGen enableGen f False outRd
+   where
+    f :: Bool -> (DataCount, Underflow) -> (Bool, Bool)
```
martijnbastiaan, line 119: I can't keep up with all the `Bool`s, could you add type aliases + a comment on what this is doing?

vmchale:
OK, still getting basically the same thing with the read-domain-synced "reset"-to-half mechanism

[elasticbufferscomplete5.pdf](https://github.com/bittide/bittide-hardware/files/9654421/elasticbufferscomplete5.pdf)
[clockscomplete5.pdf](https://github.com/bittide/bittide-hardware/files/9654422/clockscomplete5.pdf)

I think it's pretty clear the nodes are "decoupled" sort of, but the jump is still sharp and suspicious.

vmchale:
>If we do reset buffers of a node A, its clock frequency shouldn't jump radically either. It should stay where it is, until the clock control kicks in - and then it should move closer to the settle frequency.

We reset `c_est` and `x_k` (the integral of the measurement) when resetting clock control, so I don't think it's unexpected that it will do something sort of different, of course it also changes `r_k` (the measured elastic buffer occupancy).

So I don't think jumps are implausible. 

I've tweaked the line width in the below plots so that the stark jumps are visible in context of general operation:

[elasticbufferscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9655000/elasticbufferscomplete3.pdf)
[clockscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9655003/clockscomplete3.pdf)

martijnbastiaan:
Thanks for the comments! Why don't we see the jump right at the beginning of the plots?

martijnbastiaan:
Perhaps we should plot the internals of the clock control too to get a better understanding. Just simulating right up to when it jumps shod be good enough.

vmchale:
Yes agree! I'll dump to csv including the requested clock changes I think.

> Why don't we see the jump right at the beginning of the plots?

I'm not sure. Right after a reset is somewhat strange in that the occupancies will be 64, even if that clock happens to be fastest. 

Hopefully dumping more of the internals will give a better picture.

vmchale:
>Why don't we see the jump right at the beginning of the plots?

At the beginning, the elastic buffer occupancies are all the same! (and then drift apart slowly). When an elastic buffer is reset, however, it gets set to 64, while other nodes will be 70, 30, or whatnot. So measurements `r_k` will be quite different after a reset compared to the beginning.

martijnbastiaan:
Ah but you should reset _all_ elastic buffers when detecting under/overflow, not just the one that overflowed :). All elastic buffers of a single node, that is, a node can't control other node's elastic buffers.

vmchale:
Ok I have a tentative commit - 361116b6b69d10c121264978fe4c9b8aca7df7f7 - which orchestrates the resets within a node using TH

However, it's giving odd results, so I will continue looking for an explanation.

[elasticbufferscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9666821/elasticbufferscomplete3.pdf)
[clockscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9666822/clockscomplete3.pdf)

vmchale:
Still getting odd results, but different -

[elasticbufferscomplete5.pdf](https://github.com/bittide/bittide-hardware/files/9674947/elasticbufferscomplete5.pdf)
[clockscomplete5.pdf](https://github.com/bittide/bittide-hardware/files/9674949/clockscomplete5.pdf)

(I am suspicious about how the resets are orchestrated per-node so I will look there next).

vmchale:
[clockscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9678158/clockscomplete3.pdf)
[elasticbufferscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9678159/elasticbufferscomplete3.pdf)

Ended up being a bit painful to orchestrate, but it does seem to work. I will clean up the code.

elastic-buffer-sim/src/Bittide/Simulate.hs
```diff
@@ -84,65 +85,156 @@ tunableClockGen settlePeriod periodOffset stepSize _reset speedChange =
...
+  Enable writeDom ->
+  (Signal readDom DataCount, Signal readDom ForceReset, Signal readDom ResetClockControl)
+ebController size clkRead rstRead enaRead clkWrite rstWrite enaWrite =
+  unbundle
+    (go <$> rdToggle <*> outRd <*> overflowRd <*> requestRst)
+ where
+  rstReadS = unsafeToHighPolarity rstRead
+
+  -- tie off reset to mealy while we do our "reset" dance
+  ne'erRst = unsafeFromHighPolarity (pure False)
```
martijnbastiaan, line 128: I've noted this in my larger review, but we weary whenever you write this. Good change you actually wanted to use a specialized function (e.g., `delay`) or not want this at all.

martijnbastiaan (changes requested):
Hey that looks good.

It seems to me your code could be simpler and more testable if you would design separate components each handling smaller parts. Your code can also be simpler too if you define specific data types and only convert them locally; i.e. take `FifoMode` as an argument in a FIFO wrapper:

```haskell
data EbMode = Drain | Fill | Pass

ebModeToReadWrite :: EbMode -> (Bool, Bool)
ebModeToReadWrite = \case
  Fill  -> (False, True)
  Drain -> (True, False)
  Pass  -> (True, True)
```

While this is a little bit of a lie (write enable should be in the write domain) it doesn't matter for our purposes.

More importantly though, I think you code would be much simpler to follow if you design your controller such that it puppeteers multiple elastic buffers.

-------

For starters, extract your `direct` function and strip it to its minimal functionality: resetting a single buffer to its midpoint:

```haskell
type Done = Bool

resetEb
  ElasticBufferSize ->
  Clock readDom ->
  Reset readDom ->
  Signal readDom DataCount ->
  Signal readDom (EbMode, Done)
```

Like with your current code, this component should first fully drain the fifo before converging to its midpoint. The nice thing is that the state machine can be quite a bit simpler than what you have now:

```haskell
nextState = 
  case currrentState of
    Drain
      | n == 0    -> Fill
      | otherwise -> Drain

    Fill
      | n == target -> Pass
      | otherwise   -> Fill

    Pass -> 
      Pass

ebMode = nextState
done = currentState == Pass
```

Note that this component gets "stuck", and should be reset by an outer component. Combine this function into a super-function handling a bunch of elastic buffers:

```haskell
resetEbs
  ElasticBufferSize ->
  Clock readDom ->
  Reset readDom ->
  Signal readDom (Vec n DataCount) ->
  Signal readDom (Vec n EbMode, Done)
```

---------------------------

Now the outer logic looks a bit like:

```haskell
data EbControl
  = Init
  | ResetEbs
  | Wait
  | Stable
```

With roughly these steps:

1. `Init`: move to `Reset`. The purpose of `Init` is to make this FSM resetabble in its own right. It also makes sure we've reset `resetEbs` before moving to (2).
2. `Reset`: Wait for `resetEbs` to emit _done_.
3. `Wait`: if any overflow, move to `Reset`.
4. `Wait`: if all "stable" move to `Stable` (future PR).

Signals to propagate to the clock multiplier board:

1. `Init`/`Reset`: propagate `NoChange`.
2. `Wait`/`Stable`: propagate clock control signals

Signals to propagate to the EB:

1. `Init`/`Wait`/`Stable`: propagate `Pass`.
2. `Reset` propagate `resetEbs`'s output

The reset logic should be quite straightforward from there, roughly:

```haskell
resetResetEbs = state /= ResetEbs
resetClockControl = state == ResetEbs
```

Note that resets should never stand on their own, but ought to be combined with their outer resets. E.g.,

```haskell
resetResetEbs = forceReset (state /= ResetEbs) rst
resetClockControl = forceReset (state == ResetEbs) rst
```

given a function:

```haskell
forceReset :: Reset dom -> Signal dom Bool -> Reset dom
```

I could have sworn we had that in `clash-prelude`, but apparently not..

martijnbastiaan:
Also, if I were you I'd add some manually defined, simple topologies and ditch the TH while you're figuring out type signatures. Probably makes things a whole lot easier :sweat_smile:.

vmchale:
Thanks! Some of that looks much better, I'll start fixing it up :)

vmchale:
Something is going awry because of our attempts to control across fast/slow clock domains! Which I found via the unit test I just pushed - `directEbs` does not hang when the clocks are the same, but it _does_ hang with the `Slow` and `Fast` clock domains (even without dynamic clocks).

I believe this is because our current design requires that control occurs in the read domain - though I am not sure exactly what is happening.

vmchale:
Seems to be

```haskell
  ebRdOver = dualFlipFlopSynchronizer wrClk rdClk rdRst rdEna False isOverflow
  -- combinatorial loop here...
  FifoOut{..} = ebWrap rdClk rdRst rdToggle wrClk wrRst wrToggle
```

specifically. Somehow using the overflow signals (as inputs) to the controllers goes awry.

vmchale:
I have eb-controller-bumpy (which doesn't bother with any coordination) to build on while I figure out what is happening.

vmchale:
Is it possible I am being tripped up by trying to "call values from the future" here? It's mentioned here:

https://hackage.haskell.org/package/clash-prelude-1.6.4/docs/Clash-Explicit-Signal.html#v:unsafeSynchronizer

But perhaps it's possible to trip up by depending on this value even with a `dualFlipFlopSynchronizer`?

```
λ:> sampleN 100 $ dualFlipFlopSynchronizer (clockGen @D2) (clockGen @D5) resetGen enableGen 3 (fromList [(1::Int)..])
[3,3,3,2,2,3,3,3,4,4,5,5,5,6,6,7,7,7,8,8,9,9,9,10,10,11,11,11,12,12,13,13,13,14,14,15,15,15,16,16,17,17,17,18,18,19,19,19,20,20,21,21,21,22,22,23,23,23,24,24,25,25,25,26,26,27,27,27,28,28,29,29,29,30,30,31,31,31,32,32,33,33,33,34,34,35,35,35,36,36,37,37,37,38,38,39,39,39,40,40]
```

The behavior is correct but I can't tell if it depends on the "future" stuff mentioned in the documentation (which would be why the example unit test doesn't complete)

martijnbastiaan:
Try and start out with _just_ the FIFO IP. Do you get combinatorial loops in that case too?

vmchale:
I get cominatorial loops when using the Xilinx FIFO proper (eb-controller-sim-xilinx-fifo branch) (i.e. not a mock - thus the inputs/outputs should be registered properly). 

The `eb-controller-bumpy` branch has a demonstration where the node resets are _not_ coordinated, i.e. a buffer overflow only resets the clock controller and itself (rather than all elastic buffers in that node). The clocks _do_ converge, but with large jumps (ugly). However, that branch would be synthesizable and everything in `directEbs` would be synthesizable once it works.

martijnbastiaan:
I've looked at the branches, but all the temporary commits make it really hard for me to analyze. Could you clean it up such that each commit is a self-contained, well-described chunk of work? In general I'd like to ask you to use `git rebase -i` much more aggressively.

The best thing you can do to debug it is to try and make the example smaller and smaller, until you don't hit the `<<loop>>` anymore. Ideally you should arrive at a self-contained example that can be used in an issue report over at `clash-lang/clash-compiler`. (Or realize there's an actual problem with the current code surrounding the FIFO.)

vmchale:
>I've looked at the branches, but all the temporary commits make it really hard for me to analyze. Could you clean it up such that each commit is a self-contained, well-described chunk of work? In general I'd like to ask you to use git rebase -i much more aggressively.

Yes, will do.

> The best thing you can do to debug it is to try and make the example smaller and smaller, until you don't hit the <<loop>> anymore.

There appear to be two `<<loop>>`s, one can be avoided by replacing `dualFlipFlopSynchronizer wrClk rdClk rdRst rdEna False isOverflow` with `dualFlipFlopSynchronizer wrClk rdClk rdRst rdEna False (pure False)`.

vmchale:
Ok here's a minimal example, paring down what is necessary for per-node orchestration:

```haskell
ebReadOver ::
  forall readDom writeDom.
  (KnownDomain readDom, KnownDomain writeDom) =>
  Clock writeDom ->
  Clock readDom ->
  Reset readDom ->
  Enable readDom ->
  Signal readDom Overflow
ebReadOver wrClk rdClk rdRst rdEna =
  ebRdOver
 where
  ebRdOver = dualFlipFlopSynchronizer wrClk rdClk rdRst rdEna False isOverflow
  (_, isOverflow) = unbundle wrOuts
  (_, wrOuts) = elasticBuffer 128 rdClk wrClk rdToggle (pure True)

  rdToggle =
    delay rdClk rdEna True $
      mealy rdClk rdRst rdEna go () ebRdOver
   where
    go :: () -> Overflow -> ((), Bool)
    go _ _ = ((), True)
```

Basically, the read toggle signal is a constant (`True`), but it takes in `ebRdOver`, i.e. the overflow signal, converted to the read domain. 

This hangs indefinitely when I try to `sampleN` the output, but only when the write domain has a faster clock than the read domain.

martijnbastiaan:
Superseded by https://github.com/bittide/bittide-hardware/pull/189

PR #146, opened 2022-09-24
Add keep alive job to prevent runner cleanup

martijnbastiaan:
I've tested it by adding

```
  push:
    branches:
      - keep_alive
```

to `keep_alive.yml`. Job results:

https://github.com/bittide/bittide-hardware/actions/runs/3117600328

I've removed it again because we only want to run it on a schedule.

### DIFF
--- a/.github/workflows/keep_alive.yml
+++ b/.github/workflows/keep_alive.yml
@@ -1,0 +1,29 @@
+# SPDX-FileCopyrightText: 2022 Google LLC
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# GitHub automatically removes runners that have been idle for a number of days.
+# To prevent this from happening, we run a whole bunch of no-op jobs that makes
+# GitHub believe the runners are still active.
+#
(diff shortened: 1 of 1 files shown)


leonschoorl (approved):
If Github actually removes runners just for being idle to long when there just aren't any jobs to run, that seems like a bug to me.

I did find in the [github docs](https://docs.github.com/en/actions/hosting-your-own-runners/removing-self-hosted-runners):
> A self-hosted runner is automatically removed from GitHub if it has not connected to GitHub Actions for more than 14 days.

But I couldn't find anything about it removing idle runners.
Unless "not connected to GitHub Actions" means being idle.

But if this is what is takes

martijnbastiaan:
Okay, that 14 days is actually what we saw, so I probably misconfigured a systemd service. I had previously only found:

> A self-hosted runner is automatically removed from GitHub Enterprise Server if it has not connected to GitHub Actions for more than 30 days.

Hence the confusion. But that must simply be another product.

PR #148, opened 2022-09-26
Use `MemBlob` for initial contents in `wbStorage`

lmbollen:
This PR goes on top of #103

This pull requests changes the way we pass initial contents onto memory components.
For wishbone storage components we can define whether we have initial contents or not (and if we do, will they be reloaded on reset?).
We also introduce multiple ways to set the initial contents for more complex storages. (e.g. per byte).

### DIFF
--- a/bittide/src/Bittide/DoubleBufferedRam.hs
+++ b/bittide/src/Bittide/DoubleBufferedRam.hs
@@ -5,9 +5,10 @@
 {-# OPTIONS_GHC -fconstraint-solver-iterations=7#-}
 
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE GADTs #-}
+{-# LANGUAGE TypeApplications #-}
 
 module Bittide.DoubleBufferedRam where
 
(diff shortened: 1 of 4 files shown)


bittide/src/Bittide/DoubleBufferedRam.hs
```diff
@@ -21,93 +22,85 @@ import Protocols.Wishbone
...
-data InitialContent elements a where
-  NonReloadableV  :: Vec elements a -> InitialContent elements a
-  ReloadableV     :: Vec elements a -> InitialContent elements a
-  NonReloadableMb :: MemBlob elements (BitSize a) -> InitialContent elements a
-  ReloadableMb    :: MemBlob elements (BitSize a) -> InitialContent elements a
-  Undefined       :: (1 <= elements, KnownNat elements) => InitialContent elements a
+data ContentType n a
+  = Vec (Vec n a)
+  | Blob (MemBlob n (BitSize a))
+  | ByteVec (Vec (Regs a 8) (MemBlob n 8))
```
cuddlefishie, line 28: Would `BlobVec` make more sense? When I read `ByteVec` further down below I first though it was a `Vec n (Bytes m)` and had to come back up to double check

lmbollen: I agree, changed it

martijnbastiaan (approved):
LGTM, on the condition that we make an issue stating we should move everything to one type in the future. (No difference between testing and synthesis, even if it seems innocent like this case.)

Issue #121, opened 2022-09-09
Task list: Milestone 1


martijnbastiaan:
For Milestone 1 we need to implement MVP1, see:

* https://docs.google.com/presentation/d/1FKSmPLfcVCrH6pLv9rVWCFt8mquDH75FkEW6pyc2hNc/edit#slide=id.g1301b2c61bb_0_3010.
* https://docs.google.com/document/d/1HUea75PvYXaFrj6qQUjSRQKGKXk6QJ5Vyrsqrg-CDLo/edit

Tasks:

- [ ] Expose LEDs
- [ ] Memblob issue
- [ ] https://github.com/google-research/bittide/pull/103 (review @cuddlefishie )
- [x] https://github.com/google-research/bittide/pull/127 Rust firmware: Scatter+Gather (@cuddlefishie). Abstraction for metacycle register.
- [x] ~~Rust firmware: Calendar (@cuddlefishie). Write to shadow calendar, read from shadow calendar, change depth of calendar, flip-at-end-of-metacycle register.~~
- [x] https://github.com/google-research/bittide/pull/38
- [ ] Write the ping-pong application + simulation functions

martijnbastiaan:
Closed due to new SOW.

PR #151, opened 2022-09-30
Demo 1 Ping, Pong and Management programs

hydrolarus:
Contains #140 as a cherry picked commit.

Provides the Rust programs for #121


martijnbastiaan:
Let's get this one off now that we have developed the switch demo :)

PR #153, opened 2022-10-12
 Fix addressing bug and change test to cover bug 

lmbollen:
This PR fixes an addressing bug in wbStorage where wbStorage would terminate transactions with an error whenever the given address targest the upper half of its memory.
The test was adjusted to also generate inputs that should result in errors.

Furthermore the component has been changed to store data as Little Endian to conform with RISC-V.

### DIFF
--- a/bittide/src/Bittide/DoubleBufferedRam.hs
+++ b/bittide/src/Bittide/DoubleBufferedRam.hs
@@ -142,7 +142,6 @@ wbStorage' ::
   Signal dom (WishboneS2M (Bytes 4))
 wbStorage' initContent wbIn = delayControls wbIn wbOut
  where
-  depth = resize $ bitCoerce (maxBound :: Index depth)
   romOut = case initContent of
     Reloadable content-> bundle $ contentGenerator content
     other -> deepErrorX ("wbStorage': No contentgenerator for " <> show other)
@@ -184,7 +183,7 @@ wbStorage' initContent wbIn = delayControls wbIn wbOut
 
     (bitCoerce . resize -> wbAddr :: Index depth, alignment) =  split @_ @(aw - 2) addr
 
-    -- when the second lowest bit is not set, the address is considered word aligned.
+    -- When the second lowest bit is not set, the address is considered word aligned.
     -- We don't care about the first bit to determine the address alignment because
     -- byte aligned addresses are considered illegal.
     (not -> wordAligned, byteAligned) = bimap unpack unpack $ split alignment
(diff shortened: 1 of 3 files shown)


bittide/src/Bittide/DoubleBufferedRam.hs
```diff
@@ -150,7 +149,9 @@ wbStorage' SNat initContent wbIn = delayControls wbIn wbOut
     -- We don't care about the first bit to determine the address alignment because
     -- byte aligned addresses are considered illegal.
     (not -> wordAligned, byteAligned) = bimap unpack unpack $ split alignment
-    addrLegal = addr <= (2 * depth) && not byteAligned
+
+    -- The depth is the number of words (4 bytes), while the wishbone address addresses per byte.
```
vmchale, line 153: > while the wishbone address addresses per byte.

I don't understand this phrase.

lmbollen: PTAL

martijnbastiaan (changes requested):
I don't think the title matches the content. You're only fixing the `*2` => `*4` bug and tests. The rest are cosmetic changes.

martijnbastiaan (approved):

bittide/tests/Tests/DoubleBufferedRam.hs
```diff
@@ -679,67 +680,88 @@ deriving instance ShowX a => ShowX (RamOp i a)
...
-        contentList = toList $ concatMap (\(split ->(a,b)) -> b :> a :> Nil) content
-        fstGolden = deepErrorX "First element undefined."
-
+        contentList = toList $ concatMap (\(split ->(b,a)) -> a :> b :> Nil) content
       topEntity wbIn =
-        wcre @System $ wbStorage' @_ @_ @32 (NonReloadable $ Vec content) wbIn
-      topEntityInput = L.concatMap (\a ->[a, a, emptyWishboneM2S]) (ramOpToWb <$> goldenInput)
+        wcre $ wbStorage' @System @words @32 (NonReloadable $ Vec content) wbIn
+
+      goldenInput    = fromRight RamNoOp <$> generatedInput
```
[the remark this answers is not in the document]
lmbollen: Indeed all illegal inputs that should lead to an error are removed from the `goldenInput`. The behavorial model only processes the legal inputs and the errors that result from the illegal inputs are afterwards added in line `724`

```Haskell
goldenTransactions = L.zipWith expectTransaction generatedInput simGolden
```

```Haskell
expectTransaction ramOp response = case ramOp of
  Right op -> ramOpToTransaction op response
  Left  op -> Just (Error (ramOpToWb op))
```

PR #155, opened 2022-10-13
Eliminate need for ascending addresses in `singleMasterInterconnect`

lmbollen:
This PR allows for unordered baseaddresses in the `MemoryMap` for `singleMasterInterconnect`. 

Due to the method of slave selecting, the base addresses were required to be in ascending order.
With this change we first sort the 'MemoryMap' alongside the incoming busses to satisfy this requirement.
We also keep track of the original order of the incoming busses to make sure we correctly route the outgoing busses.

The tests were also adjust to now generate configurations with unordered base addresses.

Furthermore we decreased the size of the tests to improve speed, it is expected that this has no effect on the thoroughness of the tests.

### DIFF
--- a/bittide/src/Bittide/SharedTypes.hs
+++ b/bittide/src/Bittide/SharedTypes.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE UndecidableInstances #-}
@@ -27,6 +28,20 @@ data AorB = A | B deriving (Eq, Generic, BitPack, Show, NFDataX)
 swapAorB :: AorB -> AorB
 swapAorB A = B
 swapAorB B = A
+
+-- | Polymorphic record update of 'addr'.
+updateM2SAddr ::
+  BitVector addressWidthNew ->
+  WishboneM2S addressWidthOld selWidth dat ->
+  WishboneM2S addressWidthNew selWidth dat
+updateM2SAddr newAddr WishboneM2S{..} = WishboneM2S{addr = newAddr, ..}
+
+-- | Resize 'WishboneM2S's 'addr' field.
+resizeM2SAddr ::
+  (KnownNat addressWidthOld, KnownNat addressWidthNew) =>
+  WishboneM2S addressWidthOld selWidth dat ->
+  WishboneM2S addressWidthNew selWidth dat
+resizeM2SAddr WishboneM2S{..} = WishboneM2S{addr = resize addr, ..}
 
 -- | A single byte.
 type Byte = BitVector 8
--- a/bittide/tests/Tests/Wishbone.hs
+++ b/bittide/tests/Tests/Wishbone.hs
@@ -2,20 +2,24 @@
 --
 -- SPDX-License-Identifier: Apache-2.0
 
+{-# OPTIONS_GHC -fconstraint-solver-iterations=5 #-}
+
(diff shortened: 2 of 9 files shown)


bittide/src/Bittide/SharedTypes.hs
```diff
@@ -158,3 +161,7 @@ forceReset force = unsafeFromHighPolarity (unsafeToHighPolarity hasReset .||. fo
 -- | Divide and round up.
 divRU :: Integral a => a -> a -> a
 divRU b a = (b + a - 1) `div` a
+
+-- | Sort a vector using 'Data.List.sortOn'.
+sortVec :: (KnownNat n, Ord b) => (a -> b) -> Vec n a -> Vec n a
+sortVec f v = unsafeFromList (L.sortOn f (toList v))
```
cuddlefishie, line 167: I guess this won't be able to translate to HDL because it relies on `Data.List.sort`? For Haskell-simulation this should work fine, but maybe it would be nice to have a `Clash.Sized.Vec.sort` function eventually

lmbollen: Unfortunately, you're right.
I'm working on it as we speak.

martijnbastiaan:
I don't understand the need to sort addresses. Each slave/master select should happen independently of others. Any overlapping memory addresses are a configuration error and can lead to hardware failure - which is fine. We can later build infrastructure that makes sure no addresses overlap (i.e., CI failure).

martijnbastiaan:
I talked to @lmbollen at the office about this today, here is a summary (thought @cuddlefishie might be interested):

Interconnects like these are usually somewhat limited to reduce complexity significantly. Typically, you'd specify both the base address and the size of a slave in multiples of powers of two. E.g., a 4 GiB memory space might be divvied up in 4 parts of each 1 GiB. Selecting a slave therefore would only amount to checking out the two most significant bits: `00` for the first quadrant, `01` for the second, etc.

My suggestion would be to create an IP that takes advantage of this, roughly parameterize in the following aspects:

 * The size they're interested in. E.g., the "main" interconnect might be interested in the full 4 GiB space, while smaller ones only care about 1 GiB quadrants referred to earlier. The former needs to take into account the full address space (though it might ignore a lot of bits based on the number of divisions). The latter should ignore the upper 2 bits unconditionally.
 * The number of divisions they divide their memory space into. Or put in another way; the size of each subdivision. 

We can hardcode it at 32-bit addresses for now. We're probably not going to switch soon, and it will hopefully make our type signatures a bit easier.

bittide/tests/Tests/Wishbone.hs
```diff
@@ -41,106 +46,155 @@ memMapGroup = testGroup "Memory Map group"
 
 -- | generates a 'MemoryMap' for 'singleMasterInterconnect'.
 genConfig ::
-  forall nSlaves aw .
-  (KnownNat nSlaves, KnownNat aw) =>
-  Proxy nSlaves ->
-  Gen (MemoryMap nSlaves aw)
+  forall nSlaves .
+  (KnownNat nSlaves) =>
+  SNat nSlaves ->
```
martijnbastiaan, line 51:
```suggestion
  SNat nSlaves ->
```

martijnbastiaan (changes requested):
Overall LGTM.

I'll add doctest support to `bittide` in another PR.

martijnbastiaan (approved):
LGTM, but change the commit title to something more specific:

```
Eliminate need for ascending addresses in `singleMasterInterconnect`
```

Note that this strategy does make it impossible to fix a single slave to a specific address. I.e., one slave responds to MSBs `000`, while another responds to `111`. That's fine for us though.

martijnbastiaan:
@lmbollen See

https://github.com/clash-lang/clash-compiler/blob/9ae2c195a7be7b3e49eff693ade63ac17f9dfea0/clash-prelude/src/Clash/Class/BitPack/Internal.hs#L58-L62

for an example on how to use `$setup` blocks.

Issue #124, opened 2022-09-12
Use components in `bittide` in `elastic-buffer-sim`


martijnbastiaan:
We currently use a "fake" FIFO in `elastic-buffer-sim`. This should be replaced with the one in https://github.com/clash-lang/clash-compiler/pull/2270. We should probably keep the old one around, and make it easy to switch between implementations. I'm having a suspicion that the real deal will be a _lot_ slower to simulate.

Impeded by: https://github.com/bittide/bittide-hardware/issues/207

martijnbastiaan:
Fixed in https://github.com/bittide/bittide-hardware/pull/279

PR #157, opened 2022-10-18
Use Xilinx FIFO model

vmchale:
This closes https://github.com/bittide/bittide-hardware/issues/124

With the actual Xilinx FIFO model we are closer to a synthesizable clock control (what remains is the float conversion IP).

### DIFF
--- a/elastic-buffer-sim/src/Bittide/Simulate.hs
+++ b/elastic-buffer-sim/src/Bittide/Simulate.hs
@@ -9,9 +9,11 @@ Provides a rudimentary simulation of elastic buffers.
 -- SPDX-License-Identifier: Apache-2.0
 
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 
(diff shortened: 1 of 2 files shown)


elastic-buffer-sim/src/Bittide/Simulate.hs
```diff
@@ -90,12 +93,38 @@ type HasUnderflowed = Bool
 type HasOverflowed = Bool
 type DisableTilHalf = Bool
 
+elasticBufferXilinx ::
+  forall readDom writeDom.
```
martijnbastiaan, line 97: You're currently making a wrapper around the Xilinx FIFO to make it act like `elasticBuffer` (including half-way logic n stuff), but I think we should flip this upside-down. Make the model behave like a very simple FIFO (including starting a zero), and move all the half-way logic to a wrapper.

martijnbastiaan:
Superseded by https://github.com/bittide/bittide-hardware/pull/189

PR #158, opened 2022-10-20
Improve endianness throughout designs + rxUnit refactor.

lmbollen:
During the simiulation of Bittide nodes, I encountered the problem that the endianness of components was mostly big endian, while riscv uses little endian.

To fix it I changed our `RegisterBank` type to contain endianness information with a new type `ByteOrder`.

The purpose of the changes is to create clarity regarding the endianness of components (Everything accessible with Wishbone should be `LittleEndian`) and enforce the correct conversion between `RegisterBank`s and the contained data.

### DIFF
--- a/bittide/src/Bittide/Link.hs
+++ b/bittide/src/Bittide/Link.hs
@@ -2,7 +2,7 @@
 --
 -- SPDX-License-Identifier: Apache-2.0
 
-{-# OPTIONS_GHC -fconstraint-solver-iterations=0 #-}
+{-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}
 
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
--- a/bittide/src/Bittide/Link.hs
+++ b/bittide/src/Bittide/Link.hs
@@ -2,7 +2,7 @@
 --
 -- SPDX-License-Identifier: Apache-2.0
 
-{-# OPTIONS_GHC -fconstraint-solver-iterations=5 #-}
+{-# OPTIONS_GHC -fconstraint-solver-iterations=0 #-}
 
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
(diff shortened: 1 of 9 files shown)


cuddlefishie (commented):
Looks good! Some small formatting suggestions and an (for now?) unused function

bittide/src/Bittide/SharedTypes.hs
```diff
@@ -69,52 +67,61 @@ type Pad a bw  = (Regs a bw * bw) - BitSize a
...
+  => NFDataX (RegisterBank regSize content byteOrder)
+
+deriving newtype instance (KnownNat regSize, ShowX (RegisterBank regSize content byteOrder)) =>
+  ShowX (RegisterBank regSize content byteOrder)
+
+convertLe ::
+  (Paddable a, KnownNat bw, 1 <= bw) =>
+  (RegisterBank bw a 'LittleEndian, a)  ->
+  (a, RegisterBank bw a 'LittleEndian)
+convertLe (regBank, a) = (getDataLe regBank, getRegsLe a)
```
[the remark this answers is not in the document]
lmbollen: I think that indeed it's not used anywhere currently, I mostly added it for completeness. 
I'll remove it, since it's trivial to reproduce anyways.

bittide/src/Bittide/SharedTypes.hs
```diff
@@ -69,52 +67,61 @@ type Pad a bw  = (Regs a bw * bw) - BitSize a
 -- Amount of bw sized registers required to store a.
 type Regs a bw = DivRU (BitSize a) bw
 
--- | Vector of registers that stores data coming from a communication bus. It can be used
--- to store any arbitrary data type, the last register is padded with p bits. The number of
--- registers and the amount of padding depends on the bit size of the stored data.
-newtype RegisterBank regSize content =
+data ByteOrder = LittleEndian | BigEndian
+-- | Stores any arbitrary datatype as a vector of registers.
```
martijnbastiaan, line 71:
```suggestion
data ByteOrder = LittleEndian | BigEndian

-- | Stores any arbitrary datatype as a vector of registers.
```

martijnbastiaan (commented):
My biggest question is: do we really want this to be configurable? RISCV is little endian, do we care about anything else?

bittide/src/Bittide/Link.hs
```diff
@@ -2,7 +2,7 @@
 --
 -- SPDX-License-Identifier: Apache-2.0
 
-{-# OPTIONS_GHC -fconstraint-solver-iterations=5 #-}
+{-# OPTIONS_GHC -fconstraint-solver-iterations=0 #-}
```
martijnbastiaan, line 5: pls no

martijnbastiaan (changes requested):

martijnbastiaan (approved):

PR #164, opened 2022-11-03
Check stability of clock control algorithms. 

vmchale:
This is a draft at the moment to check CI etc.


elastic-buffer-sim/src/Bittide/Topology/TH.hs
```diff
@@ -137,6 +139,52 @@ plotEbsQ g = do
...
 
+-- | Successive application on a stream.
+succApp :: (a -> a -> b) -> [a] -> [b]
+succApp op (x:xs@(y:_)) = x `op` y : succApp op xs
+succApp _ _ = error "This function is meant to be called on streams."
+
+streamSettle :: (Ord a, Integral a, Show a) => a -> a -> Bool
+streamSettle x y = asInteger x - asInteger y == 0
+ where
+  asInteger n = fromIntegral n :: Integer
```
martijnbastiaan, line 150: I think we can make this a bit more configurable. They don't have to be _exactly_ the same (in practice they're never going to be), just within ~5 ppm or so. This should make running the check way quicker too.

For example, if I run `complete3` for 10_000_000 cycles / 8 minutes locally I get:

[clockscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9989661/clockscomplete3.pdf)
[elasticbufferscomplete3.pdf](https://github.com/bittide/bittide-hardware/files/9989663/elasticbufferscomplete3.pdf)

Which seems (almost?) good enough.

martijnbastiaan (approved):
In general LGTM! Do you think you can make it a bit more configurable? (See comment)

kleinreact:
What's the status of this PR?

martijnbastiaan:
I think we can close it because:

1. We've removed the TH in https://github.com/bittide/bittide-hardware.
2. We'll reuse the existing (hardware) checks to determine stability.

PR #170, opened 2022-11-10
Add property test for haxioms

martijnbastiaan:
TODO:

- [x] `timesDivRU`
- [x] `clog2axiom`
- [x] `timesNDivRU`
- [x] `timesNDivRU'` ~~**property does not hold**~~ it did hold given the right constraints, but its name is very confusing (mentioning `DivRU`, but using `Div`).
  - Isn't used anyway, removed
- [x] `timesNDivRU''` ~~**property does not hold**~~ it did hold given the right constraints, but its name is very confusing (mentioning `DivRU`, but using `Div`).
  - Replaced by `divWithRemainder`
- [x] `lessThanMax` 
  - Very weird property? Usage almost certainly wrong
  - Replaced by `leMaxLeft` and `leMaxRight`
- [x] `oneLTdivRU`


lmbollen (approved):
A few suggestions:
* Can we get a haddock comment per haxiom that explains a bit what the haxiom does?
* I feel like the names are not descriptive enough (and making them more descriptive would require LARGE names). 
* Can we check simply replace all `Div (a + ( b - 1) b` with `DivRU a b` without breaking anything? 
* Maybe add a haddock comment to every test linking to the tested function. 

I did not spot any problems, so LGTM.

martijnbastiaan:
> I feel like the names are not descriptive enough (and making them more descriptive would require LARGE names).

We could call them _[Lucas's postulates](https://en.wikipedia.org/wiki/Euclid%27s_postulates)_.

cuddlefishie:
> Can we check simply replace all `Div (a + ( b - 1) b` with `DivRU a b` without breaking anything?

At least when I tried to integrate the clash-protocols wishbone types, GHC wasn't always able to work it out when using `DivRU`, which is why there were the two `timesNDivRU'` and `timesNDivRU''` axioms (because it *also* couldn't figure out commutativity)

Since this seems to work as it is now, I think it's a simplification and also better naming! :+1:

cuddlefishie (approved):

martijnbastiaan:
> Can we get a haddock comment per haxiom that explains a bit what the haxiom does?

Sure.

> I feel like the names are not descriptive enough (and making them more descriptive would require LARGE names).

Agreed. I've changed some (see Slack discussion). I'm leaving the rest for later.

>  Can we check simply replace all Div (a + ( b - 1) b with DivRU a b without breaking anything?

I'm leaving this for later :).

> Maybe add a haddock comment to every test linking to the tested function.

Done!

PR #172, opened 2022-11-11
Use `clockTicks` in `elasticBuffer`

martijnbastiaan:
See motivation in https://github.com/clash-lang/clash-compiler/pull/2349

### DIFF
--- a/elastic-buffer-sim/src/Bittide/Simulate.hs
+++ b/elastic-buffer-sim/src/Bittide/Simulate.hs
@@ -105,18 +105,18 @@ elasticBuffer ::
   Clock readDom ->
   Clock writeDom ->
   Signal readDom DataCount
-elasticBuffer mode size clkRead clkWrite
-  | Clock _ (Just readPeriods) <- clkRead
-  , Clock _ (Just writePeriods) <- clkWrite
-  = go 0 (targetDataCount size) readPeriods writePeriods
+elasticBuffer mode size clkRead clkWrite =
+  go (clockTicks clkWrite clkRead) (targetDataCount size)
  where
-  go !relativeTime !fillLevel rps wps@(writePeriod :- _) =
-    if relativeTime < toInteger writePeriod
-    then goRead relativeTime fillLevel rps wps
-    else goWrite relativeTime fillLevel rps wps
+  go (tick:ticks) !fillLevel =
+    case tick of
+      ClockA  -> goWrite ticks fillLevel
+      ClockB  -> goRead ticks fillLevel
+      ClockAB -> go (ClockB:ClockA:ticks) fillLevel
+  go [] _ =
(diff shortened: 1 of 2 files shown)


elastic-buffer-sim/src/Bittide/Simulate.hs
```diff
@@ -105,44 +105,29 @@ elasticBuffer ::
...
   Signal readDom DataCount
-elasticBuffer mode size clkRead clkWrite
-  | Clock _ (Just readPeriods) <- clkRead
-  , Clock _ (Just writePeriods) <- clkWrite
-  = go 0 (targetDataCount size) readPeriods writePeriods
+elasticBuffer mode size clkRead clkWrite =
+  go (clockTicks clkWrite clkRead) (targetDataCount size)
  where
-  go !relativeTime !fillLevel rps wps@(writePeriod :- _) =
-    if relativeTime < toInteger writePeriod
```
DigitalBrains1, line 114: This logic is skewed. If we define `relativeTime` as:
> Relative time 0 points at the next active edge of the write clock, and `relativeTime` points at the next active edge of the read clock

then this invariant is broken: we start out with `relativeTime` pointing at the clock graph of the read clock at the point where a write clock edge would be rather than at a read clock edge. Maybe it is possible to come up with a definition of `relativeTime` that would allow you to reason about what the heck it is doing, but without drawing it I can't come up with what this does.

Anyway, that seems like a bug similar to the bug in `trueDualPortBlockRam` (hey that one doesn't have an issue report), and getting rid of it is probably the right thing to do.

There's one more thing: if two edges coincide (in this skewed counting that probably picks edges that don't really coincide), this code prefers the write clock, but the new code prefers the read clock.  This means your fill level will report the fill level from before the coinciding clock now.

DigitalBrains1 (approved):

elastic-buffer-sim/src/Bittide/Simulate.hs
DigitalBrains1: Erm wait I already applied a translation to `relativeTime` that I didn't describe. I reasoned thus:
* Normally we initialise at 0 and compare to 0
* Initialising at 0 and comparing to `writePeriod` is functionally identical to initialising at `-writePeriod` and comparing to 0.

If you first apply this translation, you can read the invariant.

martijnbastiaan: > There's one more thing: if two edges coincide (in this skewed counting that probably picks edges that don't really coincide), this code prefers the write clock, but the new code prefers the read clock. This means your fill level will report the fill level from before the coinciding clock now.

That's okay with me. Code shouldn't rely on this, and the new code follows the convention in `clockTicks`.

Thanks for the comment!

PR #178, opened 2022-11-17
Add ip component specific vivado subdirectory

lmbollen:
While creating a `bittide-instance`  for `callisto`,  I encountered the problem that we create no `ip` directory where created `ip`s can be stored.
This PR creates a local ip directory in each target's vivado folder in `_build` and makes sure we always start with fresh ips.


martijnbastiaan:
> I encountered the problem that we create no ip directory where created ips can be stored.

Is this a problem because the TCL scripts expect the dir to be there?

martijnbastiaan (approved):
> Is this a problem because the TCL scripts expect the dir to be there?

I'm going to interpret the :+1: as "yes". If so, please add to the commit body :)

PR #180, opened 2022-11-18
Pad calendar's initial contents with defined values 

lmbollen:
Partially undefined calendars do not get processed properly by Vivado and -fclash-force-undefined is broken for `BitVector`s https://github.com/clash-lang/clash-compiler/issues/2360 .


martijnbastiaan (approved):
Also add a an `-- XXX: ...` or `-- FIXME: ...` so we can grep on it.

PR #183, opened 2022-11-18
Use absolute build dirs in Shake files

### LINKED ISSUE #165: Shake inconsistent rebuild behaviour due to relative pathing
While adding components to `bittide-instances` I noticed that shake's rebuild behaviour depends on your current path.
The rebuild behavior depends on the `dist-newstyle` directory in your current path, while it `dist-newstyle` is actually located in the project folder (where `cabal.project` is located).

We should change the implementation to locate the `cabal.project` file and depend on the `dist-newstyle` in that directory.

martijnbastiaan:
Fixes #165

### DIFF
--- a/bittide-instances/bin/Shake.hs
+++ b/bittide-instances/bin/Shake.hs
@@ -9,12 +9,15 @@ module Main where
 import Prelude
 
 import Clash.Shake.Extra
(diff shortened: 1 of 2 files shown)


lmbollen (approved):
_works on my machine_
lgtm :)

PR #188, opened 2022-11-21
Only run RISC formal tests on 'main'

martijnbastiaan:
The formal tests currently run on _any_ change in `.github`. We change files in this directory quite often, but virtually never anything related to the RISC core. By only running tests on 'main' we'll still catch errors - albeit a bit late - but without waiting for them on every PR.

### DIFF
--- a/.github/scripts/has_riscv_changes.sh
+++ b/.github/scripts/has_riscv_changes.sh
@@ -20,6 +20,6 @@ fi
 
 git diff \
   --exit-code --quiet "${base_ref}".."${head_ref}" \
-  -- .github/ contranomy/ riscv-formal-config/ riscv-formal/
+  -- .github/ contranomy/ riscv-formal-config/ riscv-formal/ cabal.project
 
 [ $? -ne 0 ]


.github/workflows/ci.yml
```diff
@@ -263,7 +263,7 @@ jobs:
 
       - name: Get test matrix
         run: |
-          if .github/scripts/has_riscv_changes.sh; then
+          if "${{ github.ref_name }}" == "main" && .github/scripts/has_riscv_changes.sh; then
```
leonschoorl, line 266: This is fed to the shell, and its if syntax tries to run github.ref_name as command:
```
Run if "188/merge" == "main" && .github/scripts/has_riscv_changes.sh; then
/__w/_temp/b2e3ad64-7590-4c9c-bcda-01226f35575f.sh: 1: 188/merge: not found`
```

I think the correct syntax should be:
```suggestion
          if [[ "${{ github.ref_name }}" == "main" ]] && .github/scripts/has_riscv_changes.sh; then
```

leonschoorl (changes requested):
Alternatively we could make `has_riscv_changes.sh` more precise, something like: 
```diff
--- a/.github/scripts/has_riscv_changes.sh
+++ b/.github/scripts/has_riscv_changes.sh
@@ -20,6 +20,7 @@ fi
 
 git diff \
   --exit-code --quiet "${base_ref}".."${head_ref}" \
-  -- .github/ contranomy/ riscv-formal-config/ riscv-formal/
+  -- contranomy/ riscv-formal-config/ riscv-formal/  \
+     .github/scripts/*formal* .github/cabal.project.formal
```
And rename `.github/scripts/generate_checks.sh` to `.github/scripts/generate_formal_checks.sh`, so it's matched by the `.github/scripts/*formal*` above.

.github/workflows/ci.yml
martijnbastiaan: You'd think I'd learn after writing so many hack-scripts..

martijnbastiaan:
Changing the CI scripts to be more precise is fairly tricky: any changes to the logic _running_ the checks (say docker image, etc.) should also trigger a CI run. I'd therefore rather be somewhat more conservative. At some point we'll ditch Contranomy in favor of VexRisc, so the CI check shouldn't be long-lived anyway..

Your comment did trigger me to add `cabal.project` to the list of dependencies.

I've applied your Bash fix. Thanks!

leonschoorl (approved):
The `cabal.project` is a good addition.

PR #191, opened 2022-11-22
Add Xilinx specific 'tripleFlipFlopSynchronizer'

martijnbastiaan:
This adds a Xilinx specific `tripleFlipFlopSynchronizer`: a single-bit synchronization primitive. 

-----------

Schematic:

Selection_016

-----------

Given how close we are to a clock control demo, I'd suggest we leave the TODO for next week. I'll create issues for each of these before merging.

TODO:

 - [x] Use TCL API to generate register constraints, remove "hack" from `mkPlaceTcl`.
 - [x] Add Clash simulation / HDL simulation identity checks (like `clash-testsuite`).


lmbollen (approved):
On the condition that you add issues for the open TODOs.

PR #200, opened 2022-12-12
Nixify `bittide-hardware`

martijnbastiaan:
Note that this is an experiment. 

Wishlist:

 - [ ] Working `shell.nix` which would get us the right Scala, Python, Matplotlib, .. Maybe this is actually the only thing we should Nix for, while keeping all the other stuff managed by Cabal?
 - [ ] Nixify VexRisc (#198)
 - [ ] Nixify `bittide-instances`
 - [ ] Nixify `elastic-buffer-sim`


martijnbastiaan:
Superseded by https://github.com/bittide/bittide-hardware/pull/227

PR #201, opened 2022-12-13
Add type level time units.

lmbollen:
In Clash, time is expressed in picoseconds.
Previously,  period of 300ms would be represented by `300000000000`, or `300_000_000_000` with `NumericUnderscores`.

After this change, we can write `MilliSeconds 300` instead.

### DIFF
--- a/bittide/src/Bittide/Arithmetic/Time.hs
+++ b/bittide/src/Bittide/Arithmetic/Time.hs
@@ -4,10 +4,21 @@
 
 module Bittide.Arithmetic.Time where
 
-import Prelude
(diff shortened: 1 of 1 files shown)


martijnbastiaan (approved):
LGTM implementation-wise! Remove the trailing `.` for consistency in your commit title. Then, make the body read like a continuous text instead of three unrelated sentences (and use `` ` `` for readability purposes when reading the commits here on GitHub):

```
Clash's domains express time in picoseconds. A period of 300 milliseconds
is therefore represented as `300_000_000_000`. After this change, we can
write `MilliSeconds 300` instead.
```

PR #202, opened 2022-12-13
Lower number of Hedgehog tests in Bittide

lmbollen:
Previously every property would be tested `10_000` times. This change reduces the number of tests per property to `1_000`.

### DIFF
--- a/bittide/tests/UnitTests.hs
+++ b/bittide/tests/UnitTests.hs
@@ -33,7 +33,7 @@ tests = testGroup "Unittests"
   ]
 
 setDefaultHedgehogTestLimit :: HedgehogTestLimit -> HedgehogTestLimit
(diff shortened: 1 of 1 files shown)


DigitalBrains1:
This could significantly reduce coverage for shrinking generators, as these are quite skewed to small values. If there are a few tests that take long, perhaps you could create a solution where only those tests are run with a reduced number of tests? And perhaps you could make shrinking configurable for those tests, so coverage is less affected? I'm just spitballing here.

DigitalBrains1:
FYI, the cycling through sizes is here: `check` -> `checkNamed` -> `checkRegion` calls [`checkReport`](https://hackage.haskell.org/package/hedgehog-1.2/docs/src/Hedgehog.Internal.Runner.html#checkReport) with `size0` set to 0, and then it loops incrementing it up to and including 99 and back to 0.

martijnbastiaan (approved):

PR #210, opened 2022-12-14
`safeDffSynchronizer` with SDC files

DigitalBrains1:
- The `tripleFlipFlopSynchronizer` is renamed to `safeDffSynchronizer` because while it has three flip-flops, the first flip-flop is not part of the synchronizer. Rather, the first flip-flop is the _safe_ part: it makes sure the signal going into the synchronizer is as clean as can be.
- HDL generation generates an `.sdc` file that matches the globally unique names of the registers in the primitive and provides the desirable `max_delay` timing constraint. The unique name generation is hacky and it would be good if we could improve Clash to make this convenient, but the current solution works completely without caveats as far as I'm aware.

  Names for the two first flip-flops look like: `topEntity_dff_sync_8B6D4D15F5D7771F_dff_sync_a`

  The reason for the names containing `dff_sync` twice in the name is that the first occurence isn't part of the _locally_ unique name generated with `~GENSYM`. If we would have used `~GENSYM[a]`, it would need to be differentiated from all things named `a`. Picking `dff_sync_a` greatly reduces the contention for the name.
- A second function, `safeDffSynchronizer0`, provides access to the flip-flop in the source domain, in case the value is also needed in that domain.
- TemplateHaskell is leveraged to provide the function names in the primitive annotation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,5 @@ vhdl/
 vivado.jou
 log
 vivado_*
+tight_setup_hold_pins.txt
+.Xil
--- a/bittide-instances/bin/Shake.hs
+++ b/bittide-instances/bin/Shake.hs
@@ -71,7 +71,7 @@ targets =
   , 'ClockControl.callisto3
   , 'ElasticBuffer.elasticBuffer5
   , 'StabilityChecker.stabilityChecker_3_1M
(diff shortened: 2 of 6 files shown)


martijnbastiaan (approved):
LGTM! Let's wait for https://github.com/bittide/bittide-hardware/pull/211 to get merged before merging this.

PR #213, opened 2022-12-16
Work around clash-compiler#2376

martijnbastiaan:
See: https://github.com/clash-lang/clash-compiler/issues/2376

I've added `-fclash-debug DebugSilent` so we catch these issues earlier next time.


lmbollen (approved):
Perhaps add some indication to the code that shows that we can revert this change when clash-compile#2376 is closed.

martijnbastiaan:
Going by the comments on that issue, I'm not sure we can every fix it.. But yeah I'll add links.

Issue #205, opened 2022-12-14
Expose read/write ports of elastic buffer


martijnbastiaan:

lmbollen:
Updates:
* Exposing is done
* Working on elastic buffer control machinery.

PR #214, opened 2022-12-21
Expose read/write data ports of elastic buffer

hiddemoll:
In the previous version the read and write ports of the `xilinxElasticBuffer` and the `resettableXilinxElasticBuffer` were not exposed as we were only interested in its status (datacount and over/underflow flags). Now we do want access to the data in the elastic buffers.

See #205


lmbollen (approved):
lgtm

PR #217, opened 2022-12-23
Add Bittide domain

lmbollen:
Draft of me working on the Bittide clock control domain, this domain does not contain the elastic buffers as they all introduce new clock domains, which is hard if not impossible to generalise. 

The domain consumes a vector of `DataCount`s and `EbModes` from all elastic buffers.
This domain contains:
* Callisto
* StabilityCheckers for all incoming datacounts
* A statemachine that makes sure the clocks are stabilized and buffer occupancies are all close to `targetDataCount` 

The datacounts that are passed to callisto are set to `targetDataCount` whenever the producing elastic buffer is not in `Pass` mode, this essentially causes callisto to act as if this link does not exist.

Callisto will be reset whenever one of the elastic buffer overflows or underflows, it does so by monitoring the `EbMode`.
Recall that `EbMode` consists of the states: `Idle`, `Drain`, `Fill` and `Pass`.

The boot procedure state machine contains 4 states which will be explained below: 
* `Allunstable` : Clocks have yet to converge, callisto controls the local clock . In this state we wait untill *all* stability indicators are `True`.  Once all stability indicators are `True`, we move into the `ResetBuffers` state.
* `ResetBuffers`: We reset all elastic buffers to set all their dataCounts to `targetDataCount`. Important to note is that, since the elastic buffers are not part of this component, resetting the buffer has to be done based on the exposed state. We go to the `WaitForPassesState` as soon as not all the stability indicators are `True` anymore.
* `WaitForPasses`: We have reset all elastic buffers and are now waiting for them to become stable again. Once all stability indicators are `True`, we go into the final `Stable` state. This state should be used as indicator that system is synchronized.

### DIFF
--- a/bittide/bittide.cabal
+++ b/bittide/bittide.cabal
@@ -104,6 +104,7 @@ library
     Bittide.ClockControl.Si539xSpi
     Bittide.ClockControl.StabilityChecker
     Bittide.DoubleBufferedRam
+    Bittide.Domain
     Bittide.ElasticBuffer
     Bittide.Link
     Bittide.Node
--- a/bittide/src/Bittide/Domain.hs
+++ b/bittide/src/Bittide/Domain.hs
@@ -1,0 +1,80 @@
+-- SPDX-FileCopyrightText: 2022 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE RecordWildCards #-}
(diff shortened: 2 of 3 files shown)


martijnbastiaan:
Superseded by https://github.com/bittide/bittide-hardware/pull/279 (if that's not true, please reopen)

PR #221, opened 2023-01-02
Add verilog-ethernet subtree

lmbollen:
I'm not sure that this is the way we´d want this. Do we have to add any license related things?


martijnbastiaan:
> Do we have to add any license related things?

Let's not change anything in the subtree. Just do something like this:

https://github.com/bittide/bittide-hardware/blob/b92ba9bc7233b95b1db39bbf06fb190132c0a995/.reuse/dep5#L16-L18

We can prune as much as the tree as possible (keeping the relevant `LICENSE`s and stuff).

.reuse/dep5
```diff
@@ -36,3 +36,7 @@ License: Apache-2.0
 Files: bittide-instances/tests/reports/*
 Copyright: 2022 Google LLC
 License: Apache-2.0
+
+Files: verilog-ethernet
```
martijnbastiaan, line 40:
```suggestion
Files: verilog-ethernet/*
```

? Not sure, but this is what the other sections do.

lmbollen: Since all the included files already have licensing, I suspect that we are in the clear copyright wise. However, I had to add a MIT license text that _should_ include year and name of the copyright holder and I'm not sure if the way I added it now is correct. Can you take a look?

martijnbastiaan: Yeah I think it's fine

martijnbastiaan (approved):
If you can prune it a bit more that'd be great, but if not 🤷. Eventually this will be picked up by Nix anyway.

martijnbastiaan:
@lmbollen Create a folder `vendor` in the root of this repository, and just dump the whole repo in there using git subtrees. Make sure to add

```
/vendor/** linguist-vendored
```

to `.gitattributes`.

PR #222, opened 2023-01-03
Add missing shake instances to CI

lmbollen:
I noticed we have multiple shake instances that do not run on CI.
Does anyone have any thoughts about which we want to run on CI and when we want to have `ReducedPins` versions of the relevant components?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -652,10 +652,15 @@ jobs:
       matrix:
         target:
           - {top: callisto3,                   stage: netlist}
+          - {top: clockControlDemo0,           stage: netlist}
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
I think we should be a bit cautious with introducing long-running synthesis jobs to prevent frustration in the normal workflow. (Until we have Nix / proper caching of course.) Other than that, as many as possible to prevent regressions.

> Does anyone have any thoughts about which we want to run on CI and when we want to have ReducedPins versions of the relevant components?

IMO always use the reduced pins versions - this should give us the most realistic timing analysis.

martijnbastiaan (approved):
That is to say, works for me.

Issue #224, opened 2023-01-04
Make link configuration dynamic in Callisto


martijnbastiaan:
We're currently looking into populating a mining rig with FPGAs and clock multipliers to lay the foundations for a clock control experimentation platform. Ideally, experimenting with it would be hooked up to GitHub Actions. To get some reasonable test turnaround time, we don't want to synthesize over-and-over again. Thus, it would be nice to be able to configure topologies (i.e., link enable/disable) at runtime.

Related: 

 * https://github.com/bittide/bittide-hardware/issues/207. Changing the type signature of Callisto will entail many changes in the `TemplateHaskell`. The proposed method would allow us to get rid of hardcoded topologies, eliminating the need for `TemplateHaskell` in the first place.
 * https://github.com/bittide/bittide-hardware/issues/208. More aggressive caching would eliminate a lot of re-synthesis.

kleinreact:
Implemented with #226

Issue #207, opened 2022-12-14
Simplify TemplateHaskell infrastructure


martijnbastiaan:

kleinreact:
Implemented with #226.

`Bittide.Topology.TH` is still part of `elastic-buffer-sim`, but the module is not used anymore. Everything generated there is now part of `Bittide.Topology`. I kept `Bittide.Topology.TH` for easy comparison of both variants for the moment. If the PR review is sucessful, then `Bittide.Topology.TH` should be removed.

PR #226, opened 2023-01-05
Make link configuration dynamic in Callisto

kleinreact:
Intends to implement #224 and #207.


.github/workflows/ci.yml
```diff
@@ -132,7 +132,7 @@ jobs:
       - name: Plot mesh grid
         run: |
           cabal build elastic-buffer-sim:sim
-          cd elastic-buffer-sim && poetry run $(cabal list-bin elastic-buffer-sim:sim) +RTS -M8M -RTS plot complete3 150000 100
+          cd elastic-buffer-sim && poetry run $(cabal list-bin elastic-buffer-sim:sim) +RTS -M128M -RTS plot complete3 150000 100
```
leonschoorl, line 135: I had explicitly put in a tight memory limit here to guard against re-introduction of the memory leak fixed in #181.

kleinreact: This indeed was useful for detecting that there still is a memory leak. The 128M are meant to be a temporary workaround for checking that my changes are at least functionally correct. Should we keep the 128M within this PR and resolve the memory leak separately, or do you want to that fixed within this PR as well.

@lmbollen and I already tried to identify the leak on Friday afternoon, but we did not come to a solution yet.

martijnbastiaan (changes requested):
Some preliminary review. Could you update the PR title to better reflect what's happening? E.g.,

> Make link configuration dynamic in Callisto

I think the rest is a natural consequence of this.

bittide/tests/Tests/Haxioms.hs
```diff
@@ -198,3 +199,18 @@ prop_lessThanMax = property $ do
...
+-- Given:
+--
+--     1 <= m, n + m <= u
+--
+prop_useLowerLimit :: Property
+prop_useLowerLimit = property $ do
+  u <- forAll (genNatural 0)
+  n <- forAll (genNatural 0)
+  m <- forAll (genNatural 0)
+  assert (1 > m || n + m > u || n + 1 <= u)
```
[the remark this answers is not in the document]
kleinreact: Sure, why not. I am just curious that you can always rewrite every constraint this way.

elastic-buffer-sim/src/Bittide/Topology.hs
```diff
@@ -28,74 +25,277 @@ module Bittide.Topology
...
+import Bittide.ClockControl
+import Bittide.ClockControl.Callisto
+import Bittide.ClockControl.ElasticBuffer
 import Bittide.Topology.Graph
-import Bittide.Topology.TH
+import Graphics.Matplotlib.Ext
 import Bittide.Topology.TH.Domain (defBittideClockConfig)
+import Control.DeepSeq
+
+-- | 'Graph' with a name for
```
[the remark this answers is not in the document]
kleinreact: You have to ask the one who put it that way in `TH.hs`

elastic-buffer-sim/src/Bittide/Topology.hs
```diff
@@ -28,74 +25,277 @@ module Bittide.Topology
...
+data GraphAPI (n :: Nat) =
```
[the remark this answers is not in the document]
kleinreact: Just taken from `TH.hs`.

elastic-buffer-sim/src/Bittide/Topology.hs
```diff
@@ -28,74 +25,277 @@ module Bittide.Topology
...
+  Int ->
+  Int ->
+  Vec n Offset ->
+  Vec n [(Period, Period, Vec n (DataCount m))]
+simNodesFromGraph ccc GraphAPI{..} m k !offsets =
+  let
+    -- elastic buffers
+    genElasticBuffer i j
+      | available i j == high = elasticBuffer Error (clocks !! i) (clocks !! j)
+      | otherwise            = pure 0
```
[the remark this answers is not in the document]
kleinreact: Ah right. Forgot about that. 
I didn't know Hasklig before. Maybe a good time to give it a try.

elastic-buffer-sim/src/Bittide/Topology.hs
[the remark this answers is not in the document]
martijnbastiaan: Yeah let's clean it up. I'm assuming the name is for the filename it uses in the end?

bittide/tests/Tests/Haxioms.hs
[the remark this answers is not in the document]
martijnbastiaan: You could! :). I'm rewriting it this way because previously you'd (effectively) throw away a lot of test cases. I.e., we don't care about test cases whose inputs don't match the constraints anyway.

bittide/src/Bittide/ClockControl/Callisto.hs
```diff
@@ -128,14 +134,17 @@ callisto targetCount updateEveryNCycles dataCounts =
   r_k :: DSignal dom F.FromS32DefDelay Float
   r_k = F.fromS32 $ D.fromSignal $
     let
-      nBuffers = natToNum @n
+      nBuffers =
+        case useLowerLimit @n @m @32 of
+          Dict ->
+            sumTo32 <$> (map (unpack . pack) . bv2v <$> mask)
```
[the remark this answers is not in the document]
kleinreact: This won't work: `mask` is a signal.

.github/workflows/ci.yml
kleinreact: Changed it back to `8M`. The space leak got resolved.

bittide/src/Bittide/ClockControl/Callisto.hs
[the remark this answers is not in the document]
martijnbastiaan: Sure, but that's not really the point I'm making. You can combine `bitCoerce` and `popCount` for a much cleaner implementation. You can apply `fmap`/`<$>` to make it work on `Signal`s.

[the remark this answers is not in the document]
kleinreact: Is `popCount` synthesizable?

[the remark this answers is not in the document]
kleinreact: Interesting ... Still don't get the point why this is better than using the safe version instead?

elastic-buffer-sim/src/Bittide/Topology.hs
```diff
@@ -25,77 +22,278 @@ module Bittide.Topology
...
+  Index n ->
+  [(Period, Period, [DataCount m])]
+simulateMesh ccc Mesh{..} m k !offsets i =
+    P.take m
+  $ takeEveryN k
+  $ absTimes (available i)
+  $ bundle (clkSignals !! i, bundle $ ebs !! i)
+ where
+  -- elastic buffers
+  !ebs = imap ebv clocks
```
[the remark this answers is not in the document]
kleinreact: It has the same semantics as for arguments. The bound expression is evaluated to weak head normal form. This is actually neccessary for plugging the space leak.

bittide/src/Bittide/ClockControl/Callisto.hs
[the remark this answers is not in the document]
martijnbastiaan: The implementation complexity is lower in a few ways:

 * It's basically just `popCount`, so it's much easier for people reading the code to see what's going on. (As opposed to converting between vectors a few times.)
 * There's no need for a separate Haxiom
 * The simulation speed _should_ be much better as `popCount` has CPU primitives, while packing/unpacking is creating `Vec`s and all that jazz.

In any case, if you want to keep the the current implementation I'd suggest pulling it out and calling it `safePopCount`.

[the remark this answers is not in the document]
kleinreact: It is synthesizable, but it returns an `Int`, which is architecture dependent:
```
src/Bittide/ClockControl/Callisto.hs:137:18: error:
    • Could not deduce: 64 ~ 32 arising from a use of ‘bitCoerce’
      from the context: (HiddenClockResetEnable dom, KnownNat n,
                         KnownNat m, 1 <= n, 1 <= m, (n + m) <= 32)
        bound by the type signature for:
                   callisto :: forall (m :: Nat) (n :: Nat) (dom :: Domain).
                               (HiddenClockResetEnable dom, KnownNat n, KnownNat m, 1 <= n,
                                1 <= m, (n + m) <= 32) =>
                               DataCount m
                               -> Unsigned 32
                               -> Signal dom (BitVector n)
                               -> Signal dom (Vec n (DataCount m))
                               -> Signal dom SpeedChange
        at src/Bittide/ClockControl/Callisto.hs:(78,1)-(99,24)
    • In the first argument of ‘(.)’, namely ‘bitCoerce’
      In the first argument of ‘(<$>)’, namely ‘bitCoerce . popCount’
      In the expression: bitCoerce . popCount <$> mask
    |
137 |       nBuffers = bitCoerce . popCount <$> mask
    |
```
Therefore, I will use the `safePopCount` instead.

bittide/src/Bittide/ClockControl/Callisto.hs
```diff
@@ -125,17 +131,21 @@ callisto targetCount updateEveryNCycles dataCounts =
   k_i = pure 1e-11
   fStep = pure 5e-4
 
+  safePopCount :: Signal dom (BitVector n) -> Signal dom (Signed 32)
+  safePopCount v = case useLowerLimit @n @m @32 of
+    Dict -> sumTo32 <$> (map (unpack . pack) . bv2v <$> v)
```
martijnbastiaan, line 136: Let's move this function to `Bittide.ClockControl.Callisto.Util`, it might come in handy somewhere else. It's implementation can be a bit simpler too (don't lift to `Signal`, unpack straight into the `Vec` we need):

```haskell
popCountTo32 ::
  forall n .
  ( KnownNat n
  , (1 + n) <= 32
  , 1 <= n
  ) =>
  BitVector n ->
  Signed 32
popCountTo32 =
  sumTo32 . unpack @(Vec n (Unsigned 1))
```

And after thinking about it some more, I'm still very unsold on this approach. We could squash all these constraints by simply implementing it as:

```haskell
popCountTo32 :: KnownNat n => BitVector n -> Signed 32
popCountTo32 = resize . bitCoerce . popCount
```

This makes the function much more convenient to use, and more efficient to compute. It will fail for `BitVector n` where `n > 2147483648` but this size is utterly non-sensical for hardware designs. We can still document this fact of course.

elastic-buffer-sim/src/Bittide/Topology.hs
```diff
@@ -25,77 +22,282 @@ module Bittide.Topology
...
+  go !ts (v Clash.:- vs) =
+    force (izipWith (\i t (p, es) -> (t, p, filterAvailable i es)) ts v)
+      : go (force $ zipWith addFs ts $ map fst v) vs
+  filterAvailable i = catMaybes . toList . imap (asMaybe . available i)
+  asMaybe = \case
+    True  -> Just
+    False -> const Nothing
+
+-- | Push infinity to the vector element level
+pointwise :: (KnownNat n, NFData a) => [Vec n a] -> Vec n [a]
```
martijnbastiaan, line 248: Isn't this just:

```suggestion
transpose :: (KnownNat n, NFData a) => [Vec n a] -> Vec n [a]
```

martijnbastiaan: Why do we need the `NFData`?

elastic-buffer-sim/src/Bittide/Topology.hs
```diff
@@ -25,77 +22,282 @@ module Bittide.Topology
...
+ where
+  go !ts (v Clash.:- vs) =
+    force (izipWith (\i t (p, es) -> (t, p, filterAvailable i es)) ts v)
+      : go (force $ zipWith addFs ts $ map fst v) vs
+  filterAvailable i = catMaybes . toList . imap (asMaybe . available i)
+  asMaybe = \case
+    True  -> Just
+    False -> const Nothing
+
+-- | Push infinity to the vector element level
```
martijnbastiaan, line 247:
```suggestion
-- | Like 'L.transpose', but works on lists of vectors. Used in this module to 
-- (.. push infinity to the vector element level ..)
```

I'm not sure what "pushing infinite to the vector element level" means. Could you rephrase / elaborate?

elastic-buffer-sim/src/Bittide/Topology.hs
```diff
@@ -25,77 +22,282 @@ module Bittide.Topology
...
+  IO Offset
+genOffsets ClockControlConfig{cccDeviation} = do
+  offsetPpm <- randomRIO (-cccDeviation, cccDeviation)
+  pure (diffPeriod offsetPpm (clockPeriodFs @dom Proxy))
+
+matplotWrite ::
+  KnownNat n =>
+  String ->
+  Vec n Matplotlib ->
+  Vec n Matplotlib ->
```
martijnbastiaan, line 277: Document the function / arguments please

bittide-instances/src/Bittide/Instances/ClockControl.hs
```diff
@@ -22,4 +22,7 @@ callisto3 ::
   -- | Speed change requested from clock multiplier
   Signal Basic200 SpeedChange
 callisto3 clk rst ena dataCounts =
-  callistoClockControl clk rst ena config dataCounts
+  callistoClockControl clk rst ena config availableLinkMask dataCounts
+ where
+  -- all links available
+  availableLinkMask = pure $ v2bv $ map (const high) indicesI
```
martijnbastiaan, line 28:
```suggestion
  availableLinkMask = pure maxBound
```

bittide-instances/src/Bittide/Instances/MVPs.hs
```diff
@@ -40,9 +40,9 @@ genericClockControlDemo0 config clkRecovered clkControlled rstControlled drainFi
   speedChangeSticky =
     withClockResetEnable clkControlled rstControlled enableGen $
       stickyBits d15 (speedChangeToPins <$> speedChange)
-
+  availableLinkMask = pure $ v2bv $ map (const high) indicesI -- all links available
```
martijnbastiaan, line 43:
```suggestion
  availableLinkMask = pure maxBound
```

elastic-buffer-sim/src/Bittide/Topology.hs
```diff
@@ -25,77 +22,282 @@ module Bittide.Topology
...
+      ccc
+    . pure
+  -- clock signals
+  !clkSignals = extractPeriods <$> clocks
+  -- available link mask vectors
+  !masks = nVec (v2bv . nVec . avail)
+  avail x y = if available x y then high else low
+  nVec = flip map indicesI
+
+absTimes ::
```
martijnbastiaan, line 232: Could you document the function and its arguments?

martijnbastiaan (changes requested):
Looks good! Nice to 

As for our way of working: don't merge `main` into `branches`, rebase branches on `main` instead. This prevents a bunch of spurious commits reviewers/code archeologists have to dig through.

I might have commented on some leftovers from the old code, but let's fix that anyway.

bittide-instances/src/Bittide/Instances/ClockControl.hs
kleinreact: I like and dislike your suggestion at the same time. I like that its short and looks smart. What i don't like is that it assumes unsigned integer semantics for the vector. You could also assume signed integer semantics for the vector and suddenly your solution turns wrong.

In general, I guess there shouldn't be something like `min` and `max` for pure bitvectors. Do you really want to introduce the change?

elastic-buffer-sim/src/Bittide/Topology.hs
kleinreact: We don't. Its a leftover from my "plugging the space leak" efforts. I'll take it out.

bittide-instances/src/Bittide/Instances/ClockControl.hs
martijnbastiaan: Yeah, I can see your point. The documentation from `BitVector` states that `Num` operations (and I suppose all others, though that's not really documented) should be interpreted as `Unsigned`. It's a pretty wide-spread pattern within Clash, so I personally would be perfectly fine with it.

Otherwise: `pack (repeat high)` should also work.

martijnbastiaan: To your pick :)

elastic-buffer-sim/src/Bittide/Topology.hs
kleinreact: I don't like to change the name to `transpose`, since it name-clashes with the vector `transpose` and may confuse people. My alternative was `elementwise`, but it looks strange as a function name.

martijnbastiaan: Fair point. `transposeV` perhaps? Otherwise feel free to ignore.

bittide-instances/src/Bittide/Instances/ClockControl.hs
kleinreact: That one's much better.

elastic-buffer-sim/tests/Tests/Bittide/Simulate.hs
```diff
@@ -36,11 +36,12 @@ case_clockControlMaxBound = do
   let
     config = defClockConfig
     dataCounts = pure maxBound :> Nil
+    mask = pure $ v2bv $ fmap (const high) dataCounts
```
martijnbastiaan, line 39:
```suggestion
    mask = pure $ pack (repeat high)
```

elastic-buffer-sim/tests/Tests/Bittide/Simulate.hs
```diff
@@ -51,11 +52,12 @@ case_clockControlMinBound = do
   let
     config = defClockConfig
     dataCounts = pure 0 :> Nil
+    mask = pure $ v2bv $ fmap (const high) dataCounts
```
martijnbastiaan, line 55:
```suggestion
    mask = pure $ pack (repeat high)
```

type inference should do its job I think

martijnbastiaan (approved):
LGTM, two small changes requested

kleinreact:
Sure, just waiting for CI to get through this time.

bittide/src/Bittide/ClockControl/Callisto.hs
kleinreact: Seems like your `popCountTo32` is not synthesizable: https://github.com/bittide/bittide-hardware/actions/runs/3912332689/jobs/6686876838. I'll try to revert to my safeVersion to see whether that fixes the problem.

kleinreact: Yep, `sumTo32 . unpack @(Vec n (Unsigned 1))` works. Are you satisfied with keeping my `safePopCountTo32` then or do you still want to go with the non-synthesizable `popCount`?

kleinreact: As discussed, we stay with `safePopCountTo32` for the moment.

bittide-instances/src/Bittide/Instances/ClockControl.hs
```diff
@@ -22,4 +22,7 @@ callisto3 ::
   -- | Speed change requested from clock multiplier
   Signal Basic200 SpeedChange
 callisto3 clk rst ena dataCounts =
-  callistoClockControl clk rst ena config dataCounts
+  callistoClockControl clk rst ena config availableLinkMask dataCounts
+ where
+  -- all links available
+  availableLinkMask = pure $ pack $ repeat high
```
leonschoorl, line 28: (I think I overheard Martijn and/or Lucas already having a discussion already having a discussion over this last week.)

But I'm not a fan of dragging Vecs into this, since it creates terrible VHDL:

```VHDL
  \c$vec\ <= (Blaat_topEntity_types.array_of_std_logic'(0 to 9-1 =>  ('1') ));

  -- map begin
  r_map : for i in result_res_res'range generate
  begin
    result_res_res(i) <= (std_logic_vector'(0 => \c$vec\(i)));


  end generate;
  -- map end

  -- concatBitVector begin
  concatBitVectorIter_loop : for i_0 in 0 to (9 - 1) generate
    result_res(((i_0 * 1) + 1 - 1) downto (i_0 * 1)) <= std_logic_vector'(result_res_res(result_res_res'high - i_0));
  end generate;
  -- concatBitVector end

  result <= result_res;
```

Compared to something like `maxBound` or `complement 0` which just turns into a constant in VHDL.

kleinreact: See my argument in https://github.com/bittide/bittide-hardware/pull/226#discussion_r1069179350.

kleinreact: `complement 0` however does not assume Integer semantics. I'm fine with that one too.

PR #227, opened 2023-01-10
Add `shell.nix`

martijnbastiaan:
With the addition of VexRisc (see #198) we're now relying on quite a number of different tool(chains): C, Python, Haskell, Rust, Scala, and Verilator. This many toolchains are hard to keep in sync across different machines hence the introduction of Nix.

To drop to a shell with all the toolchains available, execute:

    nix-shell --pure

Drop `--pure` if you need access to system executables and libraries.

----------------------------

TODOs:

 - [x] Get Vivado running
 - [x] Modify CI to use the Nix env
 - [x] Update Docker image
   - [x] Run with `--pure` on CI
 - [x] Update README


cc10512:
Shouldn't the instructions in the intro of this PR be in a README file? Just saying :)

martijnbastiaan:
Yeah :). It's time we make it contain some useful info anyway!

martijnbastiaan:
So for the people wondering about the flurry of CI pushes; turns out this was quite the ride to get Nix working on CI (we're not fully out of the woods just yet).

1. The [Install Nix](https://github.com/marketplace/actions/install-nix) action doesn't work on docker images. We need to run with docker, because this is the only way we can do isolated CI at the moment. (If we want to go dockerless; we need to spin up / spin down VMs for every job, but this was too much work to implement correctly for yours truly.)
1. The docker [nixos/nix](https://hub.docker.com/r/nixos/nix) Docker image runs Nix just fine, but GitHub Actions expects certain binaries at certain places, basically restricting it to "normal" Linux distributions.
1. The Nix install script can't run as root, but creating a separate user within the docker image turned out to be impossible thanks to the way Docker and GitHub runner interact. (Basically it leaves files on disk it can't clean up in the next run due to user id mapping between the host and docker.)

In the end I managed to coerce the Nix install script into running as root through:

```bash
  groupadd -r nixbld && \
  for n in $(seq 1 10); do \
    useradd \
      -c "Nix build user $n"\
      -d /var/empty \
      -g nixbld \
      -G nixbld \
      -M -N -r \
      -s "$(command -v nologin)" "nixbld$n";\
  done
```

This is part of the `Dockerfile` introduced in this PR. I've published the Docker image over [over here](https://github.com/orgs/clash-lang/packages/container/package/nixos-bittide-hardware). Ideally this would be part of the `bittide` organization, but GitHub doesn't allow repository specific docker images anymore introducing a bunch of paperwork if we wanted to do it properly. Given that the `Dockerfile` is there for everyone to look at I'd say this is the pragmatic solution..

.github/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+FROM ubuntu:22.04
+
+RUN \
+  apt-get update && \
+  apt-get install -y curl xz-utils sudo git && \
+  rm -rf /var/lib/apt/lists/*
+
+RUN \
+  groupadd -r nixbld && \
+  for n in $(seq 1 10); do \
```
kleinreact, line 13: Why do we need 10 users for the docker? I guess its for different build jobs, but is there some reason that they cannot be executed by the same user instead?

.github/workflows/ci.yml
```diff
@@ -424,59 +421,39 @@ jobs:
   firmware-lints:
     name: Firmware Lints
     runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: git-nix-shell {0}
+    needs: [build, lint]
```
cuddlefishie, line 427: the `build` task is for building Haskell stuff only, right? This task doesn't depend on that semantically, firmware linting can potentially also be much faster so it might make sense to remove this dependency to make CI checks faster when working on firmware stuff?

shell.nix
```diff
@@ -0,0 +1,58 @@
...
+
+      # Rust toolchain. Contains:
+      #
+      #  * rust-analyzer
+      #  * rustc
+      #  * rustdoc
+      #  * rustfmt
+      #  * (.. and more)
+      #
+      ((pkgs.rustChannelOf { channel = "1.65"; }).rust.override {
```
cuddlefishie, line 36: 1.67 works too, might as well use that now :)
```suggestion
      ((pkgs.rustChannelOf { channel = "1.67"; }).rust.override {
```

shell.nix
```diff
@@ -0,0 +1,58 @@
...
+      pkgs.buildPackages.scala
+      pkgs.buildPackages.verilator
+
+      # Rust toolchain. Contains:
+      #
+      #  * rust-analyzer
+      #  * rustc
+      #  * rustdoc
+      #  * rustfmt
+      #  * (.. and more)
```
cuddlefishie, line 34: This doesn't mention things like clippy or MIRI (we should probably look into MIRI tests on CI too :eyes: ), is this just the default installation config? Should we have an explicit component list here too maybe?

cuddlefishie (commented):
This looks good so far, from my limited Nix and CI knowledge! :sweat_smile: Just left a few questions/comments

shell.nix
```diff
@@ -0,0 +1,58 @@
...
+pkgs.mkShell {
+  name = "shell";
+  buildInputs =
+    # Ideally, we'd add all (Hackage) dependencies to the dependency list so we
+    # don't have to compile everything ourselves. I haven't figured out how to
+    # do this yet without introducing conflicts with `cabal.project`.
+    #
+    # pkgs.haskellPackages.bittide-extra.env.nativeBuildInputs ++
+    #
+    [
```
kleinreact, line 15: If I want to use this locally with a fresh nix install, then `pkgs.git` is required for me as well. Otherwise I get
```
Error: cabal: The program 'git' is required but it could not be found.
```
when running `cabal build`.

.github/workflows/ci.yml
```diff
@@ -114,47 +113,34 @@ jobs:
...
-          cd elastic-buffer-sim && poetry run $(cabal list-bin elastic-buffer-sim:sim) +RTS -M8M -RTS plot complete3 150000 100
+          cabal run -- elastic-buffer-sim:sim +RTS -M8M -RTS plot complete3 150000 100
 
       - name: Upload plots
         uses: actions/upload-artifact@v2
         with:
           name: gen-plots-hs
           path: |
-            elastic-buffer-sim/_build/clockscomplete3.pdf
-            elastic-buffer-sim/_build/elasticbufferscomplete3.pdf
```
kleinreact, line 149: Maybe I am missing someting here, but where do we enter the `elastic-buffer-sim` folder now. Since I cannot see any artifacts created by the CI Actions at the moment, this does not look like its working to me.

kleinreact: Ok, the artifacts are there, I just was not on the summary view in github. Still wondering about the folder tough.

kleinreact (changes requested):
Looks solid to me and also works fine (except for one small addition -> see my comment on `shell.nix`) on my local system. The `ci.yml` still has some redundancies, but nothing that's out of scope for a human reader. The readme should cover the nix-specific settings of `ci.yml` for future updates, such that users not familiar with nix can still handle updates.

.github/docker/Dockerfile
martijnbastiaan: Good question. A forum thread suggested this, I'm not sure if the `10` carries any significance. I'll experiment.

shell.nix
martijnbastiaan: Yeah that sounds nice! Are you talking about [this](https://github.com/rust-lang/miri) MIRI?

shell.nix
martijnbastiaan: I'll add it because I think CI needs it too when running with `--pure`, but for developers _not_ using `--pure` is perfectly fine.

.github/workflows/ci.yml
```diff
@@ -59,9 +52,12 @@ jobs:
   build:
     name: Build dependencies
     runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: git-nix-shell {0}
 
     container:
-      image: ghcr.io/clash-lang/clash-ci-9.0.2:2022-12-13
+      image: ghcr.io/clash-lang/nixos-bittide-hardware:2023-01-25
```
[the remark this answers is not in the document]
martijnbastiaan: Yeah I don't know how :(. The repetition is terrible in `ci.yml` but there seems to be no way to avoid it.

.github/workflows/ci.yml
martijnbastiaan: Ah good point, I moved some thing around. I thought I reverted everything but I guess not!

.github/workflows/ci.yml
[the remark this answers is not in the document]
kleinreact: The yaml standard supports [anchors](https://yaml.org/spec/1.2.2/#692-node-anchors), which is the feature that has been especially designed for reducing redundancy. Some introduction to the topic can be found [here](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/). So you don't even need any Github features at this point, its supposed to work with every yaml configuration file.

.github/workflows/ci.yml
kleinreact: Are the artifacts than taken from the cache? Still wondering why this has not caused any error then?

.github/workflows/ci.yml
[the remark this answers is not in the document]
martijnbastiaan: YAML anchors aren't supported by GitHub Actions unfortunately.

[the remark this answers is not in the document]
martijnbastiaan: I guess we could have some anchored files somewhere else and manually convert them to GitHub Action's YAML. (And make CI check that these files are synchronized, of course.)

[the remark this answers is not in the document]
kleinreact: Ok thats sad, just read the [official issue](https://github.com/actions/runner/issues/1182) on that, which definitly downgrades my opinion about Github CI Actions. 

I propose keeping everything as it is then. It does not make sense to work around non-existing features. Either they are supposed to be introduced upstream at some or you need to switch the platform. That's how it always works.

shell.nix
cuddlefishie: Yep, that MIRI! It can test (a limited but always growing subset of) unsafe Rust code for undefined behaviour or safety violations! With all the firmware code we have that also uses unsafe, it could be nice to have some checks at some point. But that's for another PR :)

.github/workflows/ci.yml
martijnbastiaan: Sorry I mislead. Above these diff lines there's:

```diff
-          cabal build elastic-buffer-sim:sim
-          cd elastic-buffer-sim && poetry run $(cabal list-bin elastic-buffer-sim:sim) +RTS -M8M -RTS plot complete3 150000 100
+          cabal run -- elastic-buffer-sim:sim +RTS -M8M -RTS plot complete3 150000 100
```

An improvement IMO, but probably not something that should be part of this PR.

.github/docker/Dockerfile
martijnbastiaan: The documentation says:

> This creates 10 build users. There can never be more concurrent builds than the number of build users, so you may want to increase this if you expect to do many builds at the same time.

So while 10 is a bit arbitrary, I doubt we'll need more than this (and it doesn't hurt to have too many). Agreed?

shell.nix
martijnbastiaan: I didn't actually answer your question :eyes:

> is this just the default installation config

Not 100% sure. Is Clippy installed by default? If it is it might be less confusing to just leave out the comment :-).

> Should we have an explicit component list here too maybe?

I'm not sure how. There is an `extensions` option, which can contain:

```
       rust-src
       rustc-dev
       rustc-docs
       rust-std
       rust-analysis
       rls-preview
       rust-analyzer-preview
       clippy-preview
       rustfmt-preview
       llvm-tools-preview
       miri-preview
       rust-docs-json-preview
```

..which doesn't seem to match `components`.

martijnbastiaan:
Thanks for the comments both of you. I've downgraded the Rust version to 1.65, as 1.67 is giving me troubles while running `cargo clippy --target riscv32imc-unknown-none-elf --all-features`:

```
error: language item required, but not found: `eh_personality`
  |
  = note: this can occur when a binary crate with `#![no_std]` is compiled for a target where `eh_personality` is defined in the standard library
  = help: you may be able to compile for a target that doesn't need `eh_personality`, specify a target with `--target` or in `.cargo/config`

error: could not compile `vex-test-programs` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `vex-test-programs` due to previous error
error: could not compile `vex-test-programs` due to previous error
error: could not compile `vex-test-programs` due to previous error
error: could not compile `vex-test-programs` due to previous error
```

This is probably a Nix packaging issue - I'll try and make a minimal example and report it at [nixpkgs-mozilla/issues](https://github.com/mozilla/nixpkgs-mozilla/issues).

shell.nix
cuddlefishie: Since it seems to run all CI tasks successfully it seems like this should be good enough for now!

> Is Clippy installed by default?

There's different profiles, I think the `minimal` one does not come with clippy, but I think the default one does, that's why it all works?

I think the comment can stay, but it confused me a bit more than just assuming it's the default installation :sweat_smile: But I'm fine with this as it is!

README.md
```diff
@@ -4,9 +4,26 @@ SPDX-FileCopyrightText: 2022 Google LLC
...
+to manage this. To get a development shell [install nix](https://nixos.org/download.html)
+and run the following command in the root of this repository:
+
+```
+nix-shell
+```
+
+The started shell contains everything needed to develop Bittide components.
+
+TODO: Add overview of components
```
martijnbastiaan, line 23: I wasn't really planning on doing this in this PR :sweat_smile:

cuddlefishie (approved):
From my (non-expert) side this looks good!

.github/workflows/ci.yml
kleinreact: Ok, the `_build` folder also gets created in the root folder. That was what I just did not recognice before. Thats fine then. These calls will be updated soon with the new simulation checks anyway.

kleinreact (approved):
Looks all good to me.

PR #229, opened 2023-01-13
Bump CI action versions

martijnbastiaan:
[Use actions/checkout@v3 on CI](https://github.com/bittide/bittide-hardware/commit/b39b1442e4eefb46982cfcd2789ec8a72bfd3258):

Way back we decided to use v1, because it was the only version that
properly supported git submodules. We don't use submodules anymore
though, so we can freely upgrade - removing quite a few warnings
from CI.

----------------

[Use actions/cache@v3 on CI](https://github.com/bittide/bittide-hardware/commit/9e332f70e5135ba9af1070c3233eec3b7ea4cb39):

Reduces warning noise

### DIFF
--- a/.github/scripts/set_de_mirror.sh
+++ b/.github/scripts/set_de_mirror.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2022 Google LLC
+#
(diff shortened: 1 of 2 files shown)


cuddlefishie (approved):
Looks good to me!

PR #231, opened 2023-01-19
popCount Synthesis Test

kleinreact:
This branch is only for tracing down what happened in https://github.com/bittide/bittide-hardware/actions/runs/3912332689/jobs/6686876838. 

Synthesis of `popCountTo32` works without any problems in a local setup with `clash-1.6.3` and `yosys-0.24` as an alternative backend. Therefore, the problem seems to be very context specific and is further analyzed in the scope of the bittide-hardware project instead.

The error has been reduced to the following piece of code:
https://github.com/bittide/bittide-hardware/blob/popcount-test/bittide/src/Bittide/PopCountTest.hs

Additional findings of the error analysis are given in the comments. The error produced by this piece of code is printed here:
https://github.com/bittide/bittide-hardware/actions/runs/3984846590/jobs/6831556331

_(Note: the introduced simplifications can easily be compared to the original erroneous code, which is still executed by the other job of the respective synthesis action.)_


martijnbastiaan:
Oh sorry, I meant try to make a small reproducer and report it over at the clash-compiler issue tracker.

kleinreact:
> Oh sorry, I meant try to make a small reproducer and report it over at the clash-compiler issue tracker.

I already tried that, but I could not reproduce the error this way. This is why I started to strip down the error within this setup to have at least some basis for further investigations.

martijnbastiaan:
Alright, makes sense!

PR #233, opened 2023-01-19
commit Cargo.lock file for the workspace

cuddlefishie:
In the past binaries were not part of the workspace and committed their own lock files into version control. Later all crates were added to the workspace to make Rust tools work better (rust-analyzer, clippy, etc) but that also merges all Cargo.lock files into one - the workspace lock file. This file was ignored. This caused problems in the past already when dependencies were yanked from crates.io.

This should make the build process more robust and reproducible.


martijnbastiaan (approved):
yesss

PR #234, opened 2023-01-20
Run CI on every push and branch

martijnbastiaan:
PR trigger seem to be broken at the moment, though triggering on pushes still works. Given that all our development happens in PRs anyway, there is no functional difference between running on every push or running on PR updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,14 +3,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 name: CI
-on:
-  push:
-    branches:
-      - main
-  pull_request:
-    types:
-      - opened
-      - synchronize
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -5,12 +5,6 @@
 name: CI
 on:
```
kleinreact, line 6: `on: [push]` may be more compact, but semantically this is already what we want.

kleinreact (approved):
Looks good to me. Its up to you whether we should go with the compact notation or stay with this one for the moment.

.github/workflows/ci.yml
martijnbastiaan: Done

PR #245, opened 2023-01-30
Add stability checker to the simulation

kleinreact:
Adds the following new features:
* Stability checking of the elastic buffers, where
  * the stability is highlighted within the plots, and
  * the return value captures the stability of all buffers at the end of the simulation.
* Running the simulation is now separated from the simulation entity (see [`Bittide.Toplogy`](https://github.com/bittide/bittide-hardware/blob/stability-check/elastic-buffer-sim/src/Bittide/Topology.hs)), where the entity captures all synthesizable parts of the simulation architecture.
* More CLI options have been added (use `--help` for details), featuring
  * topology generation via CLI arguments,
  * reproduction of the experiment from a generated JSON file and DOT graph

---------------------

TODOs:

- [x] fix CI build problem
- [x] add the random offsets to the `simulation.json`
- [x] rebase after the nix branch has been merged
- [x] add a CLI flag for stopping simulation as soon as all buffers are stable
- [x] fix the problem of the slowdown
- [x] integrate with `staging`: #247

Other "Nice-to-Have"s:

- [x] _(make CLI arguments optional, if passed via JSON file)_
- [x] _(support passing topologies as files)_
- [x] _(support random topology generation)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,13 +100,41 @@ jobs:
         run: |
           cabal build all --only-dependencies
 
+  elastic-buffer-sim-topologies-matrix:
+    name: elastic-buffer-sim-topologies simulation matrix generation
+    runs-on: self-hosted
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Generate matrix
+        run: |
+          if [[ "${{ github.event_name }}" == "schedule" || $(.github/scripts/force_expensive_checks.sh) == "true" ]]; then
+            cp .github/simulation/all.json sim.json
+          else
+            cp .github/simulation/staging.json sim.json
+          fi
+
+      - name: Set simulation matrix
+        id: set-sim-matrix
+        run: |
+          echo "sim_matrix=$(cat sim.json | tr '\n' ' ')" | tee -a "$GITHUB_OUTPUT"
+
+    outputs:
+      sim_matrix: ${{ steps.set-sim-matrix.outputs.sim_matrix }}
+
   elastic-buffer-sim-topologies:
     name: Simulate network
     runs-on: self-hosted
     defaults:
       run:
         shell: git-nix-shell {0} --pure
-    needs: [build, lint]
+    needs: [build, lint, elastic-buffer-sim-topologies-matrix]
+
+    strategy:
+      matrix:
+        target: ${{ fromJson(needs.elastic-buffer-sim-topologies-matrix.outputs.sim_matrix) }}
+      fail-fast: false
 
     container:
       image: ghcr.io/clash-lang/nixos-bittide-hardware:2023-02-01
@@ -134,18 +162,21 @@ jobs:
           restore-keys: packages-cachebust-3-
           fail-on-cache-miss: true
 
-      # TODO: simulate + plot multiple topologies
       - name: Plot mesh grid
         run: |
-          cabal run -- elastic-buffer-sim:sim +RTS -M8M -RTS plot complete3 150000 100
+          cabal run -- elastic-buffer-sim:sim +RTS -M16M -RTS \
+            --output-mode pdf \
+            --steps ${{ matrix.target.steps }} \
+            --samples 1000 \
+            --stop-when-stable \
+            ${{ matrix.target.topology }} ${{ matrix.target.args }}
 
       - name: Upload plots
         uses: actions/upload-artifact@v3
         with:
           name: gen-plots-hs
-          path: |
-            _build/clockscomplete3.pdf
-            _build/elasticbufferscomplete3.pdf
+          path: _build
+          retention-days: 14
 
   elastic-buffer-sim-tests:
     name: elastic-buffer-sim unittests
@@ -828,6 +859,7 @@ jobs:
         contranomy-sim-tests,
         contranomy-tests,
         elastic-buffer-sim-tests,
+        elastic-buffer-sim-topologies-matrix,
         elastic-buffer-sim-topologies,
         firmware-build-examples,
         firmware-build-integration-tests,
--- a/elastic-buffer-sim/exe/Main.hs
+++ b/elastic-buffer-sim/exe/Main.hs
@@ -2,70 +2,670 @@
 --
 -- SPDX-License-Identifier: Apache-2.0
 
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE OverloadedStrings #-}
+
 module Main (main) where
 
-import Control.Monad (when)
-import Data.Proxy (Proxy (..))
-import Data.Typeable (Typeable, typeRep)
-import System.Console.Docopt
-import System.Environment (getArgs)
-import Text.Read (readMaybe)
+import Data.Aeson
+  ( ToJSON(..)
+  , FromJSON(..)
+  , Value(..)
+  , (.=)
+  , (.:)
+  , encode
+  , decode
+  , object
+  )
+import Data.Aeson.Types (typeMismatch)
+import Control.Monad (forM, forM_, when, replicateM, replicateM_)
+import GHC.Generics (Generic)
+import GHC.Int (Int64)
+import System.Exit (exitSuccess, exitFailure, die)
+import System.FilePath ((</>))
+import System.Directory (createDirectoryIfMissing)
+import Data.ByteString.Lazy qualified as BS
+import Data.Set qualified as S (fromList, toList)
+import Data.Map.Strict qualified as M ((!), fromList)
+import Data.Maybe (catMaybes)
+import Data.Graph (Graph, buildG, edges, scc)
+import Data.Array.MArray (newListArray, readArray, writeArray, getElems, freeze)
+import Data.Array.IO (IOUArray)
+import Data.Array qualified as A ((!))
+import System.Random (randomIO, randomRIO)
 
-import Bittide.Topology
+import Language.Dot.Graph
+import Language.Dot.Pretty (render)
+import Language.Dot.Parser (parse)
 
-readOrError :: forall a. (Typeable a, Read a) => String -> a
-readOrError s =
-  case readMaybe s of
-    Nothing -> error ("Could not parse '" <> s <> "' as '" ++ show (typeRep (Proxy @a)) ++ "'")
-    Just a  -> a
+import Options.Applicative
+import Options.Applicative.Help.Pretty (text)
 
-patterns :: Docopt
-patterns = [docopt|
-sim version 0.1.0
+import Bittide.Plot
 
-Usage:
-  sim csv <steps> <points>
-  sim plot complete2 <steps> <points>
-  sim plot complete3 <steps> <points>
-  sim plot complete6 <steps> <points>
-  sim plot diamond <steps> <points>
-  sim plot star7 <steps> <points>
-  sim plot tree32 <steps> <points>
-  sim plot tree23 <steps> <points>
-  sim plot hypercube3 <steps> <points>
-  sim plot hypercube4 <steps> <points>
+data Topology =
+    Diamond
+  | Line Int
+  | HyperCube Int
+  | Grid Int Int
+  | Torus2D Int Int
+  | Torus3D Int Int Int
+  | Tree Int Int
+  | Star Int
+  | Cycle Int
+  | Complete Int
+  | DotFile FilePath
+  | Random Int
+  deriving (Show, Ord, Eq)
 
-Options:
-  <points> Number of points to keep + pass to plotting library
-|]
+ttype :: Topology -> String
+ttype = \case
+  Diamond{}   -> "diamond"
+  Line{}      -> "line"
+  HyperCube{} -> "hypercube"
+  Grid{}      -> "grid"
+  Torus2D{}   -> "torus2d"
+  Torus3D{}   -> "torus3d"
+  Tree{}      -> "tree"
+  Star{}      -> "star"
+  Cycle{}     -> "cycle"
+  Complete{}  -> "complete"
+  DotFile{}   -> "dotfile"
+  Random{}    -> "random"
 
-getArgOrExit :: Arguments -> Option -> IO String
-getArgOrExit = getArgOrExitWith patterns
+instance ToJSON Topology where
+  toJSON t = object $ case t of
+    Diamond       -> [ gt ]
+    DotFile f     -> [ gt, "filepath"   .= f ]
+    Line n        -> [ gt, "nodes"      .= n ]
+    HyperCube n   -> [ gt, "dimensions" .= n ]
+    Star n        -> [ gt, "nodes"      .= n ]
+    Cycle n       -> [ gt, "nodes"      .= n ]
+    Complete n    -> [ gt, "nodes"      .= n ]
+    Random n      -> [ gt, "nodes"      .= n ]
+    Tree d c      -> [ gt, "depth"      .= d, "childs" .= c ]
+    Grid r c      -> [ gt, "rows"       .= r, "cols"   .= c ]
+    Torus2D r c   -> [ gt, "rows"       .= r, "cols"   .= c ]
+    Torus3D r c p -> [ gt, "rows"       .= r, "cols"   .= c, "planes" .= p ]
+   where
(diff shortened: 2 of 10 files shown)


martijnbastiaan (changes requested):
This is great!

A few questions / remarks:

 * Why are we doing zipping in Haskell? It seems like this would just be a matter of running `zip plots.zip *.pdf` on CI.
 * Would it be possible to stop simulation early if all the elastic buffers are stable? That way we could save a bit on CI resources.
 * Having the ability to pass the random offsets would be great. That way we could emit an error message along the lines of "check failed, rerun locally with ....".

Apart from the two comments I made separate from this message I'd be fine with merging as is btw.

kleinreact:
> Why are we doing zipping in Haskell? It seems like this would just be a matter of running zip plots.zip *.pdf on CI.

Zipping in Haskell has several advantages:
* Running a shell command in Haskell instead is even more complicated, since you need to handle `stdin`, `stdout`, etc. (if you want to do it correctly), which introduces more code than just zipping everything with the `zip-archive` library.
* The `zip` command is not always present by default in many docker images or other environments, meaning it must be explicitly installed. This way, we don't care.
* Putting results directly in a zip avoids the need for creating arbitrary `_build` folders somewhere in the system the user might not be aware of.

> Would it be possible to stop simulation early if all the elastic buffers are stable? That way we could save a bit on CI resources.

Sure, I will add a CLI flag for that.

> Having the ability to pass the random offsets would be great. That way we could emit an error message along the lines of "check failed, rerun locally with ....".

I will put it to the TODOs then.

elastic-buffer-sim/src/Bittide/Plot.hs
```diff
@@ -0,0 +1,475 @@
...
+data SimulationSettings =
+  SimulationSettings
+    { margin    :: Int
+    , framesize :: Int
+    , points    :: Int
+    , steps     :: Int
+    , mode      :: OutputMode
+    , tmpDir    :: FilePath
+    , name      :: String
+    }
```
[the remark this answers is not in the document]
kleinreact: I totally agree with your suggestion, but there is a reason for the repetitions of the `plot*` functions. 

The problem is that we impose a constraint on the size of the topologies, which is, however, calculated differently from dependent type parameters for every topology type in `Bittide.Topology.Graph`. This actually kept me busy a while to get it working with existential types. The trick is to encapsulate the graph size into an existentially quantified `SomeGraph` first and then verify the size constraints inside that existential scope afterwards. Thats at least the solution I found. 

The drawback is that the [`SomeGraph`](https://github.com/bittide/bittide-hardware/blob/stability-check/elastic-buffer-sim/src/Bittide/Plot.hs#L123) application in the `plot*` functions is of a different type for every instance, which is why you can't join them into a single function call. Have a look at [`simulatableG`](https://github.com/bittide/bittide-hardware/blob/stability-check/elastic-buffer-sim/src/Bittide/Plot.hs#L438) to get a better picture what's actually happening there.

If you are aware of any better solution than this, please let me know. That's the cleanest one I could get to so far.

elastic-buffer-sim/src/Bittide/Plot.hs
```diff
@@ -0,0 +1,475 @@
...
+plotDiamond settings@SimulationSettings{..} =
+  case ( someNat margin
+       , somePositiveNat framesize
+       ) of
+    (  Just (SomeNat (_ :: Proxy margin))
+     , Just (SomePositiveNat (_ :: Proxy framesize))
+     ) ->
+       simulateTopology defBittideClockConfig
+         (SNat @margin) (SNat @framesize) diamond settings
+    (x, y) -> invalidArgs (Just undefined, x, y)
```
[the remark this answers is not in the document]
kleinreact: Yeah, it's ok, because `invalidArgs` only checks for `Nothing`s in the tuple, but you are right, we should get rid of the `Maybe` context first and only pass the positions to the `invalidArgs` instead.

martijnbastiaan:
> Putting results directly in a zip avoids the need for creating arbitrary _build folders somewhere in the system the user might not be aware of.

There are two uses of this program:

 * To view dynamic behavior of the clock algorithm (i.e., view plots)
 * To run on CI and check for stability

That is to say, for the first use case the whole point is writing the plots somewhere. I think the default should be like it was: write the pdfs to `_build/{topology}_{params}.pdf`. I'd hate to find out I forgot to add a flag after an hour of simulation.. It's fine for the binary to emit where it wrote them / make the user aware. (I'm ambivalent about writing the JSONs there too.) 

As for zipping: while it does save on a system binary (`zip`) it relies on another (`pkg-config`). I wasn't suggesting spawning a subprocess to run `zip`, I was trying to suggest just leaving it out entirely :-). If people want to zip (or tar or ...) they can do it themselves afterwards IMO.

elastic-buffer-sim/src/Bittide/Plot.hs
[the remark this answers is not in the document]
martijnbastiaan: That makes sense. Thanks!

.github/workflows/ci.yml
```diff
@@ -122,15 +122,17 @@ jobs:
       # TODO: simulate + plot multiple topologies
       - name: Plot mesh grid
         run: |
-          cabal run -- elastic-buffer-sim:sim +RTS -M8M -RTS plot complete3 150000 100
+          cabal run -- elastic-buffer-sim:sim +RTS -M8M -RTS --output-mode pdf complete --nodes 3 --steps 150000 --samples 100
```
kleinreact, line 125: This still has to be replaced by proper simulation entries. The `sim` executable terminates with an error code now, if the buffers have not stabilized at end of the simulatino. This is why CI still fails at the moment.

elastic-buffer-sim/exe/Main.hs
```diff
@@ -2,70 +2,644 @@
...
+  fromDotName = \case
+    StringID x -> x
+    XMLID x    -> x
+
+unchain :: [a] -> [(a,a)]
+unchain = unchain' []
+ where
+  unchain' a = \case
+    x:y:xr -> unchain' ((x,y):a) (y:xr)
+    _      -> a
```
martijnbastiaan, line 645:
```suggestion
unchain :: [a] -> [(a,a)]
unchain as = zip as (tail as)
```

?

Perhaps:

```suggestion
-- | Successive overlapping pairs.
-- 
-- >>> pairwise [1, 2, 3, 4]
-- [(1,2), (2, 3), (3, 4), (4, 5)]
-- >>> pairwise []
-- []
pairwise :: [a] -> [(a,a)]
pairwise as = zip as (tail as)
```

Name inspired by Python's `itertools`.

elastic-buffer-sim/exe/Main.hs
```diff
@@ -2,70 +2,644 @@
...
+          Subgraph{}  -> Left "Subgraphs are not supported"
+        _ -> pure Nothing
+
+  let
+    edgeNames = strictSort $ concatMap (\(x, y) -> [x, y]) namedEdges
+    n = length edgeNames - 1
+    idx = (M.!) $ M.fromList $ zip edgeNames [0,1..]
+    g = buildG (0,n)
+      $ filter (uncurry (/=))
+      $ strictSort
```
martijnbastiaan, line 620: I'm not sure why we need this. Do we leak memory without this?

elastic-buffer-sim/exe/Main.hs
```diff
@@ -2,70 +2,644 @@
...
+    , available A.! (i, j)
+    ]
+
+toDot :: Graph -> String -> (Bool, GraphType, Maybe Name, [Statement])
+toDot g name = (True, Graph, Just $ StringID name, map asEdgeStatement $ edges g)
+ where
+  asEdgeStatement (x, y) =
+    EdgeStatement (map (\i -> NodeRef (XMLID ('n' : show i)) Nothing) [x,y]) []
+
+fromDot :: String -> Either String (Graph, String)
```
martijnbastiaan, line 596: Could you document the arguments/results of this function? Could you motivate its existence? I'm not sure when we would need to _read_ a dot file.

elastic-buffer-sim/exe/Main.hs
```diff
@@ -2,70 +2,644 @@
...
+  namedEdges <-
+    fmap
+      ( concatMap (map (\(x,y) -> (asString x, asString y)) . unchain)
+      . catMaybes
+      )
+    $ forM statements $ \case
+        EdgeStatement xs _ -> fmap Just $ forM xs $ \case
+          NodeRef n _ -> return n
+          Subgraph{}  -> Left "Subgraphs are not supported"
+        _ -> pure Nothing
```
martijnbastiaan, line 612: Could you pull apart these two phases into two variables and document what's happening? I can't quite follow all the monads/forM without concentrating. Seems like:

1. You're extracting all edges
2. Edges can consist of multiple nodes in dot, so you're decomposing these edges into "2-node" edges.

So I think I'd split this up like:

```haskell
-- Extract all edges from dot file. Note that dot edges can consist of more
-- than 2 nodes. We later flatten this to 2-node edges.
mutltiEdges <-
  fmap catMaybes $
    forM statements $ \case
      ..

let 
  edges = concatMap pairwise multiEdges
  namedEdges = map (bimap asString asString) twoEdges
```

The intermediate variable names should help the reader understand your code. (Not saying I picked the best names, or that this code even compiles.)

elastic-buffer-sim/exe/Main.hs
```diff
@@ -2,70 +2,644 @@
...
+      $ concatMap (\(x, y) -> [(idx x, idx y), (idx y, idx x)])
+        namedEdges
+
+  when (length (scc g) > 1) $ Left "Graph must be strongly connected"
+
+  return (g, name)
+
+ where
+  strictSort :: Ord a => [a] -> [a]
+  strictSort = S.toList . S.fromList
```
martijnbastiaan, line 630:
```suggestion
  strictSort :: Ord a => [a] -> [a]
  strictSort = S.toList . S.fromList
```

This doesn't _just_ sort, it also de-duplicates. This should be reflected in the name. (Do we want to de-duplicate? If so, it's probably clearer to just use a separate `nubOrd` step. This is more expensive, but this code isn't really performance sensitive. Code clarity >> performance for non-performance critical code.)

elastic-buffer-sim/exe/Main.hs
```diff
@@ -2,70 +2,644 @@
...
+          NodeRef n _ -> return n
+          Subgraph{}  -> Left "Subgraphs are not supported"
+        _ -> pure Nothing
+
+  let
+    edgeNames = strictSort $ concatMap (\(x, y) -> [x, y]) namedEdges
+    n = length edgeNames - 1
+    idx = (M.!) $ M.fromList $ zip edgeNames [0,1..]
+    g = buildG (0,n)
+      $ filter (uncurry (/=))
```
martijnbastiaan, line 619: Does it make sense to silently drop self-loops? Isn't this an input error?

martijnbastiaan (changes requested):
LGTM in general, I've added some code clarification requests.

elastic-buffer-sim/exe/Main.hs
kleinreact: Thats a much nicer one. Updated it.

elastic-buffer-sim/exe/Main.hs
kleinreact: Thats because `buildG` perserves multiple edges between the same nodes.
```
ghci> buildG (0,2) [(1,2),(1,2),(1,2),(1,2)]
array (0,2) [(0,[]),(1,[2,2,2,2]),(2,[])]
```
I guess thats a feature if you need those within your graphs. Topologies are not mulit-edged, i.e., strict graphs (cf. [graphviz dot](https://graphviz.org/pdf/dot.1.pdf), p. 2). Therefore, we should perserve that property at this point as well.

elastic-buffer-sim/exe/Main.hs
kleinreact: > This doesn't just sort, it also de-duplicates.

This is what the `strict` in front of `sort` is meant to indicate. `strictSort` was a very common term during my FP courses, so it feels very natural to me. However, just searching for it revealed that I maybe got a little bit biased from my university context there.

>Do we want to de-duplicate?

Technically, we don't. The dot specification disallows duplicate edges when using the `strict` keyword, which is required for topologies. On the other hand, it is user input, thus I never trust it and the dot reading library does not check for it either. Furthermore, it's totally legal to delete duplicates form the graph at application side, given to the `stirct` keyword. So why not, it doesn't hurt at least.

> If so, it's probably clearer to just use a separate nubOrd step. ... Code clarity >> performance for non-performance critical code.

I actually just had to look up `nubOrd` first to get what it is doing. Therefore, `S.toList . S.fromList` is much clearer in the that sense to me, it has the same computational cost and even gets the list sorted, which improves readability of your data. Sets also enjoy longer support than `nubOrd` in `containers`. So that's the clear winner for me at this point ;-).

elastic-buffer-sim/exe/Main.hs
kleinreact: That depends on what you consider the selfloop to express in terms of a bittide topology. In terms of connectivity, a node can synchronize its clock with itself even without a elastic buffer in between. Therefore, for me graphs with self-loops as input are just fine. The self-loops just don't have any effect on the topology that is created out of it.

elastic-buffer-sim/exe/Main.hs
kleinreact: > Could you document the arguments/results of this function?

Sure, I'll add some documentation. 

> Could you motivate its existence?

Its primarily there for reproducabilty. Each simulated topology is also written to `_build` as dot graph. Hence, we want to be able to read it back for repeating a run on exactly the same input. The `--json-args` feature also makes use of that. Note that we also support randomly generated topologies by now. Thus, if an interesting outcome is produced by that at some point, we wouldn't have any chance to reproduce that result otherwise.

elastic-buffer-sim/exe/Main.hs
kleinreact: Good point. I'll update it.

elastic-buffer-sim/exe/Main.hs
martijnbastiaan: That makes sense, add that as a comment.

elastic-buffer-sim/exe/Main.hs
martijnbastiaan: Alright, that makes sense. That puts it in the list of "strictly ascending". However, as you can see, there is ambiguity here: "strict" usually refers to the-opposite-of-lazy, which is the way I read it. I didn't consider it could mean "deduplicate and sort". Which brings me too:

> I actually just had to look up `nubOrd` first to get what it is doing.

This type of confusion is fine (within reason of course, no need to drag in category theory nomenclature everywhere). It's the leading astray type of confusion we should strive to prevent. So let's remove the ambiguity by calling it something along the lines of:

```haskell
nubAndSort   -- combines 'nub' and 'sort' from 'Prelude'
dedupAndSort -- replaces 'nub' by a more sensible word
```

> Sets also enjoy longer support than nubOrd in containers.

I'm assuming you're implying that this therefore has a higher chance of being recognized by Haskell engineers. I don't think so; `nub` has been part of the Haskell98 report. `nubOrd` is its better performing cousin.

Personally would write it as:

```
 $ sort   -- for pretty printing
 $ nubOrd -- remove duplicate edges
 $ ..
```

but

```
  $ dedupAndSort -- remove duplicate edges, sort for pretty printing
``` 

is fine too. In any case, explain the choices, I wasn't sure why things were the way the were.

elastic-buffer-sim/exe/Main.hs
kleinreact: Done.

elastic-buffer-sim/exe/Main.hs
martijnbastiaan: Right, I was thinking the `strict` referred to some kind of non-lazy behavior.

elastic-buffer-sim/exe/Main.hs
martijnbastiaan: Ok!

elastic-buffer-sim/exe/Main.hs
kleinreact: Done.

martijnbastiaan (approved):
Nice work

PR #247, opened 2023-02-02
Only run expensive CI during nightlies and on 'staging'

martijnbastiaan:
We now request to merge changes into 'staging'. In order to do so, relatively
cheap tests should pass CI. Every night, CI runs our full test suite on
'staging'. If these pass, a pull request may be created to get 'main' up to
date with 'staging'. Note that this should be automated in the future, but
this will require some engineering.
    
We currently skip the following tests on PRs:
    
  * RiscV formal tests
  * All but one synthesis run
    
Note that running only lightweight tests on PRs allows us to move
everything back to local runners. This eliminates the docker fetch
penalty introduced when introducing Nix.

Add `[force_expensive_checks]` to your commit message to force a full CI runs, even on PRs.

------------

TODO:

 - [x] Allow some way to force expensive CI runs on PRs
 - [x] Generate matrix in separate job to prevent useless (but relatively cheap) jobs
 - [x] Clean up commits


.github/workflows/ci.yml
```diff
@@ -621,6 +636,30 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Generate matrix
+        run: |
+          if [[ "${{ github.event_name }}" == "schedule" || $(.github/scripts/force_expensive_checks.sh) == "true" ]]; then
+            cp .github/synthesis/all.json checks.json
+          else
+            cp .github/synthesis/staging.json checks.json
```
kleinreact, line 651: Doesn't this alwasy overwrite the ouput generated in line 287?

.github/workflows/ci.yml
```diff
@@ -621,6 +636,30 @@ jobs:
...
+          fi
+
+      - name: Set test matrix
+        id: set-matrix
+        run: |
+          echo "check_matrix=$(cat checks.json | tr '\n' ' ')" >> "$GITHUB_OUTPUT"
+          cat "$GITHUB_OUTPUT"
```
kleinreact, line 658: Suggestion: use `tee`

.github/workflows/ci.yml
martijnbastiaan: Every job runs in a clean environment and in parallel. So `contranomy-hdl` and `bittide-instances-hdl-matrix` don't share any state/filesystem.

.github/workflows/ci.yml
martijnbastiaan: Good one, done.

.github/workflows/ci.yml
kleinreact: Right.

kleinreact:
> Add [force_expensive_checks] to your commit message to force a full CI runs, even on PRs.
We should collect and document these commit message "commands" in the main Readme or a `Development.md` somehow. Searching pull requests once you forgot which keyword it was is very cumbersome.

martijnbastiaan:
Agreed. I'll write some minimalistic instructions.

martijnbastiaan:
Something like this @kleinreact ?

kleinreact (approved):
Looks alright to me then.

PR #251, opened 2023-02-13
Add test using clash-protocols to wbStorage to check for double-ACKs

cuddlefishie:
Depends on https://github.com/clash-lang/clash-protocols/pull/52

This test catches the previous errors when double ACKs are not prevented. It also shows that delaying the ERR signal in a "no-downtime" scenario can cause synchronisation issues too.

There's two ways to resolve this ERR issue, either by keeping the signal delayed and making sure no double-ERRs can happen, or by not delaying the ERR signal at all.

For this PR I picked the second solution, that seems like it was the intended behaviour.

### DIFF
--- a/bittide/src/Bittide/DoubleBufferedRam.hs
+++ b/bittide/src/Bittide/DoubleBufferedRam.hs
@@ -250,12 +250,11 @@ wbStorage' initContent wbIn = delayControls wbIn wbOut
 
     -- It takes a single cycle to lookup elements in a block ram. We can therfore
     -- only process a request every other clock cycle.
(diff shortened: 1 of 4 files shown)


martijnbastiaan (approved):
Can we reuse the memory from `clash-protocols`? If not, can we use the simplified logic here too?

cuddlefishie:
The `memoryWb` is much more simplistic and was meant for testing or to be expanded upon. It can't replace the `wbStorage` as it is. But I can change the `delayControls` to use a state machine like in the clash-protocols PR :+1:

PR #254, opened 2023-02-14
Remove wishbone sink from processingElement

lmbollen:
The exposed write port available from the `processingElement` is originally intended for simulation, but we might want to use this for e.g. UART. Given the "low" speed of UART we would need access to the backpressure mechanism. 

For this reason we removed the exposed write port from the API of the `processingElement`. UART is to be added later.


martijnbastiaan:
Can you give some more context? Was this used for supporting `print`?

lmbollen:
Yes, it was intended to create a simulation print API that can be used in Nodes.
However, this approach prohibits reliable use in hardware because it doesn´t allow for backpressure.

martijnbastiaan (approved):
Gotcha, let's tear it out.

PR #256, opened 2023-02-14
Add wishbone packet buffers

lmbollen:
This pull requests adds the components important for our ethernet stack, these components are written in Axi4Stream, but will in the future probably be replaced with the currently experimental `PacketStream` due to lower complexity overhead.
 
1. A wishbone to axi transmitter `wbToAxiTx`:
 A component with a wishbone interface that produces an AXI transaction.
 Writing to address 1 will produce a transaction with _tlast set and all keep enables low. 
This allows us to write the entire packet to address 0 and then write to address 1 to signal the end of the packet without having to keep track of the last byte in software.

2. A wishbone accessible AXI receive buffer `wbAxisRxBuffer`:
This component contains a blockram to which AXI transactions will be written. Except for the last transactions, these transactions can *not* contain null bytes, since they will be written to the buffer as words. The memory layout is as follows:
```
0 .. 4 * (depth - 1) : The packet buffer
(4 * depth)            :  Packet length in bytes
(4 * (depth + 1)     : Status register (packet Complete and buffer full)
```
To read a packet we check the status register, if it contains data (as indicated by packet complete, buffer full or both), we read can read the packet from the buffer.

To clear a packet, we set the packet length register and status register to 0 in that order.
This buffer can hold a single packet and will stall incoming packets when it contains a complete packet.

3. A axi 1 to n scaling and packing circuit. `axisFromByteStream`
We receive the Ethernet packet as a 1 byte wide stream, but want to write it to the buffer as n bytes wide without gaps. Therefore we have to perform upscaling and packing.

4. An axi n to 1 scaling circuit. `axisToByteStream`
We produce a stream of n bytes wide, but the GMII interface we produce is 1 byte wide, therefore we have to downscale the stream to a single byte wide. 

5. Axi packing circuit `axiPacking`
For arbitrary packing without compromising on throughput. We don't currently test this.

6. Axi packet fifo `axiStreamPacketFifo`
A buffer that first consumes a complete packet and then produces it as a packet without gaps when complete (or the buffer is full)

Furthermore it introduces various utility functions for manipulating axi stream transactions


martijnbastiaan (changes requested):

bittide/src/Bittide/Axi4.hs
```diff
@@ -28,89 +46,708 @@ import Bittide.Extra.Maybe
...
-
-    nextStoredBytes
-      | upscaleDone && _tready = repeat initTempAxi
-      | otherwise              = init intermediateBytes
-
-{-# NOINLINE axisToByteStream #-}
+    undefUser = deepErrorX "axisFromByteStream: _tuser undefined"
+    -- Try to append the incoming axi to the stored axi
+    combinedAxi = axiUserMap (\(v1, v2) -> v1 :< v2) <$> concatAxi axiStored axiIn
+    inpValid = isJust axiIn && isJust combinedAxi
```
christiaanb, line 68: perhaps `inpPass`?

bittide/src/Bittide/Axi4.hs
```diff
@@ -28,89 +46,708 @@ import Bittide.Extra.Maybe
...
+    inpInvalid = isJust axiIn && isNothing combinedAxi
```
christiaanb, line 69: Perhaps `inpBlocked`?

bittide/src/Bittide/Axi4.hs
```diff
@@ -28,89 +46,708 @@ import Bittide.Extra.Maybe
...
+    axiPreShift = combinedAxi <|> (axiUserMap extendWithErrorX . extendAxi <$> axiStored)
+    axiBytes =
+      zipWith4 (\d k s u -> orNothing k (d, s, u))
+      (maybe (repeat 0) _tdata axiPreShift)
+      (maybe (repeat False) _tkeep axiPreShift)
+      (maybe (repeat False) _tstrb axiPreShift)
+      (maybe (repeat undefUser) _tuser axiPreShift)
+
+    (newData, newKeeps, newStrobes, newUsers) = unzip4 $
+      (maybe (0, False, False, undefUser) (\(d, s, u) -> (d, True, s, u))) <$> shiftPackVec axiBytes
```
christiaanb, line 81: This `shiftPackVec` seems expensive, when we know that only the `axiIn`  should contain a `tkeep` that's deasserted.

I guess it's acceptable for now. But do create an issue on the  issue tracker that this function should be made more resource efficient.

christiaanb (dismissed):

bittide/src/Bittide/Axi4.hs
```diff
@@ -28,89 +46,708 @@ import Bittide.Extra.Maybe
...
+    , _tuser = snd $ _tuser axi}
+
+  (dataA, dataB) = splitAtI $ _tdata axi
+  (keepA, keepB) = splitAtI $ _tkeep axi
+  (strbA, strbB) = splitAtI $ _tstrb axi
+  aValid = aKeepBytes || lastA
+  bValid = bKeepBytes
+  (aKeepBytes, bKeepBytes) = (or keepA, or keepB)
+  lastA = _tlast axi && not bKeepBytes
+  lastB = _tlast axi && bKeepBytes
```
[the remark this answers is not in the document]
lmbollen: The two resulting AXI transactions belong to the same stream, setting `_tlast` in both outputs would result in a second empty packet.

bittide/src/Bittide/Axi4.hs
```diff
@@ -28,89 +46,708 @@ import Bittide.Extra.Maybe
...
+  f Nothing acc = acc
+
+-- | Splits an Axi4StreamM2S into two Axi4StreamM2S. The first one contains all
+-- The data, keep and strobe vectors are split into two Vectors, the first vector is
+-- assigned to the first output, the second vector is assigned to the second output.
+-- If _tlast is not set, the outputs are only `Just` if at least one of their `keep`
+-- bits is set. If _tlast is set, the _tlast of the first output is set if the second
+-- output is `Nothing`. If the second output is `Just`, the _tlast of the first output
+-- is `False` and the _tlast of the second output is `True`.
+splitAxi ::
```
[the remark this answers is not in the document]
lmbollen: Then we can't test them.

bittide/src/Bittide/Axi4.hs
```diff
@@ -28,89 +46,708 @@ import Bittide.Extra.Maybe
...
+    (Axi4Stream dom ('Axi4StreamConfig dataWidth idWidth destWidth) ())
+    (PackedAxi4Stream dom ('Axi4StreamConfig dataWidth idWidth destWidth) ())
+axiPacking = Circuit (mealyB go (Nothing, False))
+ where
+  go (_, False) _ = ((Nothing, True), (Axi4StreamS2M False, Nothing))
+  go (axiBuffer, True) ~(axiIn :: axiType, s2mRight) = ((newStored, True), (s2mLeft, m2sRight))
+   where
+
+    -- Axi Right
+    validRight = maybe False (and . _tkeep) outputBuffer || maybe False _tlast outputBuffer || inpInvalid
```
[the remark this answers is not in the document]
lmbollen: Otherwise a malformed packet would lock up the bus.

bittide/src/Bittide/Axi4.hs
```diff
@@ -28,89 +46,708 @@ import Bittide.Extra.Maybe
...
+    statusBV = pack (packetComplete, bufferFull)
+    wbHandshake = masterActive && not err
+
+    (readData, nextReadingFifo, wbAcknowledge) = case (masterActive, internalAddress) of
+      (True, (== packetLengthAddress) -> True) -> (resize $ pack packetLength, False, wbHandshake)
+      (True, (== statusAddress) -> True) -> (resize statusBV, False, wbHandshake)
+      (True, _) -> (wbData, wbHandshake && not readingFifo, wbHandshake && readingFifo)
+      (False, _) -> (deepErrorX "undefined", False, False)
+
+    axisReady = packetError || not (packetComplete || bufferFull)
```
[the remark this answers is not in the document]
lmbollen: We can currently only store a single packet

bittide/src/Bittide/Axi4.hs
```diff
@@ -28,89 +46,708 @@ import Bittide.Extra.Maybe
...
+-- The wishbone addressing must be 4 byte aligned and is as follows:
+--  * 0 .. 4 * (fifoDepth - 1) = Read-only access into the buffer.
+--  * 4 * fifoDepth            = Byte count register.
+--  * 4 * (fifoDepth + 1)      = Status register
+--
+-- After reading a packet, the byte count must be set to 0 and the status register must be
+-- cleared.
+--
+-- The external status clear signals clear on True, set to (False, False) if not used.
+wbAxisRxBuffer ::
```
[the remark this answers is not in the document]
lmbollen: Personally I think we should switch over to a simpler protocol. AXI4Stream is too versatile to have generic components.

bittide-instances/tests/Wishbone/Axi.hs
```diff
@@ -0,0 +1,132 @@
+-- SPDX-FileCopyrightText: 2022-2024 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# OPTIONS_GHC -fplugin=Protocols.Plugin #-}
+{-# OPTIONS_GHC -fconstraint-solver-iterations=0 #-}
```
martijnbastiaan, line 7: Don't do this please. If there is a bug in the type checker this can cause it to go into an infinite loop and now you suddenly have a problem whose origin is hard to detect.

martijnbastiaan:
@lmbollen Could you look at the failing test in `Axi4.hs`? Uncomment the block and then:

```
cd bittide
cabal run unittests -- -j12 --hide-successes -p "axiStreamFromByteStream produces dense streams"
```

martijnbastiaan:
Almost there.. According to [my document :sparkles:](https://docs.google.com/spreadsheets/d/1iVRMWumZbVGUo3hlURAptXUnKqouaBs7HjWeTwPMVmU/edit?gid=0#gid=0), we're still missing:

- [ ] Direct tests for `wbToAxiTx`
- [ ] Direct tests for `packAxi4Stream`
- [ ] Direct tests for `unInterruptedAxi4Packets`
- [ ] A bugfix for the now commented `axiStreamFromByteStream produces dense streams`

lmbollen:
@martijnbastiaan I'd like to discuss if and how we would test `wbToAxiTx`

For now I've:
* added verification functions for properties we extracted from the spec
* Fixed the up and downscaling circuits
* Added tests for the properties and other functions

bittide-instances/tests/Wishbone/Axi.hs
```diff
@@ -0,0 +1,146 @@
+-- SPDX-FileCopyrightText: 2024 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# OPTIONS_GHC -fplugin=Protocols.Plugin #-}
+{-# OPTIONS_GHC -fconstraint-solver-iterations=0 #-}
```
martijnbastiaan, line 7: _Bro_

martijnbastiaan (changes requested):

bittide-instances/tests/Wishbone/Axi.hs
lmbollen: https://github.com/bittide/bittide-hardware/pull/589/commits/968b2c43c888b4618c770574dfc862e745a4bda8 :eyes:

martijnbastiaan (approved):
Let's make an issue for the missing test and let's call it a day!

PR #257, opened 2023-02-15
Add verilog-ethernet.nix

lmbollen:
We want to use the open source ethernet components found in https://github.com/alexforencich/verilog-ethernet 
without having to clone the entire repository.

This PR adds the repository as nix package and creates a nix environment variable `VERILOG_ETHERNET_SRC` that contains the path to the cloned repository.


shell.nix
```diff
@@ -51,12 +51,16 @@ pkgs.mkShell {
 
       # For Cabal to clone git repos
       pkgs.buildPackages.git
+
+      # Clone verilog-ethernet
+      pkgs.verilog-ethernet
```
martijnbastiaan, line 56:
```suggestion
      # HDL dependencies
      pkgs.verilog-ethernet
```

let's just make it a list to store our raw HDL deps

nix/nixpkgs.nix
```diff
@@ -7,6 +7,8 @@ let
     niv = (import sources.niv {}).niv;
     gitignore = import sources.gitignore { inherit (nixpkgs) lib; };
 
+    verilog-ethernet = import ./verilog-ethernet.nix { };
```
martijnbastiaan, line 10: I guess it doesn't _really_ matter, but

```suggestion
    verilog-ethernet = import ./verilog-ethernet.nix { nixpkgs };
```

?

martijnbastiaan (changes requested):

martijnbastiaan (approved):
Thanks for amending, LGTM

PR #261, opened 2023-02-17
Add protocol drivers

lmbollen:
This PR Adds simple simulation-only drivers that receive a list of master bus operations alongside a slave bus and drive a master bus that takes backpressure into account.

e.g.

A wishbone transaction will remain on the bus untill it is terminated by `acknowledge`, `err` or `retry`.
After a transaction is terminated, the next transaction will be presented on the bus in the next clock cycle.

### DIFF
--- a/bittide/tests/Tests/Shared.hs
+++ b/bittide/tests/Tests/Shared.hs
@@ -16,15 +16,17 @@ import Clash.Signal.Internal
 import Data.Constraint (Dict(Dict))
 import Data.Constraint.Nat.Extra (divWithRemainder)
 import Data.List.NonEmpty (NonEmpty)
+import Data.Maybe
 import GHC.Stack (HasCallStack)
 import Hedgehog
 import Numeric.Natural
 import Protocols (toSignals)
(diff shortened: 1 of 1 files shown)


bittide/tests/Tests/Shared.hs
```diff
@@ -266,3 +268,32 @@ fromListWithResetAndEnable inp =
       | rst       = inpAsSignal
       | ena       = xs
       | otherwise = x :- xs
+
+-- | A simple simulation-only driver that receives a list of Wishbone master requests
+-- alongside the slave bus and drives the master bus. The driver presents the next
+-- request after the current request has been terminated by `acknowledge`, `retry` or `err`.
+wishboneSimDriver ::
```
cuddlefishie, line 275: Couldn't this use `driveStandard` from `clash-protocols`?

lmbollen: I tried it and indeed. I'll put this in draft for now and close this PR once all instances have been replaced with `driverStandard`.

PR #264, opened 2023-02-20
Write settings before simulating

kleinreact:
This PR introduces some minor improvements:
* If clocks do not stabilize, a short message is output for better distinguishing the failure from other errors in CI.
* Simulation settings are also written before simulation starts such that they are available even if simulation fails.
* Heap memory limits got increased for the failing nightly simulation runs.


lmbollen (approved):
Nice work :)

PR #265, opened 2023-02-24
Fix JSON parse error

kleinreact:
Fixes 

```shell
Error when evaluating 'strategy' for job 'elastic-buffer-sim-topologies'. 
.github/workflows/ci.yml (Line: 136, Col: 17): 
  Error parsing fromJson,
.github/workflows/ci.yml (Line: 136, Col: 17): 
  After parsing a value an unexpected character was encountered: 
    {. Path '[6]', line 1, position 577.,
.github/workflows/ci.yml (Line: 136, Col: 17): 
  Unexpected type of value '', expected type: Sequence.
```


lmbollen (approved):
Should have seen that.
Too bad JSON doesn´t allow for trailing commas afaik.

Issue #249, opened 2023-02-08
Remove Contranomy


martijnbastiaan:
We've got VexRisc now!

TODO: 

 - [x] Add blackbox annotation to VexRisc
 - [x] change usage sites of contranomy to VexRiscv
     - [x] remove runtime reset-Instruction-Pointer
 - [x] Use external riscv package in https://github.com/bittide/bittide-hardware/pull/266

PR #266, opened 2023-02-27
Add VexRiscv blackbox and instance

cuddlefishie:
Part of #249 

This PR adds a blackbox instance for the VexRiscv core so that HDL can be generated properly.
It also adds a simple `bittide-instance` that uses the `ProcessingElement`, which now uses `vexRiscv` internally, to test that the blackbox passes synthesis.

Because VexRiscv can now be used, the `contranomy` and `contranomy-sim` packages have been removed together with the `riscv-formal` suite to test the contranomy core for spec compliance.


cuddlefishie:
superseded by #281

PR #273, opened 2023-04-05
Add simple fifo

lmbollen:
This PR adds a simple fifo component that I needed for the wishbone mapped UART interface

### DIFF
--- a/bittide/src/Bittide/DoubleBufferedRam.hs
+++ b/bittide/src/Bittide/DoubleBufferedRam.hs
@@ -595,8 +595,10 @@ fifo ::
   ( Signal dom (Maybe a)
   , Signal dom Bool
   , Signal dom (Index (fifoDepth + 1)))
-fifo depth@SNat fifoIn readyIn = (fifoOut, readyOut, dataCountOut)
+fifo depth@SNat fifoIn readyIn = (maybeOut, readyOut .&&. active, dataCountOut)
  where
+  maybeOut = mux active fifoOut (pure Nothing)
+  active = unsafeToLowPolarity hasReset .&&. fromEnable hasEnable
   bramOut = readNew (blockRamU NoClearOnReset depth (const undefined)) readAddr writeOp
   (readAddr, writeOp, fifoOut, readyOut, dataCountOut) =
     mealyB go initialState (fifoIn, readyIn, bramOut)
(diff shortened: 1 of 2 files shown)


bittide/src/Bittide/DoubleBufferedRam.hs
```diff
@@ -564,3 +564,57 @@ updateAddrs rdAddr (Just (i, a)) bufSelect =
 
 updateAddrs rdAddr Nothing bufSelect =
   (mul2Index rdAddr bufSelect, Nothing)
+
+{-# NOINLINE fifo #-}
+
+-- | Simple First in, First out buffer the contains a backpressure mechanism.
+fifo ::
+  forall dom fifoDepth a .
```
martijnbastiaan, line 572:
```suggestion
  forall depth a dom .
```

I'm assuming `depth` is the type argument you want to type apply most often, so it should go first. Also renamed to `depth` (from `fifoDepth`), because this function is already called `fifo`.

bittide/src/Bittide/DoubleBufferedRam.hs
```diff
@@ -564,3 +564,57 @@ updateAddrs rdAddr (Just (i, a)) bufSelect =
...
+fifo depth@SNat fifoIn readyIn = (fifoOut, readyOut, dataCount)
+ where
+  bramOut = readNew (blockRamU NoClearOnReset depth (const undefined)) readAddr writeOp
+  (readAddr, writeOp, fifoOut, readyOut, dataCount) = mealyB go (0, 0, 0, True) (fifoIn, readyIn, bramOut)
+  go ::
+    (Index fifoDepth, Index fifoDepth, Index (fifoDepth + 1), Bool) ->
+    (Maybe a, Bool, a) ->
+    ( (Index fifoDepth, Index fifoDepth, Index (fifoDepth + 1), Bool)
+    , (Index fifoDepth, Maybe (Index fifoDepth, a), Maybe a, Bool, Index (fifoDepth + 1)))
+  go (readCounter, writeCounter, dataCountGo, fifoEmpty) (fifoInGo, readyInGo, bramOutGo) =
```
martijnbastiaan, line 589: Could you make the state of the FIFO a record and document its fields?

martijnbastiaan (changes requested):
I don't feel this PR is ready. The documentation is underdeveloped - a developer can't use this function without diving into its implementation. Most notably:

 * The function documentation doesn't specify how the backpressure mechanism works. Is the boolean in the output a `ready` or more like an `ack`?
 * What happens if a write is issued when the FIFO is full?
 * What happens when a read is issued when the FIFO is empty? (For example, will it pass through write-data, or not.)

I also see some potential issues / have some questions I'd like to see addressed either in comments or in tests:

 * How does the FIFO behave during reset? Will the it signal it is ready to accept transactions or not?
 * How does the FIFO behave when enable is deasserted? Given that the enable isn't handled explicitly, I'm inclined to believe that the FIFO will silently drop elements. This might be expected behavior, but that must at the very least be documented.
 * How does the FIFO approximately work internally? 

As a last point: do we care about the few cycles of delay introduced by our Xilinx DCFIFO? I can see this function being a simple wrapper around it.

PR #277, opened 2023-04-11
Simplify `wbStorage`

lmbollen:
The current implementation of `wbStorage` allows for half-word aligned memory operation, which complicates the design significantly. Since the new RV32IMC core (VexRiscV) does not make use of that feature, I suggest we rip it out.


bittide/src/Bittide/DoubleBufferedRam.hs
```diff
@@ -108,24 +107,29 @@ wbStorageDP initial aM2S bM2S = (aS2M, bS2M)
...
+  nextActive = selectNow <$> nowActive <*> aM2S <*> aS2M <*> bM2S <*> bS2M
+  active WishboneM2S{busCycle, strobe} = busCycle && strobe
+  terminated WishboneS2M{acknowledge, err} = acknowledge || err
+  selectNow aorb am2s as2m bm2s bs2m  =
+    case (aorb, active am2s, terminated as2m, active bm2s, terminated bs2m) of
+      (_, True , _    , False, _     ) -> A
+      (_, False, _    , True , _     ) -> B
+      (A, True , True , True , _     ) -> B
+      (B, True , _    , True , True  ) -> A
+      _                                -> aorb
```
cuddlefishie, line 123: There's still no tests for `wbStorageDP`, right? Maybe it makes sense to add some since it's changed here?

bittide/src/Bittide/DoubleBufferedRam.hs
```diff
@@ -150,93 +153,45 @@ wbStorage' initContent wbIn = delayControls wbIn wbOut
...
-  splitWrite ::
-    KnownNat bits =>
-    Maybe (LocatedBits n (2*bits)) ->
-    (Maybe (LocatedBits n bits), Maybe (LocatedBits n bits))
-  splitWrite (Just (i, a)) = (Just (i,upper), Just (i, lower))
-   where
-    (upper,lower) = split a
-  splitWrite Nothing = (Nothing, Nothing)
+
+    masterWriting = acknowledge && writeEnable
```
cuddlefishie, line 185: Any reason the `&& not err` disappeared? I see it shouldn't really effect anything since if `err` then `! addrLegal`, so the write in `writeEntryGo` goes goes nowhere in the RAM, but I guess that could cause some issues since it's an invalid write?

bittide/tests/Tests/Shared.hs
```diff
@@ -266,3 +268,32 @@ fromListWithResetAndEnable inp =
       | rst       = inpAsSignal
       | ena       = xs
       | otherwise = x :- xs
+
+-- | A simple simulation-only driver that receives a list of Wishbone master requests
+-- alongside the slave bus and drives the master bus. The driver presents the next
+-- request after the current request has been terminated by `acknowledge`, `retry` or `err`.
+wishboneSimDriver ::
```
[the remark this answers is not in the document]
cuddlefishie: Oops, this is part of #261, will move the comment there!

bittide/src/Bittide/DoubleBufferedRam.hs
lmbollen: `acknowledge` implies `not err` :)

And you're right ,it would be a good idea to add tests for `wbStorageDP`.

bittide/src/Bittide/DoubleBufferedRam.hs
martijnbastiaan: Yes, but also: https://github.com/bittide/bittide-hardware/issues/304. We're going to push through because a lot of PRs depend on this one. Guess we'll see it falling apart later....

martijnbastiaan (approved):

PR #278, opened 2023-04-11
Add board programming to Shake

hiddemoll:
This PR adds board programming support to Shake. Both bitstream generation and board programming need each input and output pin to be mapped in an XDC file with the same name as the instance. With this PR, only `clockControlDemo0` and `clockControlDemo1` have XDC files with pin mappings, and can thus be flashed on hardware.

Vivado will throw an error when:
- There is no XDC file with pin mappings
- The pin mappings do not match the synthesis part
- Vivado Enterprise is needed for the synthesis part, but not correctly sourced
- No board is connected to the localhost on port 3121 (the default port)

If multiple boards are connected, the board with the lowest `hw_server`-name is chosen (alphabetically sorted). It is recommended to connect just one board to your PC.

### DIFF
--- a/bittide-instances/README.md
+++ b/bittide-instances/README.md
@@ -22,7 +22,7 @@ The build system automatically generates _false path_ constraints for all input 
 ## Prerequisites
 * We have tested the build system with Vivado 2022.1
 * To change the part for which the instances are synthesized, set the environment variable `SYNTHESIS_PART`. For the part we've bought use `SYNTHESIS_PART=xcku040-ffva1156-2-e`. Note that for this part you need to use Vivado Enterprise.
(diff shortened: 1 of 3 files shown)


bittide-instances/README.md
```diff
@@ -78,3 +78,10 @@ Example:
 ```
 cabal run -- bittide-instances:shake clockControlDemo0:bitstream
 ```
+
+## Board programming
+Example:
+
+```
+cabal run -- bittide-instances:shake clockControlDemo0:flash
```
martijnbastiaan, line 86:
```suggestion
cabal run -- bittide-instances:shake clockControlDemo0:program
```

?

martijnbastiaan (approved):
LGTM, maybe rename `flash` to `program`? You seem to be calling it "programming" everywhere else.

PR #279, opened 2023-04-11
Update Callisto to algo in `arxiv/2303.11467`

kleinreact:
Adds a variation of the reframing model of https://arxiv.org/abs/2303.11467 to the Callisto controller.

Instead of some global time `T_0`, which is implicitly shared among all nodes in the paper, the current implementation uses a local stability checker for each elastic buffer to determine whether the buffer occupancies have been stablized and centered, or not. If all of the elastic buffers of a node are stable, but not all are centered, then the controller of that node reframes. This process repeats until all elastic buffers are stable and centered eventually.

----

### TODOs:
  * [x] Fix the memory leak
  * [x] Verify that simulation produces the expected results
  * [x] Integrate #289
  * [x] ~~Enable the real FIFOs after everything has stablizied~~
  * [x] _Add a hourglass topology generator (optional)_
  * [x] ~~Reduce the size of the simulated examples to fit within github's 6h bound~~ The bounds have been exeeded to 14h.
  * [x] Add the "detect, store and wait" reframing approach

### DIFF
--- a/bittide/src/Bittide/ClockControl/Callisto.hs
+++ b/bittide/src/Bittide/ClockControl/Callisto.hs
@@ -69,14 +69,26 @@ data ControlSt = ControlSt
   -- ^ Previously submitted speed change request. Used to determine the estimated
   -- clock frequency.
   , _steadyStateTarget :: !Float
-  -- ^ Steady-state value determined at the reframing time (before correction).
+  -- ^ Steady-state value (determined when stability is detected for
+  -- the first time).
+  , _targetCorrection :: !Float
+  -- ^ Correction value to be applied at reframing time.
   , _reframeTime :: !Bool
   -- ^ flag for indicating that it's currently reframing time.
+  , _waitTime :: Unsigned 25
   } deriving (Generic, NFDataX)
 
 -- | Initial state of control
 initState :: ControlSt
-initState = ControlSt 0 NoChange 0.0 False
+initState =
+  ControlSt
+    { _z_k = 0
+    , _b_k = NoChange
+    , _steadyStateTarget = 0.0
+    , _targetCorrection = 0.0
+    , _reframeTime = False
+    , _waitTime = 20000000  -- ~100ms
+    }
 
 -- | Clock correction strategy based on:
 --
@@ -143,7 +155,7 @@ callisto margin framesize targetCount updateEveryNCycles mask allDataCounts =
   state = register initState $
     rfCheck
       <$> allStable
-      <*> allSettled
+  --    <*> allSettled
       <*> D.toSignal c_des
       <*> mux shouldUpdate updatedState state
   updatedState = D.toSignal $
@@ -151,7 +163,9 @@ callisto margin framesize targetCount updateEveryNCycles mask allDataCounts =
       <$> D.delayI (errorX "callisto: No start value [2]") z_kNext
       <*> b_kNext
       <*> D.delayI (errorX "callisto: No start value [3]") steadyStateTarget
+      <*> D.delayI (errorX "callisto: No start value [4]") targetCorrection
       <*> D.delayI (errorX "callisto: No start value [6]") reframeTime
+      <*> D.delayI (errorX "callisto: No start value [6]") waitTime
 
   -- See fields in 'ControlSt' for documentation of 'z_k', 'b_k', and css.
   z_k :: DSignal dom 0 (Signed 32)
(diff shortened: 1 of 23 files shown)


bittide/src/Bittide/ClockControl/Callisto.hs
```diff
@@ -45,22 +45,25 @@ callistoClockControl ::
...
+  -- ^ Counter option for detecting the reframing time. It is set to
+  -- 'Nothing' after the reframing has taken place.
+  , _css :: !Float
+  -- ^ stead-state value determined at the reframing time (before correction).
   } deriving (Generic, NFDataX)
 
 -- | Initial state of control
 initState :: ControlSt
-initState = ControlSt 0 0 NoChange
+initState = ControlSt 0 NoChange (Just 0) 0
```
martijnbastiaan, line 66: Maybe use record syntax?

martijnbastiaan (commented):
Drive-by review while peer programming

bittide/src/Bittide/ClockControl/Callisto.hs
```diff
@@ -95,40 +121,57 @@ callisto ::
...
  where
   dataCounts = filterCounts <$> fmap bv2v mask <*> allDataCounts
   filterCounts vMask vCounts = flip map (zip vMask vCounts) $
     \(isActive, count) -> if isActive == high then count else 0
+  scs = bundle $ map (stabilityChecker margin framesize) $ unbundle dataCounts
   updateCounter = wrappingCounter updateEveryNCycles
   shouldUpdate = updateCounter .==. 0
-  state = register initState (mux shouldUpdate updatedState state)
+  allStable = fold (&&) . (True :>) . map fst <$> scs
+  allCentered = fold (&&) . (True :>) . map snd <$> scs
```
martijnbastiaan, line 142:
```suggestion
  allStable = and . map fst <$> scs
  allCentered = and . map snd <$> scs
```

martijnbastiaan: Also, a small record with two fields `stable` / `centered` would make this more readable / less error prone.

bittide/src/Bittide/ClockControl/Callisto.hs
```diff
@@ -179,3 +209,13 @@ callisto targetCount updateEveryNCycles mask allDataCounts =
   sign NoChange = 0
   sign SpeedUp = 1
   sign SlowDown = -1
+
+  rfCheck stable centered d cst@ControlSt{..}
+    | stable && not centered && not _rft =
+        cst { _css = d
+            , _rft = True
+            }
```
martijnbastiaan, line 217: For people not having worked on the algorithm a lot (everyone but you and Sanjay I suppose :-)), this is pretty hard to read. I'm getting lost in all the abbreviations `cst`, `css`, `rft`, `d`, ... Can they be expanded?

elastic-buffer-sim/src/Bittide/Plot.hs
```diff
@@ -368,29 +396,39 @@ simulateTopology ccc margin framesize graph settings = do
...
+         (markBlue,  [pData x]  ) : a
+       |      stable x  &&      centered x  &&           not ac =
+         (markGreen, [pData x]  ) : a
+       |      stable x  && not (centered x) &&     as &&     ac =
+         (markBlue,  [pData x]  ) : (m, reverse y) : yr
+       |      stable x                      &&     as           =
+         (m,         pData x : y) : yr
+       | not (stable x)                     &&     as           =
+         (m,         reverse y  ) : yr
+       | otherwise                                              =
```
martijnbastiaan, line 426: It's probably sensible to break the 80/90 limit here. It's not a complicated piece of code, but it's also hard to format it properly. (One time only :eyes:.)

```suggestion
    a' |      stable x  && not (centered x) && not as           = (markBlue,  [pData x]  ) : a
       |      stable x  &&      centered x  &&           not ac = (markGreen, [pData x]  ) : a
       |      stable x  && not (centered x) &&     as &&     ac = (markBlue,  [pData x]  ) : (m, reverse y) : yr
       |      stable x                      &&     as           = (m,         pData x : y) : yr
       | not (stable x)                     &&     as           = (m,         reverse y  ) : yr
       | otherwise                                              = a
```

martijnbastiaan (approved):
I've got some tiny comments. If this holds up the milestone I suggest we merge it.

bittide/src/Bittide/ClockControl/Callisto.hs
kleinreact: Done.

elastic-buffer-sim/src/Bittide/Plot.hs
kleinreact: Fixed.

bittide/src/Bittide/ClockControl/Callisto.hs
kleinreact: Done.

martijnbastiaan (approved):

martijnbastiaan (approved):
LGTM. The store, wait, center approach looks _really_ clean!

PR #280, opened 2023-04-12
Add smoltcp-example

lmbollen:
This example code implements an internal loopback device based on smoltcp.

It does not use any peripherals.


firmware/examples/smoltcp-example/src/time.rs
```diff
@@ -0,0 +1,73 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Clock{
+    global_time: u64,
+    timer_last: u32,
+    ticks_per_micro: u64,
+    timer_now: *const u32
+}
+
+impl Clock {
+    pub fn new(addr: usize, frequency: u64) -> Clock {
```
cuddlefishie, line 17: Since `addr` is meant to be the value of a pointer, this should be `addr: *mut u32` so it's clear to the caller.

Also because this function blindly dereferences that point without any security checks this whole function should be `unsafe`. Since this constructs a hardware abstraction it should be unsafe, but other methods on it don't need to be unsafe :)

lmbollen:
Superseeded by #512

PR #289, opened 2023-05-03
Add `Bittide.Counter`: a way of measuring domain speed differences

martijnbastiaan:
Exposes a single function, `domainDiffCounter`, that we can use to measure speed differences between two domains. These can in turn be used to eliminate the need for using FIFOs directly to measure clock divergence.

TODO: 

 - [x] Tests
 - [x] Instance?

### DIFF
--- a/bittide/src/Bittide/Counter.hs
+++ b/bittide/src/Bittide/Counter.hs
@@ -1,0 +1,142 @@
+-- SPDX-FileCopyrightText: 2022 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+module Bittide.Counter
+  ( Active
+  , domainDiffCounter
+  ) where
+
+import Clash.Explicit.Prelude
+
+import Clash.Cores.Xilinx.Xpm (xpmCdcGray)
+import Clash.Sized.Extra (unsignedToSigned, concatUnsigneds)
+
+-- | State of 'domainDiffCounter'
+data DdcState
+  -- | In reset, or waiting for the incoming counter to change
+  = DdcInReset
+  -- | Counting and comparing with incoming domain
+  | DdcRunning (Unsigned 64)
+  deriving (Generic, NFDataX)
+
+-- | Indicates whether 'domainDiffCounter' is actively counting or still in reset.
+type Active = Bool
+
+-- | Determine speed differences between two domains. If the source domain is
+-- faster than the destination domain, the result will become smaller. Vice versa,
+-- if the source domain is slower than the destination domain, the result will
+-- become larger. This is analogous to what would happen to a FIFO's data count
+-- when continuously written to by the source domain and read from by the
+-- destination domain. To ease integration in control algorithms, this component
+-- makes sure it starts counting at zero. It also waits for the incoming counter
+-- to become active, i.e. non-zero, before starting to count itself.
+--
+-- If both domains support initial values, 'domainDiffCounter' does not need to
+-- be reset.
+--
+-- To reset this component, the reset should be asserted for at least one cycle in
+-- the source domain _plus_ four cycles in the destination domain. The reset in the
+-- destination domain should be deasserted at the same time or *after* the one in
+-- the source domain for glitchless operation.
+--
+-- __N.B.__:
+--   This function will only work properly if the given domains are pretty close
+--   to another for a number of reasons:
+--
+--     1. It uses an 8-bit Gray counter internally for CDC
+--
+--     2. It uses one 64-bit counter in each domain
+--
+--     3. Its output is constrained to @Signed 32@
+--
+--   These values have been chosen such that:
+--
+--     * The 64-bit counter only overflows once every 3000 years at 200 MHz
+--
+--     * The 32-bit output only overflows after running at maximum divergence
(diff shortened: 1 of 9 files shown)


bittide/src/Bittide/Counter.hs
```diff
@@ -0,0 +1,122 @@
...
+import Clash.Cores.Xilinx.Xpm (xpmCdcGray)
+import Clash.Sized.Extra (unsignedToSigned, glueUnsigneds)
+
+data DdcState
+  = DdcInReset
+  | DdcRunning (Unsigned 64)
+  deriving (Generic, NFDataX)
+
+-- | Indicates whether 'domainDiffCounter' is actively counting or still in reset.
+type Active = Bool
```
kleinreact, line 21: `Active` is not used currently.

bittide/src/Bittide/Counter.hs
```diff
@@ -0,0 +1,134 @@
...
+  , domainDiffCounter
+  ) where
+
+import Clash.Explicit.Prelude
+
+import Clash.Cores.Xilinx.Xpm (xpmCdcGray)
+import Clash.Sized.Extra (unsignedToSigned, glueUnsigneds)
+
+-- | State of 'domainDiffCounter'
+data DdcState
```
kleinreact, line 16: This definition matches `EscState`, so maybe reduce it to a single one. This makes it also easier to notice the 64 bit size constraint, while this way a reader alwasy has to keep in mind that `DdcState` is hardcoded to be 64 bits.

kleinreact: Other than that, its basically just `Maybe`. Decide on your own whether you like to keep use of that fact.

bittide/src/Bittide/Counter.hs
```diff
@@ -0,0 +1,134 @@
...
+-- __N.B.__: This function uses an 8-bit Gray counter internally, and will therefore
+--           only work properly if both clock speeds are pretty close to one another.
+synchronizedSuccCounter ::
+  forall n m src dst .
+  ( KnownDomain src
+  , KnownDomain dst
+  , KnownNat n
+  , 8 <= n
+
+  -- TODO: Haxiom this out:
```
kleinreact, line 88: Don't forget.

bittide/tests/Tests/Counter.hs
```diff
@@ -0,0 +1,80 @@
...
+      sampledNonZero = P.dropWhile (==0) sampled
+      len = P.length sampledNonZero
+    assertBool (">1 @ " <> show n) (len > 1)
+    assertEqual ("exp @ " <> show n) sampledNonZero (P.take len expected)
+ where
+  dut = top @D10 @D20 noRst
+  expected = P.concat [[n, n] | n <- [1..]]
+
+tests :: TestTree
+tests = $(testGroupGenerator)
```
kleinreact, line 71: I its not too much of an effort, also testing the overflow of the `Signed 16` output and the internal `Unsigned 64` counter would be great. It probably requires making these numbers smaller for testing. It depends on the open implementation efforts regarding my previous comments.

kleinreact (changes requested):
I am a bit worried about the overflow of the internal counters. Is this not handled on purpose, or have I missed something such that this is no issue in our case?

bittide/src/Bittide/Counter.hs
```diff
@@ -0,0 +1,134 @@
...
+  counter = synchronizedSuccCounter @64 clkSrc rstSrc clkDst rstDst
+
+  go :: DdcState -> Unsigned 64 -> (DdcState, (Signed 16, Bool))
+  go DdcInReset c1
+    | c1 == 0   = (DdcInReset,          (0, False))
+    | otherwise = (DdcRunning (c1 + 1), (0, True))
+  go (DdcRunning c0) c1 = (DdcRunning (c0 + 1), (c0 `subAndTruncate` c1, True))
+
+  subAndTruncate :: Unsigned 64 -> Unsigned 64 -> Signed 16
+  subAndTruncate c0 c1 = truncateB (unsignedToSigned c0 - unsignedToSigned c1)
```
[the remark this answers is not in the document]
martijnbastiaan: You're right, I didn't do my back-of-a-napkin calculation right. The goal of this would to basically never overflow in a Bittide system. Given our clock setup we know that the maximum divergence is +-100 ppm. On a 200 MHz clock this amounts to 40k ticks difference at most, every second. A 16 bit number would therefore quickly overflow (~3 seconds).

I think it's therefore best to up this to a `Signed 32`. This would allow a system to run a maximum divergence rate for 2.5 days. This should be more than enough given that neither the rate nor that duration is realistic - clock control should have kicked in way before that.

bittide/src/Bittide/Counter.hs
martijnbastiaan: Apparently this was only necessary in earlier iterations of the code :shrug:

bittide/src/Bittide/Counter.hs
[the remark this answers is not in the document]
martijnbastiaan: We decided that this is enough for now.

kleinreact (approved):
LGTM then.

PR #294, opened 2023-05-09
Add `Bittide.Extra.Maybe`

lmbollen:
Used in #235 and various other places

### DIFF
--- a/bittide-extra/src/Bittide/Extra/Maybe.hs
+++ b/bittide-extra/src/Bittide/Extra/Maybe.hs
@@ -1,0 +1,47 @@
+-- SPDX-FileCopyrightText: 2023 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+module Bittide.Extra.Maybe where
+
+import Clash.Prelude
+
+import Data.Maybe
+
+{- $setup
+>>> import Clash.Prelude
+-}
+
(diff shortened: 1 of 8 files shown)


martijnbastiaan (changes requested):
LGTM, but it seems like `cabal run bittide-extra:doctests` fails locally. Could you add it to CI?

bittide-extra/src/Bittide/Extra/Maybe.hs
```diff
@@ -0,0 +1,47 @@
...
+--
+-- >>> fromMaybesR 0 (Just 1 :> Just 2 :> Nothing :> Nil)
+-- 2
+-- >>> fromMaybesR 0 (Nothing :> Nothing :> Nothing :> Nil)
+-- 0
+--
+fromMaybesR :: a -> Vec n (Maybe a) -> a
+fromMaybesR a = fromMaybe a . fold (flip (<|>)) . (Nothing :>)
+
+-- | Returns 'Just a' when the boolean is 'True', or 'Nothing' when 'False'.
```
DigitalBrains1, line 36: Since `bittide` depends on `clash-lib`, you could also just import `clash-lib:Clash.Netlist.Util.orNothing`. But maybe you're setting the stage for adding this module to `clash-prelude` and that's why you are copying the existing function from `clash-lib` to here, to ultimately copy to `clash-prelude`?

DigitalBrains1:
> Used in #235 and various other places

I think you linked the wrong PR?

Issue #291, opened 2023-05-03
Reduce Rust binary sizes


martijnbastiaan:

cuddlefishie:
A utility program that repackages ELFs and only keeps what we are interested in can be found here: https://github.com/cuddlefishie/elf_repack.

We probably want to not actually transfer full ELFs, because all `#![no_std]` ELF readers do not support streaming, so a custom streamable format will be implemented.
This makes reduction in program size more important than gains from reducing the size of the ELF container.

cuddlefishie:
Things to investigate:

- [ ] using [`ufmt`](https://lib.rs/crates/ufmt)
- [ ] building with `build_std` to remove panic messages in `core` and optimise for size

cuddlefishie:
Results from using `ufmt` and `build_std` on the example firmware binaries

Note: these numbers are only "fixed" memory usage, the `.stack` section that grows to fill all available data memory is excluded.

To make displaying and *checking* those sizes easier, there's now [cuddlefishie/elf-limits](https://github.com/cuddlefishie/elf-limits), which was used to get those numbers below.

## `hello`

Configuration | Total memory | Instructions | Data
-- | --:| --:| --:
Baseline | 24.28 KiB | 18.05 KiB | 6.23 KiB
`build_std` | 12.04 KiB | 9.32 KiB | 2.73 KiB
`build_std` + `ufmt` | 5.73 KiB | 5.31 KiB | 428 B

## `fdt-read`

Configuration | Total memory | Instructions | Data
-- | --:| --:| --:
Baseline | 29.13 KiB | 22.29 KiB | 6.84 KiB
`build_std` | 15.38 KiB | 12.34 KiB | 3.04 KiB
`build_std` + `ufmt` | 7.66 KiB | 7.26 KiB | 404 B

PR #300, opened 2023-05-16
use a custom program streaming format

cuddlefishie:
As part of #291 , this format will reduce the amount of data that needs to be transferred to a bootloader, making iteration times shorter.

An ELF container encodes a lot more information than a simple bootloader needs. Additionally, none of the `#![no_std]` ELF readers support using a streaming format. That means the entire ELF needs to be loaded and held in memory, to then be parsed and then the contents are written out to actual memory locations, essentially using double the memory requirements on the system.

The data that is *kept* from the ELF is essentially:
- entry point
- number of segments to load
- address of segment
- whether segment contains instructions or data
- data for segment
- needed zero padding of a segment

This format allows a bootloader to keep very minimal overhead for the receiving of the data and enables reading the data directly into the memory addresses they should end up in anyway!

It's only a rough estimation, but with this, the required memory for loading a program from some `impl Iterator<Item = u8>` (that is not counted here) should be around 32 bytes!

Because the streaming format strips away information from the ELFs that are not needed, they are strictly smaller in size. (The `all_streams_are_smaller_than_elfs` property test also checks this)

TODOs:
- [x] write tests


bittide/src/Bittide/ProcessingElement/ProgramStream.hs
```diff
@@ -0,0 +1,85 @@
...
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+module Bittide.ProcessingElement.ProgramStream (
+    elfStreamStructure,
+    elfStream,
+    ZeroPaddingLength,
+    IsExecutable,
+    Segment,
+    Address
```
martijnbastiaan, line 11:
```suggestion
    , elfStreamStructure
    , elfStream
    . ZeroPaddingLength
    , IsExecutable
    , Segment
    , Address
```

bittide/src/Bittide/ProcessingElement/ProgramStream.hs
```diff
@@ -0,0 +1,85 @@
...
+
+    | PF_R `elem` elfSegmentFlags seg
+    = let
+        segData = elfSegmentData seg
+        fileSz = fromIntegral $ BS.length segData
+        memSz = fromIntegral $ elfSegmentMemSize seg
+        zeroPadding = memSz - fileSz
+        addr = fromIntegral $ elfSegmentPhysAddr seg
+        segEntry   = (False, addr, bytes segData, zeroPadding)
+      in segEntry : acc
```
martijnbastiaan, line 81: Deduplicate this with a `goSegment`?

firmware/bittide-sys/tests/program_stream.rs
```diff
@@ -0,0 +1,151 @@
...
+lazy_static::lazy_static! {
+    static ref PROG_STREAM_EXEC: PathBuf = program_stream_exec_path().unwrap();
+}
+
+#[test]
+fn find_cabal_stuff() {
+    dbg!(&*PROG_STREAM_EXEC);
+}
+
+unsafe fn write_program_to_memory(input: &mut impl Iterator<Item = u8>) -> Option<()> {
```
martijnbastiaan, line 91: This seems to always return a trivial `Some()`. What's up with this return type? Shouldn't it return simply nothing?

martijnbastiaan (changes requested):
Could you:

 * Add Haddock documentation to your Haskell functions
 * Write a PR cover letter: it should answer the question why this reduce the amount of memory we need to transfer binaries

But as far as I can tell: looks good!

firmware/bittide-sys/tests/program_stream.rs
martijnbastiaan: Great makes sense

cuddlefishie:
Added haddock comments, a new property test to make sure the streaming format will always be smaller than an ELF (just a sanity check) and the PR description was updated!

martijnbastiaan (approved):

PR #305, opened 2023-05-31
Add missing Haddock comment to `uartWb`.

lmbollen:
Title says all

### DIFF
--- a/bittide/src/Bittide/Wishbone.hs
+++ b/bittide/src/Bittide/Wishbone.hs
@@ -145,7 +145,18 @@ uartDf baud = Circuit go
    where
     (received, txBit, ack) = uart baud rxBit (dataToMaybe <$> request)
 
--- |
+-- | Wishbone accessible UART interface with configurable FIFO buffers.
+--   It takes the depths of the transmit and receive buffers and the baud rate as parameters.
+--   The function returns a 'Circuit' with a 'Wishbone' interface and a 'CSignal' for the UART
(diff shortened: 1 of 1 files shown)


bittide/src/Bittide/Wishbone.hs
```diff
@@ -145,7 +145,17 @@ uartDf baud = Circuit go
...
+--   It takes the depths of the transmit and receive buffers and the baud rate as parameters.
+--   The function returns a 'Circuit' with a 'Wishbone' interface and a 'CSignal' for the UART
+--   receive bit as inputs, and outputs a 'CSignal' for the UART transmit bit and a 'CSignal'
+--   tuple indicating the status of the UART.
+--
+--   The register layout is as follows:
+--   - Address 0: UART data register (read/write)
+--   - Address 4: UART status register (read-only)
+--     - Bit 0: Transmit buffer full flag
+--     - Bit 1: Receive buffer empty flag
```
cuddlefishie, line 158: I think this should mention how those are supposed to be read, because they take the addresses of a double-word, but both are meant to be read as bytes.

```suggestion
--   - Address 0 (byte): UART data register (read/write)
--   - Address 4 (byte): UART status register (read-only)
--     - Bit 0b00: Transmit buffer full flag
--     - Bit 0b01: Receive buffer empty flag
```

lmbollen: After a brief discussion we chose to show the masks instead.

cuddlefishie (approved):

PR #307, opened 2023-06-08
#279 with optional reframing

martijnbastiaan:
This PR builds on top of https://github.com/bittide/bittide-hardware/pull/279 to allow for optional reframing. This allows the upcoming milestone demo to be significantly simpler.

### DIFF
--- a/bittide/src/Bittide/ClockControl/Callisto.hs
+++ b/bittide/src/Bittide/ClockControl/Callisto.hs
@@ -16,7 +16,6 @@ import Data.Constraint.Nat.Extra (euclid3, useLowerLimit)
 import Bittide.ClockControl
 import Bittide.ClockControl.Callisto.Util
 import Bittide.ClockControl.StabilityChecker
-import Clash.Sized.Extra
 
 import qualified Clash.Cores.Xilinx.Floating as F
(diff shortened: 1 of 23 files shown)


bittide/src/Bittide/ClockControl/Callisto.hs
```diff
@@ -161,7 +166,9 @@ callisto margin framesize targetCount updateEveryNCycles mask allDataCounts =
   b_k = D.fromSignal (_b_k <$> state)
 
   steadyStateTarget :: DSignal dom 0 Float
-  steadyStateTarget = D.fromSignal (_steadyStateTarget <$> state)
+  steadyStateTarget
```
kleinreact, line 169: Using the `reframingEnalbed` flag at this point may cause problems, since `rfCheck` hasn't be applied yet and `steadyStateTarget` is only utilized on a state update. Thus `rfCheck` is able to overwrite the choice and keep it until the next update. If `_targetCorrection` gets updated during that time a reframe will still take place.

I'll cherry-pick this commit for #279 and propose a fix there.

martijnbastiaan: Thanks!

PR #314, opened 2023-06-23
Use relative files in Shake, but set CWD to project root

hiddemoll:
This will make commands such as 'shake some/file.xdc' work


martijnbastiaan (approved):
LGTM

PR #315, opened 2023-06-23
Add support for device tree specification embeddings

kleinreact:
The PR adds support for embedding device tree specification files into the project such that the entries of the files are directly accessible at compile time. To this end, each file with a `.dts`-ending that is listed under the `extra-source-files` field of the project's cabal file is automatically embedded. Each value of a property defined in these files can be easily queried via the uniqe path to the property within the respective tree.

------

### TODO (optional):
  - [ ] `/include/` support (should be easy)


martijnbastiaan:
This is great work! I think we should publish this on Hackage.

PR #320, opened 2023-06-28
Use less pessimistic cache invalidation

martijnbastiaan:


bittide-instances/bin/Shake.hs
```diff
@@ -39,6 +39,63 @@ import qualified Clash.Util.Interpolate as I
...
+ignorePatterns =
+  [ "*.md"
+  , ".github"
+  , ".reuse"
+  , ".vscode"
+  , "LICENSES"
+
+  -- Used for synthesis, but not for generating Clash output:
+  , "bittide-instances/data/constraints/*.xdc"
+  ]
```
hiddemoll, line 97: Either add ".license" or remove "LICENSES". Since both rarely change, I prefer the latter.

hiddemoll (changes requested):
LGTM, I only have one nitpicky change request

hiddemoll (approved):

PR #322, opened 2023-06-29
Add initial startup offsets to simulation

kleinreact:
Adds the generation and configuration of some initial startup offsets (per node) to the simulation and a legend to the plots in the reports.

### DIFF
--- a/elastic-buffer-sim/exe/Main.hs
+++ b/elastic-buffer-sim/exe/Main.hs
@@ -414,7 +414,9 @@ data Options =
     , waitTime           :: Int
     , stopWhenStable     :: Bool
     , stopAfterStable    :: Maybe Int
-    , offsets            :: [Int64]
+    , clockOffsets       :: [Int64]
+    , startupOffsets     :: [Int]
+    , maxStartupOffset   :: Int
     , outDir             :: FilePath
     , jsonArgs           :: Maybe FilePath
     , stable             :: Maybe Bool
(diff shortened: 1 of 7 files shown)


elastic-buffer-sim/exe/Main.hs
```diff
@@ -504,12 +506,31 @@ optionParser =
...
-          <> help "Initital clock offsets (randomly generated if missing"
+          <> help "Initital clock offsets (randomly generated if missing)"
+          )
+    <*> option auto
+          (  long "startup-offsets"
+          <> short 'T'
+          <> metavar "NUM LIST"
+          <> value []
+          <> showDefault
+          <> help "Initital startup offsets (randomly generated if missing)"
```
martijnbastiaan, line 522: Could you add some units to this? I'm guessing this is defined in `fs` or `ps`?

kleinreact: It's currently in clock cycles (according to the node's local clock). It was the easiest to setup. Is that ok for you or do you want to have it converted?

martijnbastiaan: Yeah that's fine, as long as that unit is mentioned

martijnbastiaan (commented):
The clock offsets look good - though I feel like the current offsets aren't necessarily realistic. While it's hard to say what the offsets should be, I feel a more realistic environment would measure this in milliseconds:

 * Our VIOs have a delay ~1ms
 * Wake-On-Lan would also measure in ms for larger systems

What do you think?

--------------------------------------

The legends look nice too, though we should probably draw them outside of the graph - in some graphs they block the most interesting part :-). Could you split that up into a separate PR?

kleinreact:
> What do you think?

I agree, however I think we should remember to differentiate between simulation experiments and features of the simulator at this point. The report shows that the new feature works as expected. That's all it needs for the PR. As soon as the feature got's into `staging` you still can adapt the parameters for creating dedicated simulation experiments. I am happy with the current default so far.

> should probably draw them outside of the graph

Thats the automatic layout of matplotlib. I'll check whether there are other placing options too. Always using the rightmost upper corner should be fine as well, since anything should have converged at that point already. And you always can take the legend out afterwards using inkscape if it is really important to look below.

martijnbastiaan (approved):
OK, sounds reasonably. Offset part of this PR LGTM then :heavy_check_mark:.

martijnbastiaan:
> Offset part of this PR LGTM then

:eyes: 

Could you make an issue stating the legends should be drawn next to the plots instead of over them? Having to whip out Inkscape to see parts is not really acceptable IMO.

martijnbastiaan:
> Always using the rightmost upper corner should be fine as well, since anything should have converged at that point already. 

To be clear I'm seeing:

Screenshot from 2023-07-01 09-02-35

PR #325, opened 2023-06-30
Add support board specific IP generation

hiddemoll:
Add support for generation of device specific IP. We already supported for setting a `part` using the environment variable `SYNTHESIS_PART`. We now also support setting a board part with the `SYNTHESIS_BOARD`. An error is thrown when both are set. If neither is set the default part is still used (`xcku035-ffva1156-2-e`).


bittide-instances/bin/Shake.hs
```diff
@@ -210,6 +210,17 @@ vivadoFromTcl tclPath =
     "vivado"
     ["-mode", "batch", "-source", tclPath]
 
+getSynthesisPart :: Action BoardPart
```
martijnbastiaan, line 213:
```suggestion
-- | Constructs a 'BoardPart' based on environment variables @SYNTHESIS_BOARD@
-- or @SYNTHESIS_PART@. Errors if both are set, returns a default (free) part
-- if neither is set.
getBoardPart :: Action BoardPart
```

martijnbastiaan (approved):

PR #327, opened 2023-06-30
Add systemd service file for Vivado Hardware Server

hiddemoll:
Since the hardware-in-the-loop tests on CI depend on the Vivado Hardware Server, the GitHub runner should not start before it. We therefore add a systemd service which starts the Vivado Hardware Server. That service must run before starting the GitHub runner service.

Since the "unit" file for the GitHub runner is generated by another script, we add an additional configuration for the GitHub runner service with the `systemd drop-in` mechanism.


martijnbastiaan (approved):

martijnbastiaan:
`s/GitLab/GitHub/g` :-)

PR #330, opened 2023-07-03
Add `.envrc`

martijnbastiaan:
After allowing this specific script with `direnv allow`, users of direnv will be auto-dropped to a Nix shell every time they enter a (sub)directory of this repository.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,4 @@
+# SPDX-FileCopyrightText: 2023 Google LLC
+#
+# SPDX-License-Identifier: CC0-1.0
+use_nix -j8


.envrc
```diff
@@ -0,0 +1,4 @@
+# SPDX-FileCopyrightText: 2022 Google LLC
```
kleinreact, line 1: 2023?

kleinreact (commented):
Sounds useful, I wasn't aware of `direnv` before. However, are you sure it should be enabled by default? This might interfer with peoples' personal settings.

.envrc
martijnbastiaan: Woops, yes

martijnbastiaan:
> However, are you sure it should be enabled by default

Direnv will only run automatically if you've explicitly allowed a folder+file with `direnv allow`. I'll update the PR description.

kleinreact (approved):
LGTM

PR #336, opened 2023-07-05
Retry connection when HW server has too few targets

hiddemoll:


bittide-instances/data/tcl/HardwareTest.tcl
```diff
@@ -15,7 +15,8 @@ set fpga_ids {
   210308B0B0C2
 }
 
-set timeout_ms 1000
+set test_timeout_ms 1000
+set hw_server_timeout_ms 5000
```
martijnbastiaan, line 19:
```suggestion
# Timeout specifying how long we should wait for a test to finish before considering
# it a failed test.
set test_timeout_ms 1000

# Timeout specifying how long to wait for hardware targets (FPGAs) to come online
set hw_server_timeout_ms 5000
```

martijnbastiaan (approved):

PR #361, opened 2023-07-19
Haskell / Rust FFI for Callisto

kleinreact:
Adds a Callisto Clock Control implementation in Rust and the required Haskell-to-Rust FFI for simulating it. The interface can be enabled with the `cccEnableRustySimulation` flag in the clock control config, which is disabled by default.

----

TODO: 
  - [x] #363 (requirement)
  - [x] Split up clock control and the Haskell-to-Rust FFI wrapper into differnet crates
  - [x] Decide on how to measure the wait time


bittide-rust-ffi/src/lib.rs
```diff
@@ -0,0 +1,530 @@
...
+    NoChange = 2
+  }
+
+// We use `static_assert` to verify that the memory layout appears to
+// be as expected for ensuring a consistent encoding at the Haskell
+// side.
+const_assert!(
+  4 ==
+    mem::size_of::<SpeedChange>()
+);
```
cuddlefishie, line 30: The Rust version we use you can do `const _: () = assert!(4 == mem::size_of::<SpeedChange>());` and get rid of the `static_assert` dependency.

bittide-rust-ffi/src/lib.rs
```diff
@@ -0,0 +1,530 @@
...
+/// Type wrapper for `Vec<StabilityIndication>`.
+pub struct VSI(Vec<StabilityIndication>);
+
+/// Allows to casts a "`void`" pointer into a non-mutable `ControlSt`
+/// reference (unsafe).
+impl ReprRust<*const ()> for &VSI {
+  fn from(ptr: *const ()) -> Self {
+    return unsafe { mem::transmute(ptr) };
+  }
+}
```
cuddlefishie, line 322: Since all access to this vector is immutable, it doesn't need to be a `Vec<>` at all, it could be a `&'a [StabilityIndication]` which can be constructed in a stable way with [`core::slice::from_raw_parts()`](https://doc.rust-lang.org/stable/core/slice/fn.from_raw_parts.html), unlike a `Vec` which has an internally unstable layout that just happens to be (ptr, len, cap) for now. The parameter could then be a `(*const (), usize)` instead and then we don't need a `transmute`

bittide-rust-ffi/src/lib.rs
```diff
@@ -247,6 +216,86 @@ const _: () = assert!(
...
+        reframing_enabled: cfg.reframing_enabled,
+        wait_time: cfg.wait_time,
+        target_count: cfg.target_count,
+    };
+}
+
+unsafe fn vsi_from_ptr<'a>(ptr: *const ()) -> &'a [callisto::StabilityIndication] {
+    let usize_ptr: *const usize = mem::transmute(ptr);
+    let data_ptr: *const callisto::StabilityIndication = mem::transmute(usize_ptr.offset(1));
+    return mem::transmute(std::slice::from_raw_parts(data_ptr, *usize_ptr));
```
cuddlefishie, line 290: You could use `ptr.cast()` here, that should already work because the type annotations in the let bindings. And in the return you don't need the `mem::transmute`, right? The slice should already be the correct type?

kleinreact: Good catch. That's a leftover of the VSI wrapper.

Isn't `cast` just the same as using `as`?

cuddlefishie: `as` also allows to change mutability, `.cast` will only change the type!

bittide/src/Bittide/ClockControl/Callisto/Rust.hs
```diff
@@ -0,0 +1,86 @@
...
+        pConfig     <- malloc
+
+        poke pVSI $ VecS stabilityChecks
+        poke pState state
+        poke pDataCounts $ VecS (DataCountS <$> dataCounts)
+        poke pConfig config
+
+        callisto_rust
+          (castPtr pConfig)
+          (fromInteger $ toInteger mask)
```
cuddlefishie, line 74: would something like `fromIntegral` work here too maybe?

bittide/src/Bittide/ClockControl/Callisto/Types.hs
```diff
@@ -0,0 +1,259 @@
...
+    pokeByteOff p 8 _steadyStateTarget
+    pokeByteOff p 12 rfState
+
+newtype VecS (n :: Nat) a = VecS (Vec n a)
+
+type instance SizeOf (VecS n a) =
+  SizeOf Int + n * SizeOf a
+
+type instance Alignment (VecS n a) =
+  Alignment a
```
cuddlefishie, line 182: Hmm so this is tricky because there could be an `a` that has higher alignment than `Int`, so let's say 16! But later down there we first write the integer to the buffer and then all the elements at `n * sizeof a + sizeof int`, but that means we would end up with the elements starting at byte 4, would would be a wrong alignment!

Either we comment that it's only for types with alignment smaller than or equal to `Int` or we need to find a more generic version?

host-tools/callisto-lib/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// The controller remains in this state for the predefined
+    /// number of cycles with the assumption that the elastic buffers
+    /// of all other nodes are sufficiently stable after that time.
+    Wait {
+        /// Stored correction value to be applied at reframing time.
+        target_correction: f32,
+        /// Number of cycles to wait until reframing takes place.
+        cur_wait_time: u32,
+    },
+}
```
cuddlefishie, line 42: One thing to be careful with here is that at some point you cast a `*const ()` to a `*const ControlSt`, which contains a `ReframingState`. The issue is that there is no validation that the tag is valid, so you can easily end up with undefined behaviour if the tag is not 0, 1 or 2 and there's no nice way to detect that either. That's why the code in `callisto-rust-split` uses a struct and a manual `u32` tag field to check it's valid.
Since we are the ones that are building the data structure on the Haskell side it's fine, but there should be some comment about this that there is this possibility if it's misused.

host-tools/callisto-lib/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+        let (_, _, _, _, x, _) = unsafe { *CASTED_CONTROL_STATE };
+        x
+    }
+);
+
+const _: () = assert!(
+    12345u32 == {
+        let (_, _, _, _, _, x) = unsafe { *CASTED_CONTROL_STATE };
+        x
+    }
```
cuddlefishie, line 163: This could be done like this, right? Should do the same thing but not be split over so many different assertions? But that's just a small "beauty" suggestion :sweat_smile: 

```suggestion
const _: () = assert!(
    {
        let fields@(_, _, _, _, _, _) = unsafe { *CASTED_CONTROL_STATE };
        fields == (54321i32, 1u32, 3.185f32, 2u32, 3.321f32, 12345u32)
    }
);
```

host-tools/callisto-lib/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+///   a `VecS n (DataCountS m)
+/// - `control_state_ptr` needs to point to a valid memory address holding
+///   a `ControlSt`
+#[no_mangle]
+pub unsafe extern "C" fn __c_callisto_rust(
+    config_ptr: *const (),
+    availability_mask: u32,
+    stability_checks_ptr: *const (),
+    data_counts_ptr: *const (),
+    control_state_ptr: *mut (),
```
cuddlefishie, line 314: The C ABI on Sys-V for pointers is the same regardless of types AFAIK, so you could just put the types here that you expect, that saves you some casts later and makes the signature a bit more useful.

bittide-rust-ffi/src/Bittide/Simulate/RustFFI.hs
```diff
@@ -0,0 +1,18 @@
...
+{-# LANGUAGE ForeignFunctionInterface #-}
+
+module Bittide.Simulate.RustFFI (callisto_rust) where
+
+import Foreign.C.Types (CUInt(..))
+import Foreign.Ptr (Ptr)
+
+foreign import ccall safe "__c_callisto_rust" callisto_rust ::
+  Ptr () -> -- ^ config
+  CUInt  -> -- ^ availability_mask
```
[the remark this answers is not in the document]
kleinreact: It's technically not necessary, as [this constraint](https://github.com/bittide/bittide-hardware/blob/fbfa2e674f797b31c30a4a215a6764640b07ba1e/bittide/src/Bittide/ClockControl/Callisto/Rust.hs#L41) always ensures that the bits of the mask fit in. 

It's still kind of unclear to me why there is the `CUInt` newtype wrapping around `Word32` overall. The documentation says that the implementation is platform specific, but its 4 bytes on 64bit systems as well. Maybe its different on Windows, but that won't affect us then. Therefore, I prefer keeping it `CUInt`, as this is the type, which is esplicitly introduces to be used for FFI operations.

bittide/src/Bittide/ClockControl/Callisto/Rust.hs
kleinreact: `Integral` is for types, which support `div`/`mod` operations, so that's not really related to a mask. I guess the correct way would be to have a `BitPack` instance for `CUInt`, but that needs to be added to `clash-compiler` then. There is one for `CUShort`, but thats the only one for the `Foreign` types that in already :grin:.

bittide/src/Bittide/ClockControl/Callisto/Types.hs
kleinreact: We can add a constraint, which requires that the alignment of `a` is equals to that one of `Int`. Then we at least ensure correct usage at compile time.

host-tools/callisto-lib/src/lib.rs
kleinreact: You don't think that the

```Rust
/// Rust sibling of `Bittide.ClockControl.Callisto.Types.ReframingState`.
```
and the `# Safety` comments for `__c_callisto_rust` are enough?

host-tools/callisto-lib/src/lib.rs
kleinreact: It could, but then you don't  know any more which field was wrong if the assertion fails. Unfortunately, the compiler does not give you that information. Therefore, I usually prefer one assertion per property.

host-tools/callisto-lib/src/lib.rs
kleinreact: Mhm :thinking:, I'm not fully convinced yet ...

Then they don't match the pointer types of the foreign import in Haskell any more, which is kind of confusing. Personally, I would like to also conretize the types on the Haskell side, but they are defined in `bittide` and thus not available in `bittide-rust-ffi`. 

The whole idea of having `bittide-rust-ffi` separated is that this allows us to statically compile the Rust lib into the Haskell library providing a nice way of separation. However, since there is no Rust in `bittide-rust-ffi` anymore, we may alos move the foreign call directly to `bittide`. How do you think about that? 

Then we can provide concrete types at both ends, but need the custom `Setup.lhs` in `bittide` as well ...

bittide/src/Bittide/ClockControl/Callisto/Rust.hs
cuddlefishie: That's fair, thanks for checking!

bittide/src/Bittide/ClockControl/Callisto/Types.hs
cuddlefishie: Yeah I think that would be good!

host-tools/callisto-lib/src/lib.rs
cuddlefishie: The `# Safety` section only talks about allocations of appriopriate sizes, not about valid values. I guess it's enough, just something that I noticed while reading through the code :woozy_face:

host-tools/callisto-lib/src/lib.rs
cuddlefishie: That's a fair point, we can keep it that way then! :)

host-tools/callisto-lib/src/lib.rs
cuddlefishie: Devil's advocate in me says they already don't match because of `u32` vs `CUInt` :stuck_out_tongue:  But that's a valid consideration. In the end the casting is a no-op so it doesn't hurt, but also making the type signature a bit more useful would be nice!

Whether we should put all this into `bittide` and then keep the custom `Setup.lhs` I'm not sure. It would make sense I think since this package is really small and doesn't really do anything useful on its own, so probably putting it into `bittide` would make sense?

kleinreact: > Devil's advocate in me says they already don't match because of u32 vs CUInt

But `CUInt` is just a `Word32` [wrapped by a newtype](https://hackage.haskell.org/package/base-4.18.0.0/docs/Foreign-C-Types.html#t:CUInt), so why is this a missmatch?

kleinreact: I integrated `bittide-rust-ffi` directly into `bittide` now added type information to all the pointers.

bittide-rust-ffi/src/Bittide/Simulate/RustFFI.hs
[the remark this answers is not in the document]
kleinreact: According to my understanding `CUInt` is the right Haskell correspondant to `u32` in Rust. `Foreign.C.Types` just provides newtype wrappers to Haskell types, which indicate that the wrapped types are suitable for usage via FFI. As `CUInt` wraps `Word32` that's correct.

Note that this should not be mixed up with `Storable` instances, which define how to poke and peek things to/from the stack/heap. So this is independent of argument types that can be used via foreign calls.

host-tools/callisto-lib/src/lib.rs
cuddlefishie: > But CUInt is just a Word32 [wrapped by a newtype](https://hackage.haskell.org/package/base-4.18.0.0/docs/Foreign-C-Types.html#t:CUInt), so why is this a missmatch?

In practice it's not a mismatch, it's just that `Word32` means "always 32 bits" while `CUInt` means "whatever C considers an unsigned integer on the current platform". For all the platforms we care about it is 32 bits but just semantically it's sliiiightly different. With `Word32` you say "give me 32 bits please" while `CUInt` is more "give me something that can hold an unsigned integer that's probably big enough for what I need it".

But this is not a blocker at all, it's fine. But since we do care about it being exactly 32 bits as it's a bitmask I personally would prefer `Word32`.

kleinreact: That's a good point. I changed it to `Word32` then.

cuddlefishie (approved):

PR #362, opened 2023-07-20
Add cargo support to shake

lmbollen:
To include rust binaries in `bittide-instances`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -464,7 +464,7 @@ jobs:
         uses: actions/checkout@v3
 
       - uses: actions/cache@v3
-        name: Cache
+        name: Cargo Cache
         with:
           path: |
             ~/.cargo/bin/
@@ -477,6 +477,15 @@ jobs:
       - name: Building Examples
         run: cargo build --release
         working-directory: firmware-binaries/
+
+      - name: Archive generated cargo binaries
+        uses: actions/upload-artifact@v3
+        with:
+          name: cargo-binaries
+          path: |
+            _build/cargo/firmware-binaries/riscv32imc-unknown-none-elf/release/
+            !_build/cargo/firmware-binaries/riscv32imc-unknown-none-elf/release/*/*
+          retention-days: 14
 
   firmware-limit-checks:
     name: Firmware Limit Checks
@@ -493,7 +502,7 @@ jobs:
         uses: actions/checkout@v3
 
       - uses: actions/cache@v3
-        name: Cache
+        name: Cargo Cache
         with:
           path: |
             ~/.cargo/bin/
@@ -502,6 +511,11 @@ jobs:
             ~/.cargo/git/db/
             ./firmware-binaries/target/
           key: cargo-firmware-bins-cache-cachebust-2-${{ hashFiles('./firmware-binaries/Cargo.lock') }}
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: cargo-binaries
+          path: _build/cargo/firmware-binaries/riscv32imc-unknown-none-elf/release
 
       - name: Install `elf-limits`
         run: cargo install --git https://github.com/cuddlefishie/elf-limits
@@ -513,7 +527,7 @@ jobs:
             --data-mem-limit 64K \
             hello \
             fdt-read
-        working-directory: firmware-binaries/target/riscv32imc-unknown-none-elf/release
+        working-directory: _build/cargo/firmware-binaries/riscv32imc-unknown-none-elf/release/
 
 
   bittide-instances-doctests:
@@ -530,6 +544,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: cargo-binaries
+          path: _build/cargo/firmware-binaries/riscv32imc-unknown-none-elf/release
 
       - name: Set CI flags
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -475,9 +475,20 @@ jobs:
           key: cargo-firmware-bins-cache-cachebust-2-${{ hashFiles('./firmware-binaries/Cargo.lock') }}
 
       - name: Building Examples
-        run: cargo build --release
+        run: |
+          export CARGO_TARGET_DIR=../_build/cargo
```
cuddlefishie, line 479: It's possible to set the target directory per workspace in the `<workspace-dir>/.cargo/config` file like
```toml
[build]
target-dir = "<path-here>"
```
Maybe that's better because then *all* Cargo commands use that and not only those which have the right environment variable set?

bittide-shake/src/Clash/Shake/Cargo.hs
```diff
@@ -0,0 +1,31 @@
...
+import Prelude
+
+import System.Exit
+import System.IO
+import System.Process
+
+cargoBuildFirmwareProgram ::
+  FilePath ->
+  -- ^ Directory to perform the call in
+  Bool ->
```
martijnbastiaan, line 15: Could you replace this with:

```haskell
data CargoBuildMode = Release | Debug
```

That way this:

```haskell
liftIO $ cargoBuildFirmwareProgram "firmware-binaries" True
```

Turns into a much more descriptive:

```haskell
liftIO $ cargoBuildFirmwareProgram "firmware-binaries" Release
```

bittide-shake/src/Clash/Shake/Cargo.hs
```diff
@@ -0,0 +1,31 @@
...
+  -- ^ release mode?
+  IO ()
+cargoBuildFirmwareProgram workingDir release = do
+  let
+    createProc =
+      (proc "cargo" (["build"] <> ["--release" | release]))
```
martijnbastiaan, line 21:
```suggestion
      (proc "cargo" ("build" : ["--release" | release]))
```

martijnbastiaan: Also, don't we want to use `cargo.sh` @cuddlefishie ?

bittide-shake/src/Clash/Shake/Cargo.hs
```diff
@@ -0,0 +1,31 @@
...
+        { cwd = Just workingDir}
+  (exitCode, _stdoutOut, stderrOut) <- readCreateProcessWithExitCode createProc ""
+
+  case exitCode of
+    ExitSuccess -> do
+      pure ()
+    ExitFailure _ -> do
+      hPutStrLn stderr "Compilation failed"
```
martijnbastiaan, line 29: Could you include what command exactly failed? E.g., when something failed I'd much rather see:

```haskell
{-# LANUGAGE QuasiQuotes #-}

import qualified Clash.Util.Interpolate as I

[..]
  case exitCode of
    ExitSuccess -> do
      pure ()
    ExitFailure _ -> do
      hPutStrLn stderr [I.i|
        `#{cmd} #{args}` terminated with exit code `3`. Stderr:

          #{stderr}

        Stdout:

          #{stdout}
     |]
```

It's slightly grimy to use `Clash.Util.Interpolate`, but I don't think that's going away any time soon :upside_down_face:. (Also note that the quasiquoter will correctly indent if stdout/stderr is a multiline string.)

martijnbastiaan (changes requested):
In general agree, but.... using artifacts does lock us into using GitHub's infrastructure. With caching we could use a caching server on our CI runner for much improved speeds: https://github.com/tespkg/actions-cache. This is not an objection to this specific PR btw, just stating that we should think about how we should move forward now that everything grows larger and larger.

In any case, changes requested :).

bittide-shake/src/Clash/Shake/Cargo.hs
```diff
@@ -0,0 +1,31 @@
...
+      hPutStrLn stderr stderrOut
+      pure ()
```
martijnbastiaan, line 31: Don't forget to insert the exit code again / use `die`

bittide-shake/src/Clash/Shake/Cargo.hs
```diff
@@ -0,0 +1,44 @@
...
+cargoBuildFirmwareProgram ::
+  FilePath ->
+  -- ^ Directory to perform the call in
+  CargoBuildMode ->
+  -- ^ release mode?
+  IO ()
+cargoBuildFirmwareProgram workingDir release = do
+  let
+    createProc =
+      (proc "cargo" (["build"] <> ["--release" | release == Release]))
```
martijnbastiaan, line 26:
```suggestion
    cmd = "cargo"
    args = "build" : ["--release" | release == Release]
    createProc =
      (proc "cargo" args)
```

bittide-shake/src/Clash/Shake/Cargo.hs
```diff
@@ -0,0 +1,44 @@
...
+        { cwd = Just workingDir}
+  (exitCode, _stdoutOut, stderrOut) <- readCreateProcessWithExitCode createProc ""
+
+  case exitCode of
+    ExitSuccess -> do
+      pure ()
+    ExitFailure exitInt -> do
+      hPutStrLn stderr [I.i|
+        `cargo build #{["--release" | release == Release]}` terminated with exit code `#{exitInt}`.
```
martijnbastiaan, line 35:
```suggestion
        `#{cmd} #{args}` terminated with exit code `#{exitInt}`.
```

bittide-shake/src/Clash/Shake/Cargo.hs
cuddlefishie: That script is mostly for users who want to check/build their code in the project root, since this one takes a working directory there's no need to invoke that script

martijnbastiaan (changes requested):
Commit messages:

 * Lose the closing full stop for consistency's sake
 * Change CI cargo building => could you change this to something more descriptive? Maybe "Use artifacts to transfer Cargo cache"?

martijnbastiaan (approved):

PR #367, opened 2023-07-26
Ignore module names in probe names for HITL-tests

hiddemoll:
`vioProbe`s are now rendered in their own module, which causes module name to be prepended to the probe name. HITL-tests now ignore this module name by searching for VIO probes which contain `probe_test_start` rather than start with it.

### DIFF
--- a/bittide-shake/data/tcl/HardwareTest.tcl
+++ b/bittide-shake/data/tcl/HardwareTest.tcl
@@ -60,7 +60,7 @@ proc print_all_vios {} {
     set w_radix 5
     foreach probe $probes {
         set type [get_property type $probe]
-        set w_name_cur [string length [get_property name $probe]]
+        set w_name_cur [string length [get_property name.short $probe]]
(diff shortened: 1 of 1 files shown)


bittide-shake/data/tcl/HardwareTest.tcl
```diff
@@ -243,14 +243,14 @@ proc get_test_names {probes_file target_dict url} {
     set_property PROBES.FILE ${probes_file} $device
     refresh_hw_device $device
 
-    set start_probes [get_hw_probes probe_test_start*]
-    if {[expr [llength start_probes] == 0]} {
-        puts "No probes found with name 'probe_test_start*', which are needed to start tests"
+    set start_probes [get_hw_probes *probe_test_start*]
```
martijnbastiaan, line 246:
```suggestion
    set start_probes [get_hw_probes */probe_test_start*]
```
?

martijnbastiaan (changes requested):

PR #377, opened 2023-08-03
enable F-extension in processing elements

cuddlefishie:
This updates `clash-vexriscv` to a version that enables the F-extension (for single precision floating point operations). [PR here](https://github.com/clash-lang/clash-vexriscv/pull/5).

This also enables the use of the F extension for `firmware-binaries`


martijnbastiaan (changes requested):
Could you fix the EOF newlines? (Also, we should really lint this on CI.)

cuddlefishie:
I can make a commit in [`clash-lang/clash-vexriscv`](https://github.com/clash-lang/clash-vexriscv/) to change the line endings there. Since everything in `clash-vexriscv/` is a git subtree I guess it makes more sense to fix it upstream?

martijnbastiaan:
Oh gotcha. In that case it doesn't matter I guess? Didn't realize it was a subtree.

martijnbastiaan (approved):
In that case :-)

PR #386, opened 2023-08-10
Add HITL test for transceiver bringup

martijnbastiaan:
TODO: 

- [x] Fix copyright notices
- [x] Don't use `makeTopEntity` / don't use `Vec` in top entity to prevent warnings
- [x] Add SYNC_IN/SYNC_OUT logic
- [x] Add debug info (number of retries, time after success, etc.)
- [x] Add documentation to test
- [x] Improve and document reset logic in transceivers

### DIFF
--- a/.github/synthesis/all.json
+++ b/.github/synthesis/all.json
@@ -18,5 +18,6 @@
   {"top": "fincFdecTests",                 "stage": "test",   "targets": "Specific [-1]"},
   {"top": "simpleHardwareInTheLoopTest",   "stage": "test",   "targets": "All"          },
   {"top": "syncInSyncOut",                 "stage": "test",   "targets": "All"          },
+  {"top": "transceiversUpTest",            "stage": "test",   "targets": "All"          },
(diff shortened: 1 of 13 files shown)


hiddemoll (approved):
This PR is difficult to review, due to a lack of documentation, especially for the transceivers. However, we want this PR merged since others depend on working transceivers. I'll approve it, but we need documentation in a future PR.

martijnbastiaan:
What about now? :)

I can't say I'm really happy about the reset manager changes. While they work in practice (and links remain stable once they're detected as stable), it seems incredibly weird that 9/10 can get up without any fuss - but 1 or 2 need to be reset quite a bit until they work.

martijnbastiaan:
I've removed it from CI, as Vivado gets stuck in that environment for some reason :-(. I'll create an issue stating we should investigate it.

martijnbastiaan:
Nvm! https://github.com/bittide/bittide-hardware/pull/386/commits/c33ceb226f653a22b0773bb5a4c0dd512c9f87e4 :sparkles:

Issue #345, opened 2023-07-05
Use ILA dump to produce plots


martijnbastiaan:
Depends on:

- [ ] https://github.com/bittide/bittide-hardware/issues/342
- [ ] https://github.com/bittide/bittide-hardware/issues/341

Does not depend on implementation software controlled clock control.

martijnbastiaan:
Done https://github.com/bittide/bittide-hardware/pull/399

PR #399, opened 2023-08-29
ILA based plot generation

kleinreact:
Implementation for #345.

A short overview of how data collection for the plots and how plotting works:

#### Setup
  * Clock control starts synchronized for all nodes `5s` after all the transceivers of the last node are up.
  * If the transceivers of the other nodes are not ready then, they fail the test.
  * The last node uses its stable clock to produce a synchronization trigger every `5ms` (output via `SYNC_OUT`)
  * All nodes synchronize on these triggers via `SYNC_IN`.
  * The time between these triggers is measured using each node's individual stable clock.
  * This offers a `global timestamp` consisting of the number of global triggers received and the clock cycles of the stable clock in between.
  * Additionally, each node produces a `local timestamp` that counts the cycles of the dynamic clock (starting with clock control).
  * During the first second, `FINC`/`FDEC`s are suppressed (i,e., not passed to the clock boards). This serves as a calibration period to measure the initial drift of the dynamic clocks  versus the "global clock". Clock control should not be affected by this.
  * Data is dumped via an ILA, which stores up to `16384` captures (more does not fit currently)

#### Each capture of the ILA dump includes
  * the `capture condition flag` indicating whether it is
    * the initial trigger (synchronized),
    * the calibration end (synchronized),
    * or a data change (not synchronized),
  * the `global timestamp`,
  * the `local timestamp`,
  * the `datacounts` of the elastic buffers including their `stable` and `settled` status,
  * the "`SpeedChange`", and
  * every change of the `reframing mode` (`detect`, `wait`, or `done`).

#### Data changes are triggered by
  * changes of the `stable` or `settled` indicators for any of the elastic buffers
  * `n` accumulated `FINC`/`FDEC`s (currently `n = 3`)
     _("accumulated" here means that if both `FINC` and `FDEC` are requested after each other, then they cancel each other out and are not reported. Only `FINC`/`FDEC`s are reported that haven't canceled out before they exceed the `n` boundary. Thus, for example `FINC`, `FINC`, `FDEC`, `FDEC`, ... is not reported for `n > 2` as the first two `FINC`s don't exceed the boundary `n`.)_
  * changes in the `reframing mode`.

#### Plots are produced as follows
* the elastic buffer plots are generated the same way as for the software based simulation report using the "global timestamp" (converted to femtoseconds) as the time reference (x-axis).
* the clock plots are also generated with the `global timestamp` as a time reference (x-axis) and report clock drift of the local timestamp against the global on calculated via 
$$\frac{\Delta t_G - (\Delta c_L \cdot (p_L + a_{\uparrow / \downarrow} \cdot t_S))}{\Delta t_G}$$
where
    * $\Delta t_G$ is the time difference of the "global static clock" between two captures in femtoseconds,
    * $\Delta c_L$ is the number of cycles that have been counted on the "local dynamic clock" between two captures,
    * $p_L$ is the default period of a local clock cycle in femtoseconds (aka `clockPeriodFs (Proxy @GthTx)`),
    * $a_{\uparrow / \downarrow}$ is the accumulated number of all `FINC`/`FDEC`s obtained via assigning `FINC ↦ 1` and `FDEC ↦ -1` and integrating them up to the current capture, and
    * $t_S$ is the step size per `FINC`/`FDEC` (aka `cccStepSize clockControlConfig`).

   This gets us the drift of the local clock against the global one per clock cycle of the global clock in femtoseconds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,12 +225,37 @@ jobs:
     name: Generate clock control simulation report
     runs-on: [self-hosted, compute]
     if: always()
-    needs: [build, lint, elastic-buffer-sim-topologies-matrix, elastic-buffer-sim-topologies]
+    defaults:
+      run:
+        shell: git-nix-shell {0} --pure
+    needs: [build, lint, elastic-buffer-sim-topologies-matrix, elastic-buffer-sim-topologies, bittide-instances-hardware-in-the-loop]
 
     container:
       image: ghcr.io/clash-lang/nixos-bittide-hardware:2023-06-30
 
     steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set CI flags
+        run: |
+          cp .github/cabal.project cabal.project.local
+
+      - name: Update Cabal index info
+        run: |
+          cabal update
+          cabal freeze
+
+      - name: Cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cabal-nix/store
+
+          key: packages-cachebust-4-${{ hashFiles('cabal.project.freeze', 'cabal.project') }}
+          restore-keys: packages-cachebust-3-
+          fail-on-cache-miss: true
+
       - name: Download simulation artifacts
         uses: actions/download-artifact@v3
         with:
@@ -244,17 +269,28 @@ jobs:
           # the required artifacts instead.
           path: artifacts
 
-      - name: Generate final report
-        shell: git-nix-shell {0} --pure
+      - name: Generate final simulation report
         run: |
           set +f # Enable globbing
           pdfunite artifacts/*/report.pdf report.pdf
 
-      - name: Upload report
+      - name: Upload simulation report
         uses: actions/upload-artifact@v3
         with:
           name: Clock Control Simulation Report
           path: report.pdf
+          retention-days: 14
+
+      - name: Generate HITL based plots
+        run: |
+          mkdir hitl-plots
+          ./cargo.sh build --release && cabal run -- elastic-buffer-sim:plot artifacts/_build-fullMeshHwCcTest-debug/Bittide.Instances.Tests.FullMeshHwCc.fullMeshHwCcTest/ila-data/probe_test_start/ hitl-plots Bittide_Instances_Tests_FullMeshHwCc_fullMeshHwCcTest_callistoClockControlWithIla
+
+      - name: Upload HITL based plots
+        uses: actions/upload-artifact@v3
+        with:
+          name: Clock Control HITL Plots
+          path: hitl-plots
           retention-days: 14
 
   elastic-buffer-sim-tests:
--- a/bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
+++ b/bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
@@ -26,16 +26,19 @@ module Bittide.Instances.Tests.FullMeshHwCc where
 import Clash.Prelude (withClockResetEnable)
 import Clash.Explicit.Prelude
 
+import Control.Arrow ((***), second)
+
 import Bittide.Arithmetic.Time
 import Bittide.ClockControl
 import Bittide.ClockControl.Callisto
 import Bittide.ClockControl.Si5395J
-import Bittide.ClockControl.Si539xSpi
+import Bittide.ClockControl.Si539xSpi (ConfigState(Error, Finished), si539xSpi)
 import Bittide.Counter
 import Bittide.ElasticBuffer (sticky)
 import Bittide.Instances.Domains
 import Bittide.Transceiver
 
+import Bittide.Instances.Tests.FullMeshHwCc.IlaPlot
 import Bittide.Instances.MVPs (stickyBits, speedChangeToPins, FINC, FDEC)
 
 import Clash.Class.Counter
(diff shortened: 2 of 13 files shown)


kleinreact:
Here is what we get with the current implementation:

#### Elastic Buffer Plot
elasticbuffers
Looks good overall. Nevertheless, we can observe that
* 49 buffers are a lot for a single plot (this gets better once we support smaller topologies).
* The initial datacounts are quite diverse. This probably is a result of the delay and order of programming the boards. The nodes which are first in the row get large counts, because they have more time to fill up than later nodes.
* The parameters for the stability detection window are not optimal yet. Basically, the run for this plot failed on one node by running into the 60s limit because of that.

#### Clocks Plot
clocks
This looks quite different against what we are used to from simulation. The drift seems to be mostely affected by the `FINC`/`FDEC`s (also plotted below). This may be necessary to compensate for the delays on the network, but this is only an estimate. Things look more like expected, if we disable the `FINC`/`FDEC` correction of the period, i.e., if we set $a_{\uparrow \ \downarrow} = 0$ (see below). However, the correction should be correct. This huge effect of the `FINC`/`FDEC`s also makes sense, since currently $t_S = 8fs$.

_(Note that the peak at about 35s results from an overflow of the `local timestamp`, which is currently limited to 32bits by the `xpmCdcGray` component. Depending on where the overflow happens between two captures it may be visible or not.)_

#### Plot of $a_{\uparrow \ \downarrow}$
fincfdecs

#### Plot with $a_{\uparrow \ \downarrow} = 0$
clocks_uncorrected

.github/workflows/ci.yml
```diff
@@ -244,19 +244,32 @@ jobs:
...
         with:
           name: Clock Control Simulation Report
           path: report.pdf
           retention-days: 14
 
+      - name: Generate HITL based plots
+        shell: git-nix-shell {0} --pure
+        run: |
+          mkdir hitl-plots
+          cabal run elastic-buffer-sim:plot -- artifacts/Bittide.Instances.Tests.FullMeshHwCc.fullMeshHwCcTest/ila-data/probe_test_start/ hitl-plots Bittide_Instances_Tests_FullMeshHwCc_fullMeshHwCcTest_callistoClockControlWithIla
```
martijnbastiaan, line 264: Once @hiddemoll is done, this can be part of a post-processing step in Shake.

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
```diff
@@ -163,8 +190,8 @@ goFullMeshHwCcTest refClk sysClk rst rxns rxps miso =
   fincFdecIla :: Signal Basic125 ()
   fincFdecIla = ila
     (ilaConfig $
-         "trigger"
-      :> "capture"
+         "trigger_0"
+      :> "capture_0"
```
martijnbastiaan, line 194: How come? Does the TCL accept this?

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc/IlaPlot.hs
```diff
@@ -0,0 +1,268 @@
...
+import Clash.Cores.Xilinx.Ila (IlaConfig(..), Depth(..), ila, ilaConfig)
+
+import Data.Maybe (isJust, fromMaybe)
+
+type AccWindowHeight = 3 :: Nat
+type SyncPulsePeriod = Milliseconds 5
+type SyncPulseIndex dom = Index (PeriodToCycles dom SyncPulsePeriod)
+
+-- | A global timestamp consists of the number of synchronized pulses
+-- received and the number of cycles of the local stable clock
```
martijnbastiaan, line 43: Do you reset this value every time you get a global tick? It seems to me that a good timestamp would be `(global ticks, local ticks since last global tick)`. What does continued counting get you?

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc/IlaPlot.hs
```diff
@@ -0,0 +1,268 @@
...
+
+-- | Accumulates over multiple @FINC@/@FDEC@s, i.e. if both @FINC@ and
+-- @FDEC@ are requested after each other, then they cancel each other
+-- out and are not reported. Correspondingly, only @FINC@/@FDEC@s are
+-- reported that haven't canceled out before they exceed the @n@
+-- boundary. Thus, for example @FINC@, @FINC@, @FDEC@, @FDEC@, ... is
+-- not reported for @n > 2@ as the first two @FINC@s don't exceed the
+-- boundary @n@.
+accwindow ::
+  forall height dom.
```
martijnbastiaan, line 93: Could you explain the use of this function? I'm not sure why you'd wanna use this.

martijnbastiaan (commented):
@cuddlefishie Notice that @kleinreact changes the bit-definition of `SpeedChange` here. We probably need to change Rust code if we rebase on top of this.

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
hiddemoll: I suppose you mean the suffix of the probe names, in which case yes. The TCL accepts any probe with the name `*/trigger*`, as long as there is only one. The same goes for `/capture*`.

kleinreact: We need this because of the other ILA. Vivado doesn't like if they share names.

.github/workflows/ci.yml
kleinreact: I put here, because this step downloads all artifacts anyway.

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc/IlaPlot.hs
kleinreact: Don't get confused, this is the timer for producing the synchronized plot.

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc/IlaPlot.hs
kleinreact: see above
>   * `n` accumulated `FINC`/`FDEC`s (currently `n = 3`)
     _("accumulated" here means that if both `FINC` and `FDEC` are requested after each other, then they cancel each other out and are not reported. Only `FINC`/`FDEC`s are reported that haven't canceled out before they exceed the `n` boundary. Thus, for example `FINC`, `FINC`, `FDEC`, `FDEC`, ... is not reported for `n > 2` as the first two `FINC`s don't exceed the boundary `n`.)_

martijnbastiaan: I'm still not sure _why_ you'd use this. Is this to prevent too much data ending up in the ILA? I.e., it's a compression technique?

kleinreact: Yes

martijnbastiaan (approved):

bittide/src/Bittide/ClockControl.hs
```diff
@@ -124,10 +125,10 @@ targetDataCount = 0
 
 -- | Safer version of FINC/FDEC signals present on the Si5395/Si5391 clock multipliers.
 data SpeedChange
-  = SpeedUp
+  = NoChange
   | SlowDown
-  | NoChange
-  deriving (Eq, Show, Generic, ShowX, NFDataX, Enum)
+  | SpeedUp
+  deriving (Eq, Show, Generic, BitPack, ShowX, NFDataX)
```
martijnbastiaan, line 131: This change should be reflected in 

```
instance (SizeOf SpeedChange ~ 4, Alignment SpeedChange ~ 4)
```

martijnbastiaan (changes requested):

martijnbastiaan (approved):

PR #400, opened 2023-08-29
Add a wishbone accessible timing utility.

lmbollen:
This peripheral and rust module enable the RISC-V core to keep track of time*.
Useful for for example software based clock control.


firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,304 @@
...
+// SPDX-License-Identifier: Apache-2.0
+/*! Time structures.
+
+The `time` module contains structures used to get some sense of time.
+Keep in mind that in a Bittide system, clock frequencies are non-constant and thus there
+is no relation to wall clock time. However, we can use this module for a best effort
+approximation of wall clock time.
+
+ - [Instant] is used to represent absolute time.
+ - [Duration] is used to represent relative time.
```
cuddlefishie, line 12: I believe it needs ``[`Instant`]`` for rustdoc to pick up on the item link?

cuddlefishie (changes requested):

bittide/src/Bittide/Wishbone.hs
```diff
@@ -327,3 +327,22 @@ fifoWithMeta depth@SNat = Circuit circuitFunction
...
+  , 2 <= addrW
+  , KnownNat nBytes
+  , 1 <= nBytes) =>
+  Circuit (Wishbone dom 'Standard addrW (Bytes nBytes)) ()
+timeWb = case (cancelMulDiv @nBytes @8) of
+  Dict -> Circuit go
+ where
+  go (wbM2S, _) = (wbS2M, ())
+   where
+    (cnt, wbS2M) = registerWb CircuitPriority (0 :: Unsigned 32) wbM2S
```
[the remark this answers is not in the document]
lmbollen: Good catch, thanks!

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,321 @@
+// SPDX-FileCopyrightText: 2023 Google LLC
+//
+// SPDX-License-Identifier: Apache-2.0
+/*! Time structures.
+
+The `time` module contains structures used to get some sense of time.
+Keep in mind that in a Bittide system, clock frequencies are non-constant and thus there
+is no relation to wall clock time. However, we can use this module for a best effort
+approximation of wall clock time.
```
martijnbastiaan, line 9: While this is true, I'm not sure how much it matters. I think it needs a note that says that while clocks are dynamic in a Bittide system they're not expected to change more than a couple of ppm.

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,321 @@
...
+impl core::cmp::PartialOrd for Instant {
+    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
+        Some(self.to_micros().cmp(&other.to_micros()))
+    }
+}
+
+impl Instant {
+    /// Construct a new `Instant` of 0 clock cycles at the frequency of the timing peripheral.
+    pub const fn zero(frequency: u64) -> Instant {
+        // Should we panic if the frequency is zero on construction?
```
martijnbastiaan, line 48: Probably!

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,321 @@
...
+}
+
+/// A representation of a relative time, stored in microseconds.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Duration {
+    micros: u64,
+}
+
+impl Duration {
+    pub const ZERO: Duration = Duration { micros: 0 };
```
martijnbastiaan, line 166: Bit too arbitrary IMO, can we remove this?

firmware-binaries/examples/hello/src/main.rs
```diff
@@ -7,19 +7,56 @@
 
 use ufmt::uwriteln;
 
-use bittide_sys::uart::Uart;
+use bittide_sys::{time::Clock, time::Duration, uart::Uart};
 
 #[cfg(not(test))]
 use riscv_rt::entry;
+const FREQUENCY: u64 = 200e6 as u64;
```
martijnbastiaan, line 14: Could you add a register that exposes the frequency? This seems error prone.

martijnbastiaan (changes requested):
Couple of small changes, but the "changes requested" is for an additional register that exposes the CPU's frequency. Looks good overall!

firmware-support/bittide-sys/src/time.rs
lmbollen: Changed to:
```
While clocks are dynamic in a Bittide System, it is expected that the deviation from
wall clock time is no worse than any on-board static clock source and the effect is 
insignificant for timing purposes.
```

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,321 @@
...
+use core::cmp;
+use core::ops;
+
+/// A representation of an absolute time value.
+///
+/// The `Instant` type is a wrapper around a `u64` value that
+/// represents a number of clock cycles, monotonically increasing
+/// since an arbitrary moment in time, such as system startup.
+///
+/// * A value of `0` is inherently arbitrary.
```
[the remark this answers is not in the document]
lmbollen: Simplified the text, PTAL

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,330 @@
...
+        self.counter_last
+    }
+
+    pub fn get_frequency(&mut self) -> u64 {
+        self.global_cycles.get_frequency()
+    }
+    /// Gets the current current number of ticks in usize, this will overflow during runtime.
+    fn get_ticks(&self) -> usize {
+        unsafe { self.counter_now.read_volatile() }
+    }
```
cuddlefishie, line 329: I'm not having a good feeling about this, because the documentation of the Wishbone component says that overflows needs to be accounted for and it seems like that's not happening here anywhere!

Would it make sense to have something like `get_ticks_since_last_check()` that could then try to recover the number of ticks even across (one) overflow? If we need this timer component to also be usable in longer running applications we need to think about how we want to deal with overflows and representing time in general.

We can keep track of changes of number of clock cycles as long as it's not more than one "overflow space", but we might want to refer to time with `Instant` or `Duration` that's longer than that? We can use `u64` or `u128` just fine, it will just be slower.

Just something to think about because right now we're not doing anything :see_no_evil:

lmbollen: I discussed this privately, the overflow gets handled in `update()` but I'll remove this function and relocate the functionality such that it's paired with the overflow handling.

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,353 @@
...
+
+impl ops::Sub<Instant> for Instant {
+    type Output = Duration;
+
+    fn sub(self, rhs: Instant) -> Duration {
+        Duration::from_micros(self.to_micros() - rhs.to_micros())
+    }
+}
+
+/// A representation of a relative time, stored in microseconds.
```
martijnbastiaan, line 163:
```suggestion
/// A representation of a relative time, stored in nanoseconds.
```
?

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,353 @@
...
+#[derive(uDebug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Duration {
+    nanos: u64,
```
martijnbastiaan, line 166:
```suggestion
    nanos: u64,
```

IMO either `ns` or `nanoseconds`, I don't think I've ever seen it abbreviated as `nanos`.

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,353 @@
...
+use ufmt::derive::uDebug;
+/// A representation of an absolute time value.
+///
+/// The `Instant` type is a wrapper around a `u64` value that represents the number
+/// of clock cycles since system startup.
+///
+#[derive(uDebug, Copy, Clone)]
+pub struct Instant {
+    clock_cycles: u64,
+    frequency: u64,
```
martijnbastiaan, line 27: Shouldn't `frequency` be a property of `Clock`?

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,353 @@
...
+// SPDX-License-Identifier: Apache-2.0
+/*! Time structures.
+
+The `time` module contains structures used to get some sense of time.
+While clocks are dynamic in a Bittide System, it is expected that the deviation from
+wall clock time is no worse than any on-board static clock source and the effect is
+insignificant for timing purposes.
+
+ - [`Instant`] is used to represent absolute time.
+ - [`Duration`] is used to represent relative time.
```
martijnbastiaan, line 12:
```suggestion
 - [`Instant`] is used to represent absolute time.
 - [`Duration`] is used to represent relative time.
```

Perhaps go with Python terminology: `TimeStamp` and `TimeDelta`. Or, if this it is common in rust to scope it to the module: `Stamp` and `Delta`?

martijnbastiaan (changes requested):
As discussed offline, the current implementation is complex in software and imposes restrictions on how the API is used (i.e., it **must** be called every X cycles lest it gives wrong answers).

We settled on the following memory map:

| Address | Access | Name       | Description                                                                 |
|---------|--------|------------|-----------------------------------------------------------------------------|
| 0       | WO     | freeze     | Write '1' freeze an internal clock cycle counter to the following registers |
| 4       | RO     | cycles_msb | MSB part of Unsigned 64, representing number of cycles since last CPU reset |
| 8       | RO     | cycles_lsb | LSB part of Unsigned 64, representing number of cycles since last CPU reset |

firmware-support/bittide-sys/src/time/self_test.rs
```diff
@@ -0,0 +1,150 @@
+use crate::time::Clock;
+use crate::time::Duration;
+use crate::time::Instant;
+
+/// Tests for the time module.
+/// It receives a pointer to the timing peripheral and returns a list of tuples containing the
+/// name of the test and a boolean indicating if the test passed.
+pub fn self_test(clock: Clock) -> [(&'static str, bool); 13] {
```
cuddlefishie, line 8: I'm not 100% sure how I would do those tests. They can't be run with the usual testing framework, but they aren't really part of the API either, so I'm not sure what the best place would be. Not really actionable, just something I think I'd like to think about a bit more.

Also like with the `Clock::self_test()` function I would recommend a `-> impl Iterator<Item = (&'static str, bool)>` return instead

firmware-support/bittide-sys/src/time.rs
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    /// Returns:
+    /// - The frequency of the time counter as a `u64`.
+    pub fn get_frequency(&self) -> u64 {
+        unsafe { self.frequency.read_volatile() }
+    }
+
+    pub fn self_test(&self) -> [(&str, bool); 13] {
+        self_test::self_test(Clock::new(self.freeze_count as *const u32))
+    }
```
cuddlefishie, line 364: I'm not sure if this should be a "member function" of this type or if it should be a free-standing function in the `self_test` module instead. If we add new tests this return type would change, which would be an API breaking change.

Maybe a `-> impl Iterator<Item = (&'static str, bool)>` is better?

firmware-support/bittide-sys/src/time/self_test.rs
```diff
@@ -0,0 +1,150 @@
...
+    let time1 = clock.elapsed();
+    time1 - target < Duration::from_micros(100)
+}
+
+/// Create a Duration of 1 hour, convert back to one hour and check if it's the same.
+pub fn duration_hour(_: Clock) -> bool {
+    let duration = Duration::from_hours(1);
+    let hours = duration.to_hours();
+    hours == 1 && hours != 0
+}
```
cuddlefishie, line 88: This function (and all following ones) don't actually need a `Clock` instance, at most they need a frequency. This means that all those functions be run as unit tests on the host too. For the ones needing a frequency, we could just give it a constant value and also run them on the host as unit tests. Anything only using `Duration` or `Instant` doesn't need to be tested on the hardware but can be run on the host.

firmware-support/bittide-sys/src/time/self_test.rs
lmbollen: In that case I think we should just keep the testing module completely separate.
We can include it whenever we deem it necessary and remove it when don't.

I'll take a look at the return types.

firmware-support/bittide-sys/src/time.rs
lmbollen: I'll make it free standing only

firmware-support/bittide-sys/src/time/self_test.rs
lmbollen: Yes I'm aware, I added it as an argument anyway to keep the type signature the same, allowing me to iterate over the functions in a uniform way.

Would you prefer I separate those iterators into functions with clocks and functions without clocks?

martijnbastiaan (approved):
bjoetiful

PR #403, opened 2023-09-08
Add software based clock control (milestone 4)

cuddlefishie:
This PR adds a firmware binary that performs the clock control calculations. It also adds a hardware-in-the-loop test showing that the clock control works as expected.

Because none of the examples that we actually use make use of the character device abstraction, it was removed. With it the default panic handler that used that character device was also removed. Every binary now needs to install their own panic handler, but given that different binaries might use different methods of reporting/responding to panics this seems like a valid change.

The `fdt-read` example wasn't used anywhere and depended on the character device, so it was removed.

Note: the only "real" difference between hardware clock control and software clock control is that the latter runs about three times as slow. That is, it pushes an update to the clock board every 3 microseconds, instead of every microsecond.

# Software clock control plots
Screenshot from 2023-09-26 17-08-29

# Hardware clock control plots (for comparison)
Screenshot from 2023-09-26 17-08-04


martijnbastiaan (approved):
We should probably deduplicate similar bits between our hardware-only and software-only instance, but I think this more than suffices for now!

PR #406, opened 2023-09-19
Fix `FullMeshHwCc` resets

kleinreact:
This PR introduces a reset line for the `FullMeshHwCC` tests, which is synchronized among all nodes via `SYNC_OUT` and `SYNC_IN`. 

The components of `FullMeshHwCc` are now reset in two stages. The first stage is driven by `startTest` (such that it is reset on `startTest` being low) and is used to produce the `SYNC_OUT` signal and read it back via `SYNC_IN`. The communicated data then is used to establish the **synchronized** reset, which is asserted whenever `SYNC_IN` is low for more than `5ms`. If not, then `SYNC_IN` also synchronizes the global time stamp via flipping every `5ms`. The reset line that gets recovered from `SYNC_IN` then drives test components themselves via offering the reset line for the second stage.

By that, the implementation also offers a synchronized end trigger, which can be used to finalize the tests simultaneously.

### DIFF
--- a/bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
+++ b/bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
@@ -327,57 +327,67 @@ fullMeshHwCcTest refClkDiff sysClkDiff syncIn rxns rxps miso =
   (sysClk, sysLock0) = clockWizardDifferential (SSymbol @"SysClk") sysClkDiff noReset
   sysLock1 = xpmCdcSingle sysClk sysClk sysLock0 -- improvised reset syncer
   sysRst = unsafeFromActiveLow sysLock1
+    `orReset` unsafeFromActiveLow startTest
+  --
+  -- 'syncOutGenerator' is used to drive the 'SYNC_OUT' signal, which
+  -- is only connected for the last node in the network and wired back
+  -- to 'SYNC_IN' of all nodes from there.
+  --
+  -- Note that all nodes are in reset before their local 'startTest' VIO
+  -- signal gets asserted, as 'startTest' is directly driving 'sysRst'.
+  -- Thus, for the other nodes to capture the 'SYNC_OUT' signal correctly,
+  -- the node receiving the `startTest` rising edge last must be the one
+  -- with it's 'SYNC_OUT' physically connected to the 'SYNC_IN' of all
+  -- nodes in the network. This assumption is tested by
+  -- 'Bittide.Instances.Tests.SyncInSyncOut'.
+  syncOut =
+      dflipflop sysClk
+    $ syncOutGenerator sysClk startTest
+    $ trueFor5s sysClk testRst allUp
 
-  syncIn1 = (startTest .&&.)
-    $ unsafeToActiveLow
+  -- first synchronize SYNC_IN to the local clock and filter from
+  -- potential glitches
+  syncIn1 =
+      unsafeToActiveLow
     $ resetGlitchFilter (SNat @1024) sysClk
     $ unsafeFromActiveLow
     $ xpmCdcSingle sysClk sysClk syncIn
 
+  -- generate a pulse on every change of SYNC_IN
   syncInChangepoints =
-    changepoints sysClk syncStartRst enableGen syncIn1
+    changepoints sysClk sysRst enableGen syncIn1
 
-  syncInRst =
-      unsafeFromActiveLow
-    $ sticky sysClk sysRst
-    $ isRising sysClk sysRst enableGen False
-      syncIn1
+  -- recover the activity and readiness states from SYNC_IN
+  (syncActive, syncStart) = unbundle $ syncInRecover sysClk sysRst syncIn1
 
-  testRst = sysRst `orReset` syncInRst
+  -- tests are reset with on `sysRst` or if not synchronously active
+  testRst = sysRst `orReset` unsafeFromActiveLow syncActive
 
-  syncStart =
-      sticky sysClk sysRst
-    $ isFalling sysClk sysRst enableGen False
-      syncIn1
-
-  syncStartRst = unsafeFromActiveLow syncStart
-
-  startBeforeAllUp = sticky sysClk sysRst
+  -- checks that tests are not synchronously start before all
+  -- transceivers are up
+  startBeforeAllUp = sticky sysClk testRst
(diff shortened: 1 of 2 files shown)


martijnbastiaan (changes requested):
Great, thanks! I think this solves my problems, but two things:

1. `syncOut` needs to be registered. (See review comment)
2. Like we discussed on Slack, please make the circuit (trivially) revertible to its initial state. I think this should suffice:

```haskell
sysRst = unsafeFromActiveLow sysLock1 `orReset` unsafeFromActiveLow startTest
```

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
```diff
@@ -328,53 +328,64 @@ fullMeshHwCcTest refClkDiff sysClkDiff syncIn rxns rxps miso =
...
 
-  syncIn1 = (startTest .&&.)
-    $ unsafeToActiveLow
+  -- use the 'syncOutGenerator' to generate the SYNC_OUT signal, which
+  -- is only physically connected on the last node in the network,
+  -- while that output is read by all nodes in the network via
+  -- SYNC_IN.
+  syncOut =
+    syncOutGenerator sysClk sysRst enableGen
+      (startTest, trueFor5s sysClk sysRst allUp)
```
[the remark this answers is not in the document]
kleinreact: We don't need to introduce a delay. I solved the problem more elegantly now.

kleinreact:
> 2. Like we discussed on Slack, please make the circuit (trivially) revertible to its initial state. I think this should suffice:
> 
>   ```sysRst = unsafeFromActiveLow sysLock1 `orReset` unsafeFromActiveLow startTest```

That's still dangerous, because `startTest` is asserted at different times for each node. Hence, if we do so, we may miss to detect the synchronized `startTest` properly, because nodes are waiting in their `sysRst` for too long.

I updated it now such that every component using the current `sysRst` is basically independent from being reset between two tests. 

For everything that's dependent on being reset between two tests, `testRst` should be used instead.

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
```diff
@@ -327,57 +327,65 @@ fullMeshHwCcTest refClkDiff sysClkDiff syncIn rxns rxps miso =
...
+  -- to 'SYNC_IN' of all nodes from there.                                 |
+  --                                                                       |
+  -- Note that all nodes are in reset before their local 'startTest' VIO   |
+  -- signal gets asserted, as 'startTest' is directly driving 'sysRst'. ----
+  -- Thus, for the other nodes to capture the 'SYNC_OUT' signal correctly,
+  -- the node receiving the `startTest` rising edge last must be the one
+  -- with it's 'SYNC_OUT' physically connected to the 'SYNC_IN' of all
+  -- nodes in the network. This assumption is tested by
+  -- 'Bittide.Instances.Tests.SyncInSyncOut'.
+  syncOut = syncOutGenerator sysClk startTest
```
martijnbastiaan, line 343: > (See previous comment)
>>  We don't need to introduce a delay. I solved the problem more elegantly now.

How did you fix the potential glitches on `syncOut`? As a rule of thumb, any combinatorial logic connected to a pin can close logic. (This isn't entirely true, but it's so hard to predict that it is good to keep this as a rule.)

martijnbastiaan (changes requested):
LGTM, but I don't see how you fixed the potential glitchy `SYNC_OUT`. Could you elaborate?

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs
kleinreact: [That line](https://github.com/bittide/bittide-hardware/blob/6335a6ebdf07b4aacf24599248cb422e4302bef8/bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc/IlaPlot.hs#L256C4-L256C4) should be fixing it, shouldn't it?

martijnbastiaan: That could in theory create even more glitches! The problem here is that signals are allowed to "settle" between their hold and setup window. That is, the window marked with `XXXX` in this diagram:

setup and hold timing checks

So _in theory_ the VIO could output some complicated logic, making it move `startTest` to various values before settling on its final value. The same goes for the output of your mealy machine: its implemented using a bunch of muxes whose outputs might be mapped two different path lengths causing race conditions / glitches. In general we don't care, because glitches are okay within that time window / in the synchronous abstraction. However, when connected to a pin this makes the other side potentially see all the in-between states!

Check out [Glitches in combinational circuits ](https://vlsiuniverse.blogspot.com/2017/10/glitches-in-combinational-circuits.html) for a better explanation.

Because even trivial circuits can trigger glitches, we generally just use a registered output (`delay`) unless there is a very good reason to believe the circuit doesn't produce glitches + an extra register would mess up the design.

martijnbastiaan: As an example of how this can go wrong, this function:

https://github.com/bittide/bittide-hardware/blob/1338fd3d95a297a36a950dba9f7ed97f6eabe0c4/bittide/src/Bittide/ClockControl.hs#L187-L188

Used to _not_ have a `dflipflop`, which made the clock boards see very short FINC/FDEC pulses. This messed up clock control quite a bit..

kleinreact: Alright, gotcha. So this is basically not releated to `syncOutGenerator` at all, because `SYNC_OUT` is really directly driving the output pin here. I always thought this is handled somewhere later. So [in fact it wasn't even handled correctly](https://github.com/bittide/bittide-hardware/blob/f23b4e80838d949ba38a8828e8b1539d129c5324/bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc.hs#L304) before merging #399 already.

I alwasy misunderstood this in the sense that `syncOutGenerator` would require even another additional latch beside that one for the output  pin.

martijnbastiaan: Right, `syncOutGenerator` is fine on its own. Though you could make the argument that because its, well, the SYNC_OUT generator, it should just produce glitchless signals.

martijnbastiaan (approved):

Issue #412, opened 2023-09-27
Improve and investigate clock control hardware plots


martijnbastiaan:
Though we currently produce plots for clock control on the boards, we:

1. We don't use intuitive units (i.e., we should use PPM)
2. The local timestamps are currently limited to 32 bits due to the limits and necessity of the CDC Gray. This causes overflows that are visible within the plots, but which can be avoided by only reporting the cycles since the last capture instead.
3. Due to the size limitations of 3, it might be useful to generate a capture on a regular interval, even if the clock remains stable. This offers a more robust plot for the elastic buffers and allows to bound the number of cycles between two captures.
4. The bit sizes of the local and global time stamps are chosen arbitrarily. With the adaption of 4 they can be reduced substantially saving memory for the dumps.
5. The bit sizes of the dumped elastic buffer bounds can be also reduced via only reporting the difference against the last capture. A smaller bound needs to estimated for that, which should work most of the time. If it still gets exceeded in some cases we can just trigger a capture to reset it to zero in that case.
6. We should produce plots in some "baseline" scenarios and see if they match our expectations. This should give us much more confidence that what we're plotting is correct.

kleinreact:
All except the last enhancement are covered by #415.

PR #415, opened 2023-09-29
ILA plot improvements

kleinreact:
This PR corrects the clock plots of the HITL tests and implements 2, 3, 4 and 5 of  #412. Furthermore, it introduces a more uniform interface for the easy usage of ILA plots among different tests.

The PR does not implement request 1 of  #412, as PPMs are a relative notion and are not recommended to be used by the SI (cf. [The International System of Units, 9th edition](https://www.bipm.org/documents/20126/41483022/SI-Brochure-9.pdf), Section 5.4.7). Instead, we keep measuring the drift in femtoseconds, as it is also done in the simulation plots.

Enhancement 6 of  #412 will be postponed to a later PR, as it requires the introduction of new dedicated HITL tests, which can be considered independently.

---

- [x] Depends on #444 being merged first.


bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc/IlaPlot.hs
```diff
@@ -112,6 +173,67 @@ accwindow _ clk rst ena =
...
+                 )
+
+  store = ( , Just 0) . Just . ( , 0, 0)
+
+  maxOverflows :: Unsigned (SubZero (BitSize (Index m)) n)
+  maxOverflows =
+    let x :: Unsigned (BitSize (Index m))
+        x = bitCoerce (maxBound :: Index m) `shiftR` natToNum @n
+     in case subZeroMin @(BitSize (Index m)) @n of
+          Dict -> checkedTruncateB x
```
martijnbastiaan, line 236: I think this is correct, but I can't quite convince myself due to this function doing multiple things at once. I think this should simplify quite a bit if we do a couple of things:

 * Make `overflow` overflows unrecoverable and reflect that in the type signature. I.e., the output type should be a three-state: `data CounterDiff m = Wait | Ok (Index m) | Error` (name as you see fit). I think this is sensible to do: if an output overflow happens, that's a system design failure in our context, right?
 * Split the function into two: one that assumes that its output is bigger or equal to its input: `(n <= m) => .. -> Signal dom (Maybe (Unsigned n)) -> Signal dom (Maybe (Unsigned m))`. I.e., this function should deal with _input_ overflows but not _output_ overflows. The other function can use that function to implement `.. -> Signal dom (Maybe (Unsigned n)) -> Signal dom (CounterDiff m)`

bittide-instances/src/Bittide/Instances/Tests/FullMeshHwCc/IlaPlot.hs
```diff
@@ -112,6 +173,67 @@ accwindow _ clk rst ena =
...
+overflowResistantDiff ::
+  forall dom n m.
+  (KnownDomain dom, KnownNat n, KnownNat m, 1 <= n, 1 <= m) =>
+  Clock dom ->
+  Reset dom ->
+  Enable dom ->
+  -- | Take the current counter value a the new reference if asserted
+  Signal dom Bool ->
+  -- | The input counter
+  Signal dom (Unsigned n) ->
```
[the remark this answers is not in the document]
kleinreact: This isn't the intention beheind the two signals. The `Unsigned` counter is meant to be fed continiously (so in every cycle), while the pulse may come only when and then.

bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
```diff
@@ -0,0 +1,664 @@
...
+toPulse ::
+  KnownDomain dom =>
+  Clock dom ->
+  Reset dom ->
+  Enable dom ->
+  Signal dom Bool ->
+  Signal dom Bool
+toPulse clk rst ena = mealy clk rst ena transF Nothing
+ where
+  transF s i = (Just i, maybe False (i /=) s)
```
lmbollen, line 422: Could you document what this does and why we need it?

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
```diff
@@ -168,40 +163,39 @@ goFincFdecTests clk rst clkControlled testSelect miso =
...
     )
 fincFdecTests diffClk controlledDiffClock spiIn =
   ((testDone, testSuccess), fIncDec, spiOut)
  where
   clkControlled = ibufds controlledDiffClock
 
-  (clk, clkStable0) = clockWizardDifferential (SSymbol @"pll") diffClk noReset
-  clkStable1 = xpmCdcSingle clk clk clkStable0 -- improvised reset syncer
+  (clk, clkStableRst) = clockWizardDifferential diffClk noReset
+  clkStable1 = unsafeToActiveLow clkStableRst
```
lmbollen, line 197: Are the outputs of `clockWizardDifferential` already synchronized here?

bittide-instances/src/Bittide/Instances/Hitl/FullMeshHwCc.hs
```diff
@@ -230,7 +222,7 @@ fullMeshHwTest ::
   , "ALL_UP" ::: Signal Basic125 Bool
   , "ALL_STABLE"   ::: Signal Basic125 Bool
   )
-fullMeshHwTest refClk sysClk testRst IlaControl{..} rxns rxps miso =
+fullMeshHwTest refClk sysClk IlaControl{syncRst = rst, ..} rxns rxps miso =
```
lmbollen, line 225: Personally I find this a bit confusing. Why not switch `rst` and `syncRst` ?


Then you'll just have:
```hs
fullMeshHwTest refClk sysClk IlaControl{..} rxns rxps miso = _
 where
   rst = syncRst `orReset` otherReset
```

bittide-instances/src/Bittide/Instances/Hitl/FullMeshHwCc.hs
```diff
@@ -343,67 +328,34 @@ fullMeshHwTest refClk sysClk testRst IlaControl{..} rxns rxps miso =
   nFincsSynced = xpmCdcGray txClock sysClk nFincs
   nFdecsSynced = xpmCdcGray txClock sysClk nFdecs
 
-  captureFlag = captureCounter .==. pure (maxBound :: Index (PeriodToCycles Basic125 (Milliseconds 1)))
-  captureCounter = register sysClk sysRst enableGen 0 (satSucc SatWrap <$> captureCounter)
+  captureFlag = (== (maxBound :: Index (PeriodToCycles Basic125 (Milliseconds 1))))
+    <$> (let r = register sysClk syncRst enableGen 0 (satSucc SatWrap <$> r) in r)
```
lmbollen, line 332: again `riseEvery` ?

bittide-instances/src/Bittide/Instances/Hitl/FullMeshHwCc.hs
```diff
@@ -343,67 +328,34 @@ fullMeshHwTest refClk sysClk testRst IlaControl{..} rxns rxps miso =
...
+    (toEnable ((== Just SpeedUp) <$> clockMod))
+    (0 :: Unsigned 32)
+    (satSucc SatBound <$> nFincs)
 
-  isFdec = speedChange2 .==. pure (Just SlowDown)
-  nFdecs = register txClock clockControlReset (toEnable isFdec) (0 :: Unsigned 32) (nFdecs + 1)
+  nFdecs = register txClock clockControlReset
+    (toEnable ((== Just SlowDown) <$> clockMod))
+    (0 :: Unsigned 32)
+    (satSucc SatBound <$> nFdecs)
```
lmbollen, line 342: again `regEn`?

bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
```diff
@@ -0,0 +1,664 @@
...
+  Signal dom SpeedChange
+accWindow _ clk rst ena =
+  flip (mealy clk rst ena) (True, minBound :: Index height)
+    $ \(d,s) -> \case
+         NoChange -> ((d, s), NoChange)
+         x        -> let d' = if x == SpeedUp then d else not d in if
+           |     d' && s == maxBound -> ((not d, minBound), x       )
+           | not d' && s == minBound -> ((not d, minBound), NoChange)
+           |     d'                  -> ((d,     s + 1),    NoChange)
+           | otherwise               -> ((d,     s - 1),    NoChange)
```
lmbollen, line 318: Personally I prefer
```hs
f c r e = mealy c r e go init
 where
  go (d,s) = \case
```
over 
```hs
f c r e = flip (mealy c r e) (init)
 $ \(d,s) -> \case
 ```

**Review by lmbollen — CHANGES_REQUESTED** [review]
Didn't check the 1.9  bump changes.
Overtall LGTM but given that it's a pretty big PR I'd like someone else to take a look too.

File: bittide-instances/src/Bittide/Instances/Hitl/FullMeshSwCc.hs [hunk 9]
```diff
@@ -300,67 +285,34 @@ fullMeshHwTest refClk sysClk testRst IlaControl{..} rxns rxps miso =
...
-  goState _counter False = 0
-
-  goOutput = (== maxBound)
-
 -- | Top entity for this test. See module documentation for more information.
 fullMeshSwCcTest ::
-  "SMA_MGT_REFCLK_C" ::: DiffClock Basic200 ->
-  "SYSCLK_300" ::: DiffClock Basic300 ->
+  "SMA_MGT_REFCLK_C" ::: DiffClock Ext200 ->
+  "SYSCLK_300" ::: DiffClock Ext300 ->
```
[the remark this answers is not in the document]
kleinreact: Yes. They don't necessarily have to be outside the FPGA, but the naming scheme should indicate that they are not generated as part of this top entity. If you have any better name in mind, just let me know.

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
kleinreact: I'd hope so, since the documentation states:

_This function incorporates [resetSynchronizer](https://hackage.haskell.org/package/clash-prelude-1.8.1/docs/Clash-Signal.html#v:resetSynchronizer)s to convert the locked output port into proper [Reset](https://hackage.haskell.org/package/clash-prelude-1.8.1/docs/Clash-Explicit-Signal.html#t:Reset) signals for the output domains which will keep the circuit in reset while the clock is still stabilizing._

bittide-instances/src/Bittide/Instances/Hitl/FullMeshHwCc.hs
kleinreact: It's because I like to pass `syncRst` (with that variable name) to `callistoClockControlWithIla` later, but the resets for `fullMeshHwTest` and `callistoClockControlWithIla` are different although they share the same name. The latter reset is an extension of the prior one, which is why I prefer to use the "overwriting semantics" via

```haskell
syncRst = rst `orReset` (unsafeFromActiveLow (fmap not spiErr))
```
which is intentionally placed right below to be easy to catch.

This is a general issue all over the place, as we extend the reset lines multiple times to incorporate more and more failure indicators. Maybe it's worth to think about a more general solution for this as well.

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
martijnbastiaan: Yes, this has been safe as of the latest Clash release thanks to the work of @DigitalBrains1 !

bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
```diff
@@ -0,0 +1,687 @@
...
+             in xpmCdcGray clk sysClk lts
+
+  -- collect all plot data
+  localData =
+    let rfStageChange CallistoResult{..} = case reframingState of
+          Detect {} -> ToDetect
+          Wait {}   -> ToWait
+          Done {}   -> ToDone
+
+        height = SNat :: SNat AccWindowHeight
```
martijnbastiaan, line 496:
```suggestion
        height = SNat @AccWindowHeight
```

You could also do ^

bittide/src/Data/Constraint/Nat/Extra.hs
```diff
@@ -40,6 +44,20 @@ instance KnownNat n => KnownNat1 $(nameToSymbol ''OneMore) n where
     _ -> SNatKn 1
   {-# INLINE natSing1 #-}
 
+type family SubZero (a :: Nat) (b :: Nat) :: Nat where
```
martijnbastiaan, line 47:
```suggestion
type family SatSubZero (a :: Nat) (b :: Nat) :: Nat where
```

What do you think of this? It follows the term level `satSub SatZero ...`.

bittide/src/Data/Constraint/Nat/Extra.hs
```diff
@@ -40,6 +44,20 @@ instance KnownNat n => KnownNat1 $(nameToSymbol ''OneMore) n where
...
+  SubZero 0 b = 0
```
martijnbastiaan, line 48: Hmmm, this might be subtly different from the term level `subZero a b = if a <= b then 0 else a - b`. As in, if `b` is bottom on the term level the equation will return bottom, but if `b` is bottom on the type level it will give an answer. I'm not 100% sure about this though, maybe type family resolution is strict in its arguments. But.. better safe than sorry, so lets remove this case?

martijnbastiaan (commented):
Overall looks good to me! I've got a few questions and very small suggestions.

.github/hitl/topology.tikz
```diff
@@ -0,0 +1,45 @@
+
+\begin{tikzpicture}[>=latex,line join=bevel,]
```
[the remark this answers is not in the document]
kleinreact: Yes.

The file has been auto generated to match with the style of the simulation plots, but then was adapted manually afterwards to produce a nicer picture. I consider this as a temporary solution until we also enable configurable topologies for the HITL tests.

bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
kleinreact: Good catch!

bittide/src/Data/Constraint/Nat/Extra.hs
kleinreact: You are right, but unfortunately, we cannot drop the first case because of https://hackage.haskell.org/package/ghc-typelits-knownnat-0.7.10/docs/GHC-TypeLits-KnownNat.html (cf. FAQ).

bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
```diff
@@ -0,0 +1,687 @@
...
+data IlaPlotSetup dom =
+  IlaPlotSetup
+    { sysClk :: Clock dom
+      -- ^ The stable system clock.
+    , sysRst :: Reset dom
+      -- ^ The system's reset line.
+    , allUp :: Signal dom Bool
+      -- ^ A boolean signal indicating that all transceivers are up.
+    , startTest :: Signal dom Bool
+      -- ^ The test start signal coming from the HITLT VIO interface.
```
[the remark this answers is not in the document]
martijnbastiaan: Makes sense!

bittide/src/Data/Constraint/Nat/Extra.hs
martijnbastiaan: Oh wow, thanks. Could you link to that FAQ in the source?

martijnbastiaan (approved):

lmbollen (approved):

PR #417, opened 2023-10-06
Fix Simulation Plots

### LINKED ISSUE #418: Rust simulation bug
Rust simulation is not working correctly, which causes the nightlies to fail (cf. [#3096](https://github.com/bittide/bittide-hardware/actions/runs/6460749987)). The error must have been introduced with #403.

I did an additional [sanity check](https://github.com/bittide/bittide-hardware/actions/runs/6465746017) that rust simulation indeed was working before. The additional check was necessary, since #399 modified the bit representation of the [`SpeedChange`](https://github.com/bittide/bittide-hardware/pull/399/files#diff-063d90119a39b6232e751eab2148bb48db4b3af035c195593b60f17739354590L146-R157) in Clash, which required a similar change on the Rust side. This change, however, was not introduced by #399, but has been fixed later by #403 (beside a lot of other changes). The sanity check lives on top of the last commit before #403 and only introduces the `SpeedChange` bit representation fix in Rust (cf. 1e8cfcd).

According to the plots that are generated after #403 has been merged it seems that either the availability mask or the indicator bits are messed up. Note that (almost) complete topologies still seem to work, where "almost" probably is sufficient, because there are still neighbors that drive Callisto to synchronize. This would also explain why especially the `circle` and `star` topology always fail.

My guess is that [this change](https://github.com/bittide/bittide-hardware/pull/403/files#diff-10e4aef4505ec498c775134228401716e1cbefdb9d36ea6e8eafda11232ea30fR318-R335) introduced the error, because it is the only one that is related to this. However, by only looking at the code I cannot see something suspicious yet.

kleinreact:
The elastic buffer data counts of disabled links default to `0`, which is why they automatically turn to be stable and settled after some time. _(Note that there always are elastic buffers and stability checkers for all links of the complete topology in order to support enabling and disabling links at runtime at some point)._

This is why the [old check](https://github.com/bittide/bittide-hardware/blob/a0e4f271efb30500bda95c19fe6d2ee1b40885eb/firmware-support/bittide-sys/src/callisto.rs#L154)
```haskell
stability_checks.iter().all(|x| x.stable())
```
was working (unintentionally) without considering the availability mask. The [new check](https://github.com/bittide/bittide-hardware/blob/3b2ad4858bf9073b11aa05f953f197e198bc56f3/firmware-support/bittide-sys/src/callisto.rs#L137)
```haskell
links_stable.count_ones() == n_buffers
```
however breaks if the stability mask also contains stability indicators for disabled links.

This PR fixes the error by removing the disabled links from the stability mask, i.e., their indicator values are ignored.

----

* Successfully reverts simulation to the old behavior (cf. [#3128](https://github.com/bittide/bittide-hardware/actions/runs/6559343030)).
* Fixes #418.
* Depends on #419 to be fixed first.

### DIFF
--- a/host-tools/callisto-lib/src/lib.rs
+++ b/host-tools/callisto-lib/src/lib.rs
@@ -325,7 +325,7 @@ pub unsafe extern "C" fn __c_callisto_rust(
                 mask |= 1 << i;
             }
         }
-        mask
+        mask & availability_mask
     };
 
     callisto::callisto(


martijnbastiaan:
> Depends on https://github.com/bittide/bittide-hardware/issues/419 to be fixed first.

Done!

martijnbastiaan (approved):
Nice job finding this one!

PR #420, opened 2023-10-13
Add post processing step to HITL infrastructure

hiddemoll:
This PR adds post processing of ILA data to Shake. When a target is assigned a post processing executable, it is included as part of `<target>:test`. It can also executed with `<target>:post-process`. 

----
Specific parts to review:
- The `--hardware-targets` flag now has a default value of `OneAny`. Preferably this flag has _no_ default. 
- Is the usage clear from the READMEs, code documentation and example?


martijnbastiaan (approved):
LGTM in general! Nice work.

bittide-instances/src/Bittide/Instances/Hitl/Post/PostProcess.hs
```diff
@@ -0,0 +1,102 @@
...
+    [testName, toFpgaNum -> fpgaNum, takeBaseName -> ilaName] =
+      splitDirectories relativeCsvPath
+
+-- | Lookup the TestName in the Map, shows a better error message than Map.!?
+-- when the TestName does not exist in the Map.
+getTest :: IlaCsvPaths -> TestName -> Map FpgaNum (Map IlaName FilePath)
+getTest ilaCsvPaths testName =
+  case ilaCsvPaths Map.!? testName of
+    Just x -> x
+    Nothing -> error $ "Could not find key: " <> testName
```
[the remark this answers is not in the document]
martijnbastiaan: If you want to get rid of qualified operators (`m Map.!? k`) you could consider using `lookup` (`lookup k m`).

.github/scripts/cache.py
```diff
@@ -0,0 +1,348 @@
...
+
+BUILD_CACHE_BUST = 2
+BUILD_KEY_PREFIX = f"build-products-{BUILD_CACHE_BUST}-"
+BUILD_CACHE_PATTERNS = (
+    f"{PWD}/.ghc.environment.*",
+    f"{PWD}/_build/",
+    f"{PWD}/clash-vexriscv/clash-vexriscv/build_out_dir/",
+    f"{PWD}/dist-newstyle/",
+    f"{PWD}/firmware-support/target/",
+    f"{PWD}/host-tools/target/",
```
martijnbastiaan, line 72:
```suggestion
    f"{PWD}/_build/",
    f"{PWD}/clash-vexriscv/clash-vexriscv/build_out_dir/",
    f"{PWD}/dist-newstyle/",
```

After: https://github.com/bittide/bittide-hardware/pull/427. The `.ghc.environment.*` removal is unrelated: I just don't think we need it.

nix/bin/cache
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+cd $(git rev-parse --show-toplevel)
```
martijnbastiaan, line 2:
```suggestion
cd "$(git rev-parse --show-toplevel)"
```

Otherwise it will fail if the path contains spaces

PR #429, opened 2023-11-09
Bounded clock control

kleinreact:
This PR adds a cost function to clock control, which imposes bounds on the maximal number of monotone clock modifications. The cost is applied in terms of a temporal delay, which causes `FINC` and `FDEC` requests of the clock controller to take longer to be passed to the clock boards the closer the clock gets adpated towards the physical limits.

In terms of clock control a temporal delay is always applicable, as clock modifcations are only measured indirectly via buffer occupancies used exclusively for feedback control. Hence, as long as the measured feedback of the controller stays the same it will continue to request the same clock modification.

Consequently, imposing a temporal cost can only delays clock operations, but cannot mix things up. We need to consider, however, that the more inert the clock operations are, the more time the elastic buffers will fill or drain. This is why we impose costs in one direction only, e.g., if clock speed mostly increases such that the upper limit gets reached, then further increments come at a higher price, but decrements are still free. The same applies in the other direction, where we assume that clock speeds are between the limits at the center initially. This way, we add an implicit feedback for clock control, which endorses the controller to stay within the limits.

An exponentially increasing cost-function is best suited for this. However, it's implementation requires a counter for implementing the delay, which increases linearly in bit-width according to number of exponential steps. We observed that clock control peaks at about `8k` of `FINC`/`FDEC`s in most of our simulated examples, which is why we cannot apply an exponential cost immediately. Instead, we use a softer scheme to give clock control more freedom as long as it stays within the limits.

In particular, we use a mixed scheme that increases cost logarithmically for the first $2^{14}$ monotone modifications. Hence, clock control is almost not affected by the cost within this region. Beyond, we use a linear increasing scheme for the next $2^{14}$ monotone modifications and then finally switch to an exponential scheme in the end.  The cost function is summarized in the table below, where $w_0 = 0$ and the second column denotes the integral weight of the applied increments and decrements according to their sign function $\sigma$, i.e., for

* $\sigma(x) = -1 $, if clock speed gets decreased,
* $\sigma(x) = 0 $, if clock speed keeps unchanged, and
* $\sigma(x) = 1 $, if clock speed gets increased.

<table>
  <tr>
    <th>$i$</th>
    <th>Weight: $w_{i}=|\sum\limits_{t=t_{init}}^{t_{now}} \sigma(t)|$</th>
    <th>Cost: $c(x)$ for $w_{i-1} \leq x < w_{i}$</th>
    <th>Mode</th>
  </tr>
  <tr><td colspan=4></td></tr>
  <tr>
    <td>$1$</td>
    <td>$2^{12}$</td>
    <td>$0$</td>
    <td rowspan=2>negligible</td>
  </tr>
  <tr>
    <td>$2$</td>
    <td>$2^{14}$</td>
    <td>$1$</td>
  </tr>
  <tr><td colspan=4></td></tr>
  <tr>
    <td>$3$</td>
    <td>$\sum\limits_{n=13}^{14}2^n$</td>
    <td>$2$</td>
    <td rowspan=4>logarithmic</td>
  </tr>
  <tr>
    <td>$4$</td>
    <td>$\sum\limits_{n=12}^{14}2^n$</td>
    <td>$3$</td>
  </tr>
  <tr>
    <td>...</td>
    <td>...</td>
    <td>...</td>
  </tr>
  <tr>
    <td>$16$</td>
    <td>$\sum\limits_{n=0}^{14}2^n$</td>
    <td>$15$</td>
  </tr>
  <tr><td colspan=4></td></tr>
  <tr>
    <td>$17$</td>
    <td>$2^{15}$</td>
    <td>$16$</td>
    <td rowspan=4>linear</td>
  </tr>
  <tr>
    <td>$18$</td>
    <td>$2^{15}+1$</td>
    <td>$17$</td>
  </tr>
  <tr>
    <td>$19$</td>
    <td>$2^{15}+2$</td>
    <td>$18$</td>
  </tr>
  <tr>
    <td>...</td>
    <td>...</td>
    <td>...</td>
  </tr>
</table>

We use 64-bit counter for measuring the delay in terms of clock ticks, so it never should run into it's limits in practice.

Tracking and application of the temporal cost is implemented in hardware, as it easly allows the precise execution of the temporal delay. This approach is also more modular, because it just gets plugged after the clock controller. Moreover, if the process of controlling the clock speed changes in the future, then this module can be easily adapted without additional modifications of clock control.

------

### TODOs:

 - [x] Simulation encounters a memory leak due to the cost based delay. Hopefully an easy one, which only requires some strictness annotations.
 - [x] Timing constraints are violated on some of the builds. The cost function itself imposes certain gate-level depth, but this does not seem to be the limiting factor (as some first tests indicated). Wrongly latching the output of clock control may be the issue as well.
 - [ ] ~~Add SW implementation and disable the HW one.~~
 - [ ] Code cleanup and more code documentation.

### DIFF
--- a/bittide/bittide.cabal
+++ b/bittide/bittide.cabal
@@ -115,6 +115,7 @@ library
     Bittide.ClockControl.Callisto.Util
     Bittide.ClockControl.Foreign.Rust.Callisto
     Bittide.ClockControl.Foreign.Sizes
+    Bittide.ClockControl.QuotaControl
     Bittide.ClockControl.Registers
     Bittide.ClockControl.Si5391A
     Bittide.ClockControl.Si5395J
@@ -190,6 +191,7 @@ test-suite unittests
     Tests.Axi4
     Tests.Calendar
     Tests.ClockControl.Si539xSpi
+    Tests.ClockControl.QuotaControl
     Tests.Counter
     Tests.DoubleBufferedRam
     Tests.ElasticBuffer
--- a/bittide/src/Bittide/ClockControl/Callisto.hs
+++ b/bittide/src/Bittide/ClockControl/Callisto.hs
@@ -19,8 +19,8 @@ import Bittide.ClockControl
 import Bittide.ClockControl.Callisto.Util
 import Bittide.ClockControl.Callisto.Types
 import Bittide.ClockControl.Foreign.Rust.Callisto
+import Bittide.ClockControl.QuotaControl
 import Bittide.ClockControl.StabilityChecker
-import Bittide.Extra.Maybe
 
 import qualified Clash.Cores.Xilinx.Floating as F
 import qualified Clash.Signal.Delayed as D
@@ -70,18 +70,35 @@ callistoClockControl clk rst ena ClockControlConfig{..} mask allDataCounts =
         (clockControl controlConfig mask scs dataCounts state)
         state
 
+      guardedSpeedChangeRequest =
+        mealyB temporallyAlign Nothing
+          ( shouldUpdate
+            -- callisto already introduces a lot of gate-level
+            -- operations, which is why we need to introduce a
+            -- delay here to meet timing constraints. This won't
+            -- cause problems, as the @temporallyAlign@ state
+            -- machine compensates for the temporal cost
+            -- introduced by bounding clock control anyway.
+          , delay Nothing $ quotaControl $ delay NoChange $ fmap _b_k state'
+          )
+
       stabilityCheck = stabilityChecker
         cccStabilityCheckerMargin
         cccStabilityCheckerFramesize
     in
       CallistoResult
-        <$> (orNothing <$> shouldUpdate <*> fmap _b_k state')
+        <$> guardedSpeedChangeRequest
         <*> scs
         <*> allStable
         <*> allSettled
         <*> (rfState <$> state')
 
  where
+  temporallyAlign oldValue (upd, newValue) =
+    if upd
+    then (Nothing, newValue <|> oldValue <|> pure NoChange)
+    else (newValue <|> oldValue, Nothing)
+
   controlConfig =
     ControlConfig
       { reframingEnabled = cccEnableReframing
--- a/bittide/src/Bittide/ClockControl/QuotaControl.hs
+++ b/bittide/src/Bittide/ClockControl/QuotaControl.hs
@@ -1,0 +1,124 @@
+-- SPDX-FileCopyrightText: 2023 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+{-# LANGUAGE RecordWildCards #-}
+
+module Bittide.ClockControl.QuotaControl
+  ( CostDelayBits
+  , QuotaControlState(..)
+  , quotaControl
+  , quotaControlTF
+  ) where
+
+import Clash.Prelude
+import Bittide.ClockControl
+
+-- | The number of bits required for measuring the cost delay.
+type CostDelayBits = 64 :: Nat
+
+-- | The internal state of the controller's state machine.
+data QuotaControlState =
+  QuotaControlState
+    { speedChangeRequest :: SpeedChange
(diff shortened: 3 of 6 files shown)


lmbollen:
Since we are planning to move clock control to software, I think it's more worthwhile to implement this functionality in Rust. Afaik we want to remove the hardware implementation at some point for the gained flexibility.

martijnbastiaan:
@lmbollen Agreed. I think Felix started working on this in hardware, because its currently the only platform that can produce data for the plots.

@kleinreact Can we do anything to port that data production over to software?

kleinreact:
@lmbollen, @martijnbastiaan: I don't see any advantage in moving this to software for several reasons:

1. For clock control, timing is not really critical, but for applying the temporal cost it is. Therefore, measuring the delay in hardware is much easier and more precise. Moreover, it's not even clear how we would implement the delay in software yet. It not only requires a system call to measure time, but also some concurrent interleaving with clock control.

   Note that the exponential rise at the boundary may require waiting for days or weeks, so it must be interleaved with clock control at some point. At the same time, clock modification requests back towards the center must be fast, as the elastic buffers will fill or drain quickly otherwise.

2. Increasing or decreasing clock speed happens in hardware and is already tailored towards the used clock source and interface. Hence, if these things are fixed in hardware anyway, why not adding the boundary checks in front of that interface in hardware as well? The same holds for the clock characteristics (like step size, minimum and maximum speed), which basically determine the stages of the cost function in the end.

3. Implementing the cost function and delay in software will be very inefficient. Note that a 32-bit counter will be too small for implementing the delay, but we are using a 32-bit architecture. Adding much more extra load in software on top of clock control looks kind of dangerous to me. Timing still is a critical factor here.

4. There is no real benefit in having the boundary checks in software in the end, as there is not much choice in implementing it. We count the number of `FINC`s and `FDEC`s, apply the cost function and wait accordingly. That's it. The only thing that may be adapted is the cost function, which can be put into memory for that.

   However, I would keep that as a later TODO, as choosing a good memory representation is non-trivial and it only matters if we are planning for some long term drifting experiments on an ASIC anyway.

5. Not to forget: In hardware we can work with a modular approach just plugging pieces of functionality after each other, which we cannot do that easily in software. Therefore, we should not combine multiple modules of functionality to be executed in software at once, as then things start to mix up. Everything is  quite intertwined already, which is why I always prefer to stick with a modular approach. 

Do you have a concrete example that shows an advantage of doing all that extra work of moving the boundary check to software? Maybe I'm just missing a point.

lmbollen:
@kleinreact @martijnbastiaan 

I think first and foremost, we intend to move clock control to software completely because it offers more flexibility than the hardware approach and Google intends to run it on software in the end AFAIK (please verify @martijnbastiaan).

> For clock control, timing is not really critical, but for applying the temporal cost it is.

How critical? What would be the minimum required granularity of timing control?

> Moreover, it's not even clear how we would implement the delay in software yet. 

My suggestion would be finishing #400 to get a sense of time and in software we keep track of when we last applied `FINC` and `FDEC` seperately and when we can apply again based no the cost.

> Note that the exponential rise at the boundary may require waiting for days or weeks, so it must be interleaved with clock control at some point. At the same time, clock modification requests back towards the center must be fast, as the elastic buffers will fill or drain quickly otherwise.

I don't see how this would not be achievable in software, we don't need to make the processor _wait_ for that delay, it can just conditionally perform operations based on time while keeping the main loop running.

> why not adding the boundary checks in front of that interface in hardware as well?

We want to move the clock control to software because it offers more flexibility than hardware solutions and we don't require the cycle accurate control or the calculation throughput offered by the hardware solution. 

> There is no real benefit in having the boundary checks in software in the end.
Implementation flexibility, centralized clock control logic, resources and meeting timing comes to mind.

It'd probably be a good idea to discuss this online instead of on GH :)

martijnbastiaan:
The conclusion at Big G was that we won't see runaway events. We might want to revisit this in the future if we conclude otherwise.

PR #432, opened 2023-11-15
Use `!cancelled()` instead of `always()`

### LINKED ISSUE #243: `all` succeeds if jobs are skipped (and some succeed)
As title says. This is _usually_ behavior we want, but is undesirable when a user cancels a run. For example:

https://github.com/bittide/bittide-hardware/actions/runs/4017315135/jobs/6901591012

Screenshot 2023-01-26 at 17-59-47 Add Nix to CI · bittide_bittide-hardware@0511832

martijnbastiaan:
As mentioned in the GitHub Actions documentation:

> Avoid using always for any task that could suffer from a critical
> failure, for example: getting sources, otherwise the workflow may
> hang until it times out. If you want to run a job or step
> regardless of its success or failure, use the recommended
> alternative: `if: ${{ !cancelled() }}`

Fixes #243

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -193,7 +193,7 @@ jobs:
             ${{ matrix.target.topology }} ${{ matrix.target.args }}
 
       - name: Generate report
-        if: always()
+        if: ${{ !cancelled() }}
         run: |
           sed -i -e "s/n\([0-9]*\)/\1/g" _build/topology.gv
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
Seen in action: https://github.com/bittide/bittide-hardware/actions/runs/6876061749

hiddemoll (approved):

PR #452, opened 2024-01-22
Move clock control to the stable clock

kleinreact:
This PR moves clock control from the dynamic to the stable system clock for the HITL tests. Furthermore, the following updates have been applied as consequence:

* Simplification of the test setup, as many of the previously used CDC components are not required any more.
* `AccWindowHeight` gets increased from `3` to `5`, as the system clock runs at 300Mhz, thus producing more base noise then the dynamic clock running at 200Mhz. Without this change, the ILA buffers would fill up too fast.
*  An extended variant of `domainDiffCounter` gets introduced, which allows to capture the domain difference in a different domain than the source and destination domain of the diff counter.
* The CLI of the `plot` executable gets simplified due to some strange CI error, complaining that the provided prefix won't match any more. The updated version automatically scans for suitable files, so passing the prefix as an argument turns obsolete. _(Note that this change gets only introduced to get CI working again)_

### DIFF
--- a/bittide/src/Bittide/Counter.hs
+++ b/bittide/src/Bittide/Counter.hs
@@ -5,12 +5,14 @@
 module Bittide.Counter
   ( Active
   , domainDiffCounter
+  , domainDiffCounterExt
   ) where
 
 import Clash.Explicit.Prelude
 
 import Clash.Cores.Xilinx.Xpm (xpmCdcGray)
 import Clash.Sized.Extra (unsignedToSigned, concatUnsigneds)
+import Clash.Explicit.Reset.Extra (Asserted(..), xpmResetSynchronizer)
 
 -- | State of 'domainDiffCounter'
 data DdcState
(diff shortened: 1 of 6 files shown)


lmbollen:
Could you split up the work into separate commits?
It's harder to review a PR that does multiple things, but consists of a single commit.

bittide/src/Bittide/Counter.hs
```diff
@@ -83,6 +85,35 @@ domainDiffCounter clkSrc rstSrc clkDst rstDst =
...
+  , KnownDomain src
+  , KnownDomain dst
+  , HasSynchronousReset ext
+  , HasDefinedInitialValues src
+  , HasDefinedInitialValues dst
+  ) =>
+  Clock ext -> Reset ext ->
+  Clock src -> Clock dst ->
+  Signal ext (Signed 32)
+domainDiffCounterExt clkSys rstSys clkSrc clkDst = truncateB <$>
```
martijnbastiaan, line 102:
```suggestion
domainDiffCounterExt clkExt rstExt clkSrc clkDst = truncateB <$>
```

bittide/src/Bittide/Counter.hs
```diff
@@ -83,6 +85,35 @@ domainDiffCounter clkSrc rstSrc clkDst rstDst =
...
+  ((-) <$> extendedGrayCounter clkSrc <*> extendedGrayCounter clkDst)
```
martijnbastiaan, line 103: Clever!

martijnbastiaan (approved):
Could you put the PR cover letter comments into the commit bodies too?

bittide/src/Bittide/Counter.hs
kleinreact: Thanks.

PR #453, opened 2024-01-24
Fix asymmetric buffer occupancies

### LINKED ISSUE #449: Reframing on HITLT
Reframing currently fails on the HITL tests to successfully recenter the buffer occupancies of the elastic buffers (cf. [#3455](https://github.com/bittide/bittide-hardware/actions/runs/7445553156)).

image

It has to be investigated whether this is a theory, implementation, or methodology issue. 

_Note that  waiting times are sufficiently large. All of the nodes store the right correction values for reframing (captured at the beginning of the red markings indicating the waiting times)._

kleinreact:
#415 changes the ILA dump structure for the elastic buffer sizes to only report differences since the last and the current capture in order to compress the amount of data being reported. 

It may be that a difference grows too fast and exceeds the compressed data-size-boundary between two captures, which is why a fail-safe procedure was implemented creating some additional capture before that will happen (resetting the difference counter). However, this protection mechanism was not setup correctly. As a consequence, the compressor sometimes ends up in a situation, where it starts measuring a new difference although the previous difference never got reported. Hence, it's accumulated value got lost, which caused the plotted graphs to miss some amount of the buffer changes resulting in the asymmetric plots.

This PR fixes this issue, via correctly handling the overflow-protection trigger and the difference output creation of the data compressor.

- [x] fixes  #449

### DIFF
--- a/bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
@@ -452,8 +452,6 @@ callistoClockControlWithIla dynClk clk rst ccc IlaControl{..} mask ebs =
         height = SNat @AccWindowHeight
         idcs = unbundle (stability <$> result)
 
-        ebsC = ebsDiffCompress scheduledCapture <$> ebs
-
         -- get the points in time where the monitored values change
         stableUpdates  = changepoints clk rst enableGen <$> (fmap stable <$> idcs)
         settledUpdates = changepoints clk rst enableGen <$> (fmap settled <$> idcs)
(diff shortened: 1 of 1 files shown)


bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
```diff
@@ -443,51 +443,48 @@ callistoClockControlWithIla dynClk clk rst ccc IlaControl{..} mask ebs =
...
-              truncDiff = truncateB @_
-                @CompressedBufferSize
-                @(m - CompressedBufferSize)
-                diff
-           in (, truncDiff)
-            $ bool lastDataCount curDataCount
-            $ trigger || abs diff > half
-     in mealyB clk rst enableGen transF (0 :: DataCount m)
+  (ebDataChange, ebsC) = second unbundle
+    $ mealyB clk rst enableGen transF (repeat 0) (scheduledCapture, bundle ebs)
```
martijnbastiaan, line 474:
```suggestion
  (ebDataChange, unbundle -> ebsC) =
    mealyB clk rst enableGen transF (repeat 0) (scheduledCapture, bundle ebs)
```

A view pattern is more obvious than a HO combinator IMO. What do you think?

martijnbastiaan (changes requested):
I'm not against refactors, but could you please split out the refactor and actual change into two commits? It's kinda hard to see what has _actually_ changed now.

----------

Could you change the main commit message to:

```
Trigger data capture on FINC/FDEC accumulator overflow

<PR cover letter>

Fixes #449
```

I.e.

```
Technical thing that has changed

Context, references, and documentation

Fixes #1234

Co-authored-by: blablabla
```

Note that `Trigger data capture on FINC/FDEC accumulator overflow` is how I understand the current patch, but as said I found it kinda hard to see what actually changed :upside_down_face:

martijnbastiaan:
Nice detective work in any case!

bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs
kleinreact: It depends. The higher order use of `second` here results in a linear control flow, while view patterns introduce a branching flow. I'll personally prefer to use linear flow, as it is easier to read, copy paste (while coding), and extend with type annotations if necessary. I decided to use `second` here, as it is already imported anyway, so it comes at no cost of an extra import.

martijnbastiaan (approved):

martijnbastiaan:
Please don't forget the `fixes #...`. It gives a nice cross-reference in the git history. Plus, it automatically closes issues.

PR #457, opened 2024-01-25
HITL infra for parameter configuration via YAML

lmbollen:
The PR allows us to write YAML files that describe tests.
Whenever we want to use execute multiple HITL tests on a single DUT, we always want to change some parameters per run.

Now we have the infra where we hard-code extra starting probes and based on that choose different parameters, but a better way would be to set the parameters with a configuration file free from the testbench.

This way we can easily run N tests on a design without changing the testbench.
In the future this infra enables us to simply generate these yaml files for extra flexibility.

### DIFF
--- a/bittide-instances/data/test_configs/default.yml
+++ b/bittide-instances/data/test_configs/default.yml
@@ -1,0 +1,10 @@
+# SPDX-FileCopyrightText: 2024 Google LLC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Tcl can not handle empty lists pr dicts, so we supply a null values instead.
+default:
+  probes: null
+
+tests:
+  single_test: null
--- a/bittide-instances/src/Bittide/Instances/Hitl/Tcl/ExtraProbes.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/Tcl/ExtraProbes.hs
@@ -1,0 +1,56 @@
+-- SPDX-FileCopyrightText: 2024 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+module Bittide.Instances.Hitl.Tcl.ExtraProbes where
+
+import Clash.Prelude
+
+import Clash.Cores.Xilinx.Dna
+import Bittide.Instances.Domains
+import Clash.Annotations.TH ( makeTopEntity )
+import Clash.Cores.Xilinx.Extra
+import Clash.Cores.Xilinx.VIO
+import Data.Maybe
+
+-- | A circuit that verifies the correct behavior of the TCL infrastructure for
+-- setting extra probes in Hitl tests.
+extraProbesTest ::
+  "CLK_125MHZ" ::: DiffClock Ext125 ->
+  "success" ::: Signal Ext125 Bool
+extraProbesTest diffClk = testSuccess
+ where
+  clk = ibufds diffClk
+
+  testSuccess = testResult <$> testState <*> extraProbe <*> fpgaId
+  testDone = testStart .&&. fmap isJust fpgaId
+  rst = unsafeFromActiveLow testStart
+  fpgaId = withClockResetEnable clk rst enableGen deviceDna defaultSimDNA
+  (testStart, testState, extraProbe) =
+    unbundle $
+    setName @"vioHitlt" $
+    vioProbe
+      ("probe_test_done" :> "probe_test_success" :> "fpgaId" :> Nil)
+      ("probe_test_start" :> "testState" :> "extraProbe" :> Nil)
+      (False, SetDefaultProbes, maxBound)
+      clk
+      testDone
+      testSuccess
+      fpgaId
+{-# NOINLINE extraProbesTest #-}
+
+-- | Produce the test result based on the test state and the extra probe value.
+-- These values should correspond to the yaml configuration.
+testResult :: TestState -> BitVector 96 -> Maybe (BitVector 96) -> Bool
+testResult s extraProbe fpgaId = case (s, extraProbe) of
+  (SetDefaultProbes, 0) -> True
+  (SetTestProbes, 0xDEADABBA) -> True
+  (SetFpgaSpecificProbes, _) -> extraProbe == fromMaybe defaultSimDNA fpgaId
+  _ -> False
+
+data TestState
+  = SetDefaultProbes -- Check if the default probes from the yaml file are set
+  | SetTestProbes -- Check if the test specific probe values are set
+  | SetFpgaSpecificProbes -- Check if the DNA device identifier is set
+  deriving (Generic, NFDataX)
+
+makeTopEntity 'extraProbesTest
--- a/bittide-shake/data/tcl/HardwareTest.tcl
+++ b/bittide-shake/data/tcl/HardwareTest.tcl
@@ -2,6 +2,45 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+# Tools to run hardware-in-the-loop (HITL) tests. Most users should consider
+# `runTestGroup` as the main entry point. This function runs a group of tests
+# according to a test configuration file. The test configuration file is defined
+# in YAML, and looks like the following:
+#
+# ```yaml
+# defaults:
+#   probes:
+#     $probe1: 0
+#     $probe2: 0
+#
+# tests:
+#   $test_name:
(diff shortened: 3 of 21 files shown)


kleinreact (commented):
Great work. I only have some minor comments suggesting some improvements. 

Other than that, CI still fails because you changed the `probe_test_start` probe name to `default_name` causing a different file path for the ILA dumps. Moreover, don't forget to rebase on `staging` as report generation still may fail do to name changes otherwise.

bittide-instances/data/test_configs/default.yml
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: 2022-2023 Google LLC
+#
+# SPDX-License-Identifier: Apache-2.0
+.default:
+  dummy: "Tcl can not handle empty dictionaries, so we need a dummy entry here"
+
+tests:
+  # tests contains all the tests that must be run with their corresponding configuration
+  default_name:
+    dummy: "Tcl can not handle empty dictionaries, so we need a dummy entry here"
```
DigitalBrains1, line 10:
```suggestion
.default: null  # Tcl does not support empty scalars, explicit null is needed

tests:
  # tests contains all the tests that must be run with their corresponding configuration
  default_name: null  # Tcl does not support empty scalars, explicit null is needed
```

hiddemoll (commented):
Nice work. I only have a few minor syntax requests

bittide-shake/data/tcl/HardwareTest.tcl
```diff
@@ -474,22 +525,73 @@ proc print_test_results {done success start_time end_time} {
...
+    set test_names [dict keys $tests]
+    set test_names [lsearch -all -inline -not -exact $test_names .default]
+    return $test_names
+}
+
+proc get_test_config_dict {fp} {
+    puts -nonewline "Loading Test config file $fp: "
+    set fd [open $fp r]
+    set yaml_dict [::yaml::yaml2dict [read $fd]]
+    close $fd
```
DigitalBrains1, line 543:
```suggestion
    set yaml_dict [::yaml::yaml2dict -file $fp]
```

bittide-instances/data/test_configs/default.yml
DigitalBrains1: By the way, the Tcl `yaml` package version 0.4.1 has no problem with empty scalars, unfortunately, Vivado includes version 0.3.3.

bittide-instances/src/Bittide/Instances/Hitl/Tcl/ExtraProbes.hs
```diff
@@ -0,0 +1,55 @@
...
+import Clash.Cores.Xilinx.Dna
+import Bittide.Instances.Domains
+import Clash.Annotations.TH ( makeTopEntity )
+import Clash.Cores.Xilinx.Extra
+import Clash.Cores.Xilinx.VIO
+import Data.Maybe
+
+-- | A circuit that verifies the correct behavior of the TCL infrastructure for
+-- setting extra probes in Hitl tests.
+extraProbesTest ::
```
martijnbastiaan, line 17: Don't forget to add a `NOINLINE`. GHC might give it a different name if you don't.

bittide-shake/data/tcl/HardwareTest.tcl
```diff
@@ -474,22 +520,79 @@ proc print_test_results {done success start_time end_time} {
...
+    set changed_vios []
+    foreach probe_dict $probe_dicts {
+        dict for {vio_name vio_value} $probe_dict {
+            set probe [get_hw_probes ${vio_prefix}/${vio_name}]
+            if {[lsearch -exact $changed_vios $probe] }  {
+                lappend changed_vios $probe
+            }
+            set bit_width [get_property width $probe]
+            set hex_width [expr {(3 + $bit_width)/4}]
+            set vio_value [format "%0${hex_width}llX" [expr $vio_value]]
```
martijnbastiaan, line 577:
```suggestion
            set vio_value [format "%0${hex_width}llX" [parse_integer $vio_value]]
```

Could you make sure that value that's being passed in corresponds to one of these regexes:

```
0x[0-9A-F]+
0b[0-1]+
0o[0-7]+
[0-9]+
```

`expr` can do arbitrary function calls.

martijnbastiaan: Some ChatGPT magic:

```tcl
proc matchNumericFormat {str} {
    # Combine the patterns with logical OR
    set pattern {^(0x[0-9A-Fa-f]+|0b[0-1]+|0o[0-7]+|[0-9]+)$}

    # Check if the string matches any of the patterns
    if {[regexp $pattern $str]} {
        return 1  ;# True, the string matches one of the patterns
    } else {
        return 0  ;# False, the string does not match any of the patterns
    }
}

# Example usage
set strings {0x1A3F 0b101001 0o755 12345 notANumber}

foreach str $strings {
    if {[matchNumericFormat $str]} {
        puts "$str matches one of the patterns."
    } else {
        puts "$str does not match any of the patterns."
    }
}
```

martijnbastiaan: And yes you can be sure to get review comments if you copy paste that at verbatim :-)

DigitalBrains1: I think you just want to check with `string is entier $vio_value`.

DigitalBrains1: Boo, Vivado's Tcl is too old for that. use `wideinteger`.
```tcl
if {![string is wideinteger $vio_value]} {
  puts "\$vio_value not an integer: $vio_value"
  exit 1
}
```
(I don't know why you don't use `error`? Is the result not pleasing to the eye? But I just followed what was there already in the file, which is `puts` / `exit 1`)

DigitalBrains1: Finally, this seems to work just fine
```tcl
format %0${hex_width}llX $vio_value
```
(I also removed superfluous quoting)

martijnbastiaan: > I don't know why you don't use error? 

I just copied straight from ChatGPT. Like I said

> And yes you can be sure to get review comments if you copy paste that at verbatim :-)

:-)

martijnbastiaan: > Finally, this seems to work just fine

Ah of course. Everything is a string in TCL after all!

DigitalBrains1: >> I don't know why you don't use `error`?
>
> I just copied straight from ChatGPT. Like I said

Wrong "you" :-). This script, `bittide-shake/data/tcl/HardwareTest.tcl`, uses `puts` and `exit 1` whenever it handles an error condition, so I followed suit in my suggestion, but asked why it was done in this way. This suddenly becomes more relevant, because if you just do `format %0${hex_width}llX $vio_value` in the code and `$vio_value` is not a number, it'll behave pretty much like `error`, and not like `puts` / `exit 1`. So if it was a conscious decision to use `puts` and `exit 1` for handling all error conditions, this will have to be handled explicitly. If normal Tcl error handling is acceptable, you can just not check `$vioValue`, invoke `format` and let that throw an error.

DigitalBrains1: At a glance, the main difference between `puts` / `exit 1` and Tcl error handling is that the latter outputs a stack trace rather than just the message. They both exit with a return code 1 in Vivado, and both output their messsages on `stdout`.

martijnbastiaan: Then I guess `error` is preferable!

martijnbastiaan (approved):
GLTM

@DigitalBrains1 indicated he wanted to look at the TCL more closely. Let's talk to him in the office today.

kleinreact (changes requested):
It all looks good to me from a technical point of view, but due to the amount of changes introduced I am still missing some documentation giving a quick overview of on the new features and essential changes. 

Could you please add such an overview in a Readme or somewhere else in the sources making it easier to the yet unknowledged and future us to grasp what's going on?

kleinreact (approved):
LGTM. Thanks for the update.

PR #458, opened 2024-01-26
Switch SYNC_IN/OUT cable setup

kleinreact:
The new setup utilizes the USER SMA GPIO connectors instead of the pmods + level shifters.


martijnbastiaan (approved):

martijnbastiaan:
@kleinreact: I've added @DigitalBrains1 to the Google group so you should be able to add him as co-author now. Not sure what I did to warrant co-authorship though :eyes:.

martijnbastiaan:
Also, probably add to your commit message that we switched this out due to glitches on the PMOD wires. With the SMA connections we've got shielded and properly terminated connections.

PR #460, opened 2024-01-31
HITLT YAML Configuration Interface

kleinreact:
This PR adds a dedicated  `vioHitlt` test interface to Clash and the `yamlpack` tool to convert between human readable YAML configuration files and the corresponding machine readable ones for passing data to the VIO test interfaces.

For detailed  information on the update consider:
  * the haddock documentation of  [Bittide.Instances.Hitl](https://github.com/bittide/bittide-hardware/blob/hitl-yaml-cfgs/bittide-instances/src/Bittide/Instances/Hitl.hs)
  * the updated [HITLT README](https://github.com/bittide/bittide-hardware/blob/hitl-yaml-cfgs/bittide-instances/src/Bittide/Instances/Hitl/README.md)
  * and the new [HITLT YAML Configurations README](https://github.com/bittide/bittide-hardware/blob/hitl-yaml-cfgs/bittide-instances/data/hitltconfigs/README.md)

---
### TODO:
  - [ ] Adapt `HardwareTest.tcl` to the packed YAML configuration format.
  - [ ] Add some more advanced parameter set to the `FullMeshHwCC` test.


martijnbastiaan:
This might come off as a bit harsh, but I don't think converting one YAML structure to another warrants the introduction of 10+ custom type families and type level witnesses. I'm having great trouble properly understanding what's happening, so I worry about:

1. How less experienced Haskell/Clash developers are going to understand this (which is apparently me)
2. How we're going to maintain this

I'd urge you to find a better balance between (type) safety and implementation simplicity. Alternatively, I'd like to convinced of the necessity of the complexity.

--------------------

So for the more constructive part of the review: why don't we start defining tests straight from Haskell? It seems to me that a structure like this:

```yaml
testname1:
  default: <cfg>  # overwrites the default configuration value for this test case
  disable: true # disables the test case (default: false)
  fpga:
    0:              # first fpga according to Bittide.Instances.Hitl.Setup
      disable: true # disables the test execution for this fpga (default: false)
      data: <cfg>     # sets a new configuration value for this fpga only
    1:              # second fpga according to Bittide.Instances.Hitl.Setup
      ...
testname2:
  ...
```

Can be expressed in Haskell reasonably easily. Something like:

```haskell
data HitlTest a = HitlTest
  { default :: a
  , disable :: Bool -- As a side node, avoid negative names, call this enable: https://davej.io/2018/05/negative-naming.html
  , fpga :: FpgasConfig a
  }

type HitlTests a = Map String (HitlTest a)
```

Which could then implement a `ToJSON`:

```haskell
instance BitPack a => ToJSON (HitlTests a) where
  ..
```

to produce:

```yaml
testname1:
  <id0>: <binary data>
  <id1>: <binary data>
  ...
testname2:
  ...
```

And so now the only thing you need to worry about is that the config uses the same `a` as the test. This way the implementation will stay much simpler, while offering almost all of the type guarantees.

martijnbastiaan:
One other thing is that we're now giving up the ability to have multiple input probes. This doesn't matter that much for the Haskell/Clash side of things, but in practice we're going to be bound to view this through Vivado's / TCL's lens, at which point the probe values will basically look like binary garbage.

martijnbastiaan:
I guess I should clarify after having a short chat with @lmbollen. One of the ideas of this PR is that the starting point would be non-packed YAML files, the idea being that these would be easy to generate. My review was based on: why not start straight from Haskell? That way the implementation LOC would shrink by an order of magnitude, as we would get a parser (and all of its correct-by-construction benefits) for free. Plus, test cases would still be easy to generate (because Haskell).

So, my proposal would a general module:

```haskell
-- | Easy to write test config
data HitlTest a = .......

-- | Collection of tests
type HitlTests a = Map String (HitlTest a)

-- | 'HitlTests', but the way TCL likes it. Should already have a ToJSON.
type PackedHitlTests = Map String (Map FpgaId Integer)

-- | VIO with all the probes the TCL needs. Outputs 'Just' if a test has started, otherwise 
-- 'Nothing'.
hitlVio :: [..] -> Signal dom (Done, Success) -> Signal dom (Maybe a)

-- | Like 'hitlVio', but allows custom probes
hitlVioWithExtraProbes :: MoreComplexTypeSignature

-- Perhaps some convenience functions
testAll :: a -> HitlTests a
testOnly :: FpgaId -> a -> HitlTests a

toPacked :: BitPack a => HitlTests a -> PackedHitlTests
toPacked = ...

-- | Provides a way to call from command line. Is primarily used to generate TCL-friendly
-- YAML. Could in the future be used to provide a mapping from a bunch of bits recovered
-- through the VIO, to something human readable.
defaultMain :: BitPack a => HitlTests a -> IO ()
defaultMain = ...
```

Per test module:

```haskell
testConfig = [
  ("test_1", HitlTest .....)
  ("test_2", HitlTest .....)
]

[..]

-- | Generate TCL test config
main :: IO ()
main = defaultMain testConfig
```

I'm not hung up on these _exact_ type signatures, but I think these definitions are easy to grasp for beginners, easy to extend, and cause nice error message if you mess up. The only way this API could be abused is if `vioHitl` gets given a different config type than `defaultMain`, but :shrug: .

Anyway, this is all to guide the discussion. We can talk about it IRL.

martijnbastiaan:
After more discussion with @lmbollen, the packed HITL:

```yaml
testname1:
  - id: <id0>
    probes_defaults:
      <probe_1>: <binary data>
      <probe_2>: <binary data>
    probes:
      <probe_1>: <binary data>
      <probe_2>: <binary data>
  - id: <id1>
    probes:
      <probe_1>: <binary data>
      <probe_2>: <binary data>
  ...
testname2:
  ...
```

This:

 * Assigns an explicit order to the FPGA IDs. This is what we want in the future, when we make the TCL index agnostic. (I.e., no mapping "FPGA in rack" => "FPGA ID" in TCL.)
 * Retains the ability to set multiple probes, though Haskell won't use it for now.
 * Retains the ability to be hand-written due to `probes_defaults` - without it probe files would be pretty cumbersome to write.

Hardcoded probes (`start`, `success`, `done`) will remain a thing.

kleinreact:
Having a configuration format that works independently of particular Haskell types, but still is compatible with this approach, sound like a great improvement to me. Then we still are flexibility in the choosing the best solution depending on the actual application and are not particularly limited by the interface.

martijnbastiaan:
Superseded by https://github.com/bittide/bittide-hardware/pull/462

PR #461, opened 2024-02-02
Add `license-year` check to CI

lmbollen:
A check that verifies that the copyright notice in all files touched by a pull request have been updated to the year the commit was made.

### DIFF
--- a/.github/scripts/check_copyright_years.py
+++ b/.github/scripts/check_copyright_years.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python3
+"""
+Make sure all files have an up-to-date copyright notice.
+"""
+
+# SPDX-FileCopyrightText: 2024 Google LLC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import re
+import sys
+import git
+
+from typing import Set, Tuple
+
+SPDX_MAGIC = "SPDX-FileCopyrightText:"
+
+YEAR_RE = re.compile(r"\d{4}")
+
+IGNORE_STARTSWITH = (
+    "clash-vexriscv",
+)
+
+def iter_change_types(diff_index, change_types):
+    for change_type in change_types:
+        yield from diff_index.iter_change_type(change_type)
+
+def get_pr_commits(repo, common_ancestor):
+    """
+    Gets all commits from the current one to (but not including) a common ancestor
+    given as an argument.
+    """
+    for commit in repo.iter_commits():
+        if commit.binsha == common_ancestor.binsha:
+            break
+        yield commit
+
+def get_relevant_files_from_diff_index(diff_index):
+    """
+    Get all filepaths that are added, modified, or renamed in a DiffIndex.
+    """
+    for diff in iter_change_types(diff_index, ("A", "M", "R")):
+        yield diff.b_path
+
+def get_copyright_years(repo, commit, path):
+    content = repo.git.show(f"{commit.hexsha}:{path}")
+    lines = content.split("\n")
+
+    for lineno, line in enumerate(lines, start=1):
+        if SPDX_MAGIC in line:
+            matches = YEAR_RE.findall(line)
+            if not matches:
+                # XXX: We don't error on "poorly formatted" SPDX headers, as our
+                #      current check also triggers on things like our SPDX_MAGIC
+                #      definition.
+                continue
+            yield int(matches[-1]), lineno
+
+def is_valid_commit_file(repo, commit, path) -> bool:
+    """
+    Checks whether a specific file in a specific commit has up-to-date copyright
+    notices.
+    """
+    commit_year = commit.authored_datetime.year
+    max_year, line_no = max(get_copyright_years(repo, commit, path))
+
+    if max_year != commit_year:
+        print(f"{commit.hexsha} {path}:{line_no} Unexpected copyright year. Expected {commit_year}, got {max_year}.", file=sys.stderr)
+
+    return True
+
+
+def is_valid_commit(repo, commit, seen_files) -> Tuple[bool, Set[str]]:
+    """
+    Check whether all files touched in given commit have up-to-date copyright
+    notices. Ignores `seen_files` to make sure only the oldest commits introducing
+    an issue get listed.
+
+    Returns a tuple of (everything_ok, checked_files)
+    """
+    if not commit.parents:
+        print("Ignoring root commit..", file=sys.stderr)
+        return True, set()
+
+    parent = commit.parents[0]
+    files = set(get_relevant_files_from_diff_index(parent.diff(commit)))
+    files -= seen_files
+    files = {f for f in files if not f.startswith(IGNORE_STARTSWITH)}
+
+    all_valid = True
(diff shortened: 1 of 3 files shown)


.github/scripts/license_years.py
```diff
@@ -0,0 +1,59 @@
...
+commits = []
+for commit in all_commits:
+  if commit.hexsha == common_ancestor_hash:
+    break
+  commits.append(commit.hexsha)
+
+# Check if the year in the license header is correct for each file in each commit
+all_valid = True
+for commit in commits:
+  obj = repo.commit(commit)
```
martijnbastiaan, line 31: So first you convert a commit object into a hexdigest, and now you're converting it to a commit object again? Just use commit objects throughout.

.github/scripts/license_years.py
```diff
@@ -0,0 +1,59 @@
...
+  commit_year = obj.committed_datetime.year
+  if len(obj.parents) == 0:
+    # Must be an initial commit
+    continue
+  parent = obj.parents[0]
+  # get the diff of the commit
+  diffs = parent.diff(obj)
+  for diff in diffs:
+    # Only check added or modified files
+    if diff.change_type in ['A', 'M']:
```
martijnbastiaan, line 41: Factor this whole thing out into a function `get_changed_files(commit)`

.github/scripts/license_years.py
```diff
@@ -0,0 +1,59 @@
...
+      fp = diff.b_path
```
martijnbastiaan, line 42: Don't you want `b_rawpath` here? Does that point directly to the file in `.git`?

.github/scripts/license_years.py
```diff
@@ -0,0 +1,59 @@
...
+      content = git.show(f"{commit}:{fp}").split('\n')
+      for line in content:
+        # Check if the line contains "SPDX-FileCopyrightText:"
```
martijnbastiaan, line 45:
```suggestion
```

Does what it says on the tin. Comments should either:

* Clarify _why_ things happen
* Or in case of complicated code _what_ happens (but ideally it should be factored out in a function)

In this case, just remove it.

.github/scripts/license_years.py
```diff
@@ -0,0 +1,59 @@
...
+        if "SPDX-FileCopyrightText" in line:
+          # Extract all the years from the line and get the last one
+          last_year = int(re.findall(r"\d{4}", line)[-1])
+          if last_year != commit_year:
+            all_valid = False
+            print("File", fp, "in commit", commit, "has incorrect year. Expected", commit_year, "but found", last_year)
```
martijnbastiaan, line 51: Use f-strings

```suggestion
            print(f"File {fp} in commit {commit} has incorrect year. Expected {commit_year} but found {last_year}")
```

.github/scripts/license_years.py
```diff
@@ -0,0 +1,59 @@
...
+  for diff in diffs:
+    # Only check added or modified files
+    if diff.change_type in ['A', 'M']:
+      fp = diff.b_path
+      content = git.show(f"{commit}:{fp}").split('\n')
+      for line in content:
+        # Check if the line contains "SPDX-FileCopyrightText:"
+        if "SPDX-FileCopyrightText" in line:
+          # Extract all the years from the line and get the last one
+          last_year = int(re.findall(r"\d{4}", line)[-1])
```
martijnbastiaan, line 48: This will crash if the returned list is empty

martijnbastiaan: Factor out the regex. It's good practice to compile before. So define a global:

```python
YEAR_RE = re.compile(r"\d{4}")
```

.github/scripts/license_years.py
```diff
@@ -0,0 +1,59 @@
...
+          if last_year != commit_year:
+            all_valid = False
+            print("File", fp, "in commit", commit, "has incorrect year. Expected", commit_year, "but found", last_year)
+          break
+
+if not all_valid:
+  print("Incorrect year found in some files. Please update the year in the license header")
+  exit(1)
```
martijnbastiaan, line 56:
```suggestion
  print("Incorrect year found in some files. Please update the year in the license header", file=sys.stderr)
  exit(1)
```

Errors and logging should go to stderr. Alternatively:

```python3
exit("my error")
```

Also check your other `print`s.

.github/scripts/license_years.py
martijnbastiaan: Also, the source code of [`iter_change_type`](https://github.com/gitpython-developers/GitPython/blob/f4ce70974d8ca0e4d2699b4d5f1016880815ae39/git/diff.py#L222) suggest that you need more than just look at `change_type`. To be safe, change this to:

```python
def iter_change_types(diff_index, change_types):
    for change_type in change_types:
        yield from diff_index.iter_change_type(change_type)
```

martijnbastiaan (changes requested):
@hiddemoll asked me to look at this.

General comments:

 * Use PEP8, which mostly means 4-spaces for indents. (I messed that up in another Python script too, maybe we need a linter.)
 * Put logic in a `main()`. At the bottom of your script use `if __name__ == '__main__':`. That makes your script reusable. See [this SO answer](https://stackoverflow.com/questions/419163/what-does-if-name-main-do)

.github/scripts/license_years.py
martijnbastiaan: Also, print `{fp}:{lineno}` so VSCode users can CTRL+click.

.github/scripts/check_copyright_years.py
```diff
@@ -0,0 +1,129 @@
...
+        print(f"{commit.hexsha} {path}:{line_no} Unexpected copyright year. Expected {commit_year}, got {max_year}.", file=sys.stderr)
+
+    return True
+
+
+def is_valid_commit(repo, commit, seen_files) -> Tuple[bool, Set[str]]:
+    """
+    Check whether all files touched in given commit have up-to-date copyright
+    notices. Ignores `seen_files` to make sure only the oldest commits introducing
+    and issue get listed.
```
hiddemoll, line 77:
```suggestion
    an issue get listed.
```

.github/scripts/check_copyright_years.py
```diff
@@ -0,0 +1,129 @@
...
+        exit(f"No commits on top of base branch '{merge_base_branch}'. Empty PR?")
+
+    # Get the commits that are in the current branch but not in the merge target
+    pr_commits = get_pr_commits(repo, common_ancestor)
+
+    # Check if the year in the license header is correct for each file in each commit
+    all_valid = True
+    seen = set()
+    for commit in pr_commits:
+        (commit_valid, files_in_commit) = is_valid_commit(repo, commit, seen)
```
hiddemoll, line 115: You can also just ignore this one, as both with and without brackets is valid. You unpack _without_ brackets in line 65, so removing the brackets here is more uniform.
```suggestion
        commit_valid, files_in_commit = is_valid_commit(repo, commit, seen)
```

hiddemoll (approved):
Nice work!

PR #462, opened 2024-02-06
Add `hitlVio` and make tests generate their own test data

martijnbastiaan:
Blocked on:

 - [x] https://github.com/bittide/bittide-hardware/pull/457

TODO:

 - [x] Documentation

Future work (other PRs):

 - [ ] Put test definitions in ONE place. It's currently cumbersome to add a new test, you need to modify five (!) separate places (`bin/Hitl.hs`, `Shake.hs`, `all.json`, `staging.json`, the test itself).
 - [ ] Add options to Shake to select specific tests / custom test definition files
 - [ ] Restore ability to mention "any FPGA".
 - [ ] Listen to FPGAs defined in YAMLs, instead of taking them from CI JSONs


bittide/src/Clash/Hitl.hs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+  goN probes = PackedTest
+    { targets=PackedList
+      [ PackedTarget
+          { target=ByIndex (toInteger id_)
+          , probes=PackedProbes [("probe_test_data", BitVector.unsafeToNatural (pack dat))]
+          }
+      | (id_, dat) <- probes]
+    }
```
lmbollen, line 195: Why not:
```hs
-- | Convert an \"unpacked\" "HitlTests" to a packed version. The packed version
-- is convertable to JSON, which in turn can be interpreted by the @HardwareTest.tcl@.
toPacked :: forall a. BitPack a => HitlTests a -> PackedTests
toPacked hitlTests = PackedTests {defaults, tests}
 where
  bs = natToInteger @(BitSize a)
  defaults = PackedProbes (if bs == 0 then [] else [("probe_test_data", 0)])
  tests=fromList [(name, go probes) | (name, probes) <- toList hitlTests]

  go probes = PackedTest {targets=PackedList targetList}
   where
    targetList
      | bs == 0 =
        [PackedTarget
          { target=ByIndex (toInteger id_)
          , probes=PackedProbes []
          } | (id_, _) <- probes]

      | otherwise =
        [PackedTarget
          { target=ByIndex (toInteger id_)
          , probes=PackedProbes [("probe_test_data", BitVector.unsafeToNatural (pack dat))]
          } | (id_, dat) <- probes
        ]
    }

File: bittide/src/Clash/Hitl.hs [hunk 2]
```diff
@@ -0,0 +1,289 @@
...
+
+hitlVio dflt clk done success =
+  P.mux start (Just <$> dat) (pure Nothing)
+ where
+  (P.unbundle -> (start, dat)) =
+    P.setName @"vioHitlt" $
+      vioProbe
+        ("probe_test_done" :> "probe_test_success" :> Nil)
+        ("probe_test_start" :> "probe_test_data" :> Nil)
+        (False, dflt) clk done success
```
lmbollen, line 272: Please differentiate between `BitSize a ~ 0` and ` 1 <= BitSize a` in the same way as you eventually do in `toPacked`.

bittide/src/Clash/Hitl.hs
```diff
@@ -0,0 +1,289 @@
...
+
+-- | Instantiate this VIO in a design you'd like to test with hardware in the
+-- loop. Its output is set to 'True' if a test is not running, and will be
+-- set to 'False' if it is.
+hitlVio1 ::
```
lmbollen, line 277: I dislike `name<n>` functions because you have to look up how they differentiate.
how about using `hitlVioWithConfig` and `hitlVio` or `hitlVioBool`? Any other choices are fine too imo.

bittide-instances/bin/Hitl.hs
```diff
@@ -0,0 +1,139 @@
...
+import System.FilePath ((</>))
+
+import qualified Bittide.Instances.Hitl.BoardTest as BoardTest
+import qualified Bittide.Instances.Hitl.FincFdec as FincFdec
+import qualified Bittide.Instances.Hitl.FullMeshHwCc as FullMeshHwCc
+import qualified Bittide.Instances.Hitl.FullMeshSwCc as FullMeshSwCc
+import qualified Bittide.Instances.Hitl.SyncInSyncOut as SyncInSyncOut
+import qualified Bittide.Instances.Hitl.Transceivers as Transceivers
+import qualified Bittide.Instances.Hitl.VexRiscv as VexRiscv
+
```
lmbollen, line 34: but

bittide-instances/bin/Hitl.hs
```diff
@@ -0,0 +1,139 @@
...
+import qualified Bittide.Instances.Hitl.Tcl.ExtraProbes as ExtraProbes
+
```
lmbollen, line 36: why

bittide-instances/bin/Hitl.hs
```diff
@@ -0,0 +1,139 @@
...
+    { name = show nm
+    , yaml = yamlContents
+    }
+
+-- | Create config based on a 'HitlTests'
+makeConfig :: forall a. C.BitPack a => TH.Name -> HitlTests a -> IO Config
+makeConfig nm config = pure $ Config
+  { name = show nm
+  , yaml = packAndEncode config
+  }
```
lmbollen, line 102: Can we get some output that shows what it did?
Currently we just:
```
[nix-shell:~/bittide-hardware]$ cabal run hitl write-configs
Up to date

```

bittide-instances/bin/Hitl.hs
```diff
@@ -0,0 +1,139 @@
...
+    -- known identifiers
+    fqn :: Maybe String
+  }
+
+-- | Argument parser: currently very boring. Only parses @write-configs@.
+argParser :: Parser Arg
+argParser = subparser $
+  command
+    "write-configs"
+    (info writeConfigsParser (progDesc "Write all known configs to _build/hitl"))
```
lmbollen, line 76: Can you add a hint that you can supply a fully qualified name?

lmbollen, line 76: Perhaps also a `list ` command that lists the targets.

lmbollen (changes requested):
Looks very good!
Some minor requests and suggestions

bittide/src/Clash/Hitl.hs
martijnbastiaan: I like `hitlVioBool`

bittide-instances/bin/Hitl.hs
martijnbastiaan: The problems with adding output is that if every command does that, you'll get drowned in uninteresting messages in a large composed system (like Shake). But, like we discussed, I'll add a `-v` flag.

bittide-instances/bin/Hitl.hs
martijnbastiaan: Because it's an extra level deeper, I've mixed it in now

bittide/src/Clash/Hitl.hs
martijnbastiaan: I don't like nested `where`s. See currently style, which is a compromise. Ack?

bittide/src/Clash/Hitl.hs
martijnbastiaan: Like this?

kleinreact (changes requested):
It's probably me, but grasping the overall design flow of this new infrastructure still seems a bit unclear. I think a short description of the overall architecture and the design principles applied would make things clearer. 

I am also missing kind of a design guide for creating a new test. The simple `()` and enum use cases I get from the adapted existing tests, but how does a `HitlTest a` for some `a` that's not `()` or an `Enum`-instance look like? This case is not covered yet by any existing test making it unclear how the new interfaces is meant to be used under consideration of such extensions. 

This makes it especially hard to build something on top of this.

martijnbastiaan:
Haddock now includes a section on the general flow. In addition, it lists three examples specifying how to create a test config for:

1. A test without config
2. A test with using an Enum
3. A completely custom test

The module documentation now also points out that you don't _have_ to define it in Haskell.

--------------------------

Screenshot 2024-02-26 at 11-36-01 Clash Hitl

kleinreact (approved):
Great update in terms of documentation.

PR #470, opened 2024-02-29
Use proofs instead of tests

kleinreact:
Testing truth statements is kind of bogus, so let's prove that they are valid once and for all and save energy on CI.


martijnbastiaan (approved):
Nice work!

PR #471, opened 2024-03-01
Add HITLT topology support

kleinreact:
The PR adds a new HITL test for testing different topologies.

---

## Depends on

* #490
* #492 
* #494
* #495 
  - #496 
* #499 
* #503


martijnbastiaan (approved):
VERY cool. After we merge this I think we should put some effort in moving everything to one (SW) instance. We currently have quite a bit of duplicated code lying around.

Issue #450, opened 2024-01-10
Topology dependent waiting times


kleinreact:
Clock control simulation fails regularly during the nighty CI.

Investigating the simulated topologies reveals that the values for the reframing waiting times are not appropriate. The problem is that choosing good values for the waiting times depends on the topology. Mostly the hourglass and linear topologies fail to re-center, which is caused by the waiting times being too short. 

Especially for the linear topologies the propagation of clock synchronization between both ends takes long causing the requirement for longer waiting times. We observe that the longer the linear topology, the larger must be the waiting time in general.

It has to be investigated how to better determine waiting times depending on the topology to avoid the issue.

kleinreact:
I mean reframing waiting times. I'll make it more precise in the description.

PR #480, opened 2024-03-14
Reduce simulated topologies to the safe ones

kleinreact:
Some of the simulated topologies sometimes fail on nightly CI. The problem is known (cf. #450) and can be fixed by further increasing the waiting time, but this would also increase the duration of the nightly CI jobs substantially. 

Thus, as there is no benefit of running even longer jobs at night for detecting implementation faults, this PR instead reduces the number of simulation jobs to a safe subset, which is sufficient for identification of potential uprising implementation errors that are directly detected by the simulation runs.


martijnbastiaan (approved):
Great!

PR #482, opened 2024-03-20
Elf mem padding

lmbollen:
The size of the initialized blockRams for our RISC processors are derived from their initial contents. This leads to memory locations that are not included in the initial contents being non-existent, preventing the processor to prefetch into this memory and preventing us from loading larger programs than initially synthesized for.

Ideally we want the ability to separately set the size of the blockrams. However, there are a couple of challenges that make this problem non-trivial. We currently allow blockrams to be initialized with a variety of initial memory types using the `ContentType` and `InitialContent` data structures:

```hs
data ContentType n a
  = Vec (Vec n a)
  | Blob (MemBlob n (BitSize a))
  | BlobVec (Vec (Regs a 8) (MemBlob n 8))
  | File FilePath
  | FileVec (Vec (Regs a 8) FilePath)
  
data InitialContent elements a where
  NonReloadable :: ContentType elements a -> InitialContent elements a
  Reloadable    :: ContentType elements a -> InitialContent elements a
  Undefined     :: (1 <= elements, KnownNat elements) => InitialContent elements a
```
We mainly use `Blob` for synthesizable designs and `Vec` for unittests.
AFAIK we can currently not create blockrams with partially defined initial contents, which is why we can't set the size of the blockrams separately.

This pull request allows us to work around this obstacle by enabling us to pad the `MemBlob` to the desired size.


bittide/src/Bittide/ProcessingElement/Util.hs
```diff
@@ -22,25 +21,43 @@ import qualified Data.ByteString as BS
...
+  -- | Optional size to which we should pad the instruction memBlob and data memBlob
+  (Maybe Int, Maybe Int) ->
   -- | Source file, assumed to be Little Endian.
   FilePath ->
   -- | Optional tuple of starting address and filepath to a device tree.
   Maybe (I.Key, FilePath) ->
   -- | (instruction memBlob, data memBlob)
    Q Exp
-memBlobsFromElf byteOrder elfPath maybeDeviceTree = do
+memBlobsFromElf byteOrder (iSize, dSize) elfPath maybeDeviceTree = do
```
martijnbastiaan, line 50: I think you should error if the given `iSize` > detected `iSize`. Same for `dSize`.

martijnbastiaan (changes requested):

bittide/src/Bittide/ProcessingElement/Util.hs
lmbollen: Already happens in `padToSize`

martijnbastiaan (approved):
`Optional size` => specify "in bytes"

bittide/src/Bittide/ProcessingElement/Util.hs
```diff
@@ -22,25 +22,45 @@ import qualified Data.ByteString as BS
...
 import qualified Data.List as L
 
+-- | Given a `Maybe Int` and a list, if the `Maybe Int` is `Just s`, pad the list to
+-- size `s` with the given element. If the length of the list is greater than `s`,
+-- throw an error. If the `Maybe Int` is `Nothing`, return the list as is.
+padToSize :: HasCallStack => Maybe Int -> a -> [a] -> [a]
+padToSize Nothing _ l = l
+padToSize (Just s) a xs
+  | s >= l = xs <> L.replicate (s - l) a
+  | otherwise = error "Bittide.ProcessingElement.Util: List is longer than the specified size"
```
cuddlefishie, line 32: I guess since this `error` is only hit during TemplateHaskell execution it's easy enough to figure out what exactly is the issue? Mostly concerned about ease of debugging when you encounter this

cuddlefishie (approved):

bittide/src/Bittide/ProcessingElement/Util.hs
lmbollen: Would you like me to extend the error message with the relevant sizes?

cuddlefishie: Since this is only ever run in TemplateHaskell, you could make this function return `Q [a]` and then use `iListPadded <- padToSize "instruction memory" (fmap ((`div` 4) . (+3)) iSize) 0 iList` (and the same for the data memory).

With the name as an argument the error message can be better, and if it's in the `Q` monad you also get to use `reportError`, so if both sizes are wrong both errors will be reported.

PR #486, opened 2024-03-22
Parse Si5395J registers from CSV

martijnbastiaan:
# Old discussion
This PR's purpose has changed from bumping step size to just parsing clock configs from CSV.

<details>
  <summary>Click here to see old cover letter / plots</summary>
_This PR used to increase the step size to 1 PPM, now it just parses CSVs_ 

Testing first commit:

* https://github.com/bittide/bittide-hardware/actions/runs/8392930235

We had to redo the clock configuration, which made ClockBuilder Pro insert a few more (spurious) register values, hence slightly increasing the config type nats.

# Before
~20 seconds to stabilization, buffer occupancy +-4000.

Screenshot from 2024-03-22 17-36-36

# After
2 seconds (or much less?!) to stabilization, buffer occupancy +-6.

Screenshot from 2024-03-22 17-51-29

</details>

# TODO
- [X] Document all functions


kleinreact (changes requested):
I tested this on [mini-bittide](https://github.com/bittide/bittide-hardware/tree/mini-bittide), but still get some unclear results:
## Before 
**(with 100000 FINCs applied to one of the two nodes prior test start)**

#### Clocks
image

#### Elastic Buffers
image

## After
**(with 100 FINCs applied to one of the two nodes prior test start)**

#### Clocks
image

#### Elastic Buffers
image

---

Unfortunately, with 100000 FINCs applied in the "After" case, the test seems to die and does not produce any useful output more, which might be due to some exceeded limits. I'll do some more tests with 1000 and 10000 FINCs to see whether it converges still that fast. If so, we either need to reduce the timeout limit of the test or we need to choose some step size in between 1ppm and 1ppb.

Thus, I formally request changes until the additional tests are done and we hopefully have some better insights on this.

---

Other than that it all looks quite nice. Only the CSV parser still offers some space for improvement, but that is secondary right now.

bittide/src/Bittide/ClockControl/ParseRegisters.hs
```diff
@@ -0,0 +1,83 @@
...
+type Triple = (Integer, Integer, Integer)
+type Lines = [String]
+
+-- | Parse a hex string into an integer.
+--
+-- >>> parseHex "1234"
+-- Right 4660
+-- >>> parseHex "foobar"
+-- Left "parseHex: could not parse: foobar. Got: [(15,\"oobar\")]"
+parseHex :: String -> Either String Integer
```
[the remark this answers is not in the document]
martijnbastiaan: Well.. the extension is CSV but it's not really :woozy_face:. Basically it looks like:

```
# bla bla
# Preamble
<some actual CSV>
# blabla
# Config
<some actual CSV>
# blabla
# Postamble
<some actual CSV>
```

kleinreact:
Here are the results for 1000 FINCs applied (with the update of this PR) to one of the two nodes prior to the test start. The test succeeds and you might be able to observe the fast convergence in the beginning:

### Clocks
image

### Elastic Buffers
image

---

I don't get any results for 10000 FINCs tough, where the test fails and only the following ILA outputs are produced
```console
$ ls _build/vivado/Bittide.Instances.Hitl.FullMeshHwCc.fullMeshHwCcTest/ila-data/CC/*/
Bittide_Instances_Hitl_FullMeshHwCc_fullMeshHwCcTest_fullMeshHwCcTest111_result_128.csv
Bittide_Instances_Hitl_FullMeshHwCc_fullMeshHwCcTest_fullMeshHwCcTest111_result_128.vcd
Bittide_Instances_Hitl_FullMeshHwCc_fullMeshHwCcTest_fullMeshHwCcTest111_result_76.csv
Bittide_Instances_Hitl_FullMeshHwCc_fullMeshHwCcTest_fullMeshHwCcTest111_result_76.vcd
Instances_Hitl_FullMeshHwCc_fullMeshHwCcTest_callistoClockControlWithIla_0_result_131.csv
Instances_Hitl_FullMeshHwCc_fullMeshHwCcTest_callistoClockControlWithIla_0_result_131.vcd
Instances_Hitl_FullMeshHwCc_fullMeshHwCcTest_callistoClockControlWithIla_result_103.csv
Instances_Hitl_FullMeshHwCc_fullMeshHwCcTest_callistoClockControlWithIla_result_103.vcd

$ cat _build/vivado/Bittide.Instances.Hitl.FullMeshHwCc.fullMeshHwCcTest/ila-data/CC/*/*.csv
Sample in Buffer,Sample in Window,TRIGGER,trigger_0,capture_0,probe_milliseconds,probe_allStable0,probe_transceiversFailedAfterUp,probe_nFincs,probe_nFdecs,probe_net_nFincs
0,0,1,1,0,0000,0,0,00000000,00000000,000000000
Sample in Buffer,Sample in Window,TRIGGER,trigger_0,capture_0,probe_milliseconds,probe_allStable0,probe_transceiversFailedAfterUp,probe_nFincs,probe_nFdecs,probe_net_nFincs
0,0,1,1,0,0000,0,0,00000000,00000000,000000000
Sample in Buffer,Sample in Window,TRIGGER,trigger_0,capture_0,probe_milliseconds,probe_allStable0,probe_transceiversFailedAfterUp,probe_nFincs,probe_nFdecs,probe_net_nFincs
Sample in Buffer,Sample in Window,TRIGGER,trigger_0,capture_0,probe_milliseconds,probe_allStable0,probe_transceiversFailedAfterUp,probe_nFincs,probe_nFdecs,probe_net_nFincs
```
This is confusing, as the [IlaPlot ILA](https://github.com/bittide/bittide-hardware/blob/staging/bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs#L533-L541) should produce a completely different header than the ones being present in the generated files. Correct headers are produced for the aforementioned succeeding test.

martijnbastiaan:
10000 FINCs at 1ppm is ~1% of the clock synthesized for. This is well outside of the boundaries where we are guaranteed to have our designs working. It's almost impossible to reason about designs when their digital abstraction is broken, so let's not!

For now, I think we should max out at 100 ppm, which gives us a reasonable chance to still fit within realistic offsets. Even for that we should add constraints though. But.. I do think that's somewhat out of scope for this PR.

martijnbastiaan:
What's cool though: even for an offset of a 1000 ppm, we only need fairly small buffers it seems!

kleinreact:
I'm still not happy with the outcome tough. If I zoom in on the initial 100 FINCs shift then I get

image

so it basically immediately centers. This just does not match the simulation results at all. Simulation chooses random initial offsets below a maximum of 400ppm, so it should be fine to expect some similar results here.

bittide/src/Bittide/ClockControl/ParseRegisters.hs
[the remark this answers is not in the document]
kleinreact: And it seems that you need that "Comment lines" to identify the different sections, otherwise I'd just remove them from the file.

Industry just using standards as they should ... :disappointed:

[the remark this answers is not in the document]
martijnbastiaan: yep yep yep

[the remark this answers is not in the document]
martijnbastiaan: I've reused the parsing logic now. I did simplify its implementation by a lot, I'll ping you on Slack.

martijnbastiaan:
@kleinreact I've split up the PR into this PR and https://github.com/bittide/bittide-hardware/pull/492, making this PR uncontroversial (hopefully).

bittide/src/Bittide/ClockControl/ParseRegisters.hs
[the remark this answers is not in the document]
martijnbastiaan: I haven't switched to cassava (or any other CSV parsing library), because:

* This isn't really CSV.. :unamused:
* I feel the code is still pretty simple/readable
* Over time we're going to move this to Rust, so I doubt getting this in a perfect state is really worth it

bittide-extra/src/Numeric/Extra.hs
```diff
@@ -0,0 +1,28 @@
...
+parseHex :: forall a. BitPack a => String -> Either String a
+parseHex s = do
+  result <- foldM parseDigit (0 :: Integer) s
+  if result > natToNum @(2^BitSize a - 1)
+  then Left $ "Value is out of range: " <> show result
+  else Right $ unpack (fromInteger result)
+ where
+  parseDigit !a c
+    | not (isHexDigit c) = Left $ "Non-hexadecimal digit: " <> [c] <> " in " <> s
+    | otherwise = Right $ shift a 4 .|. fromInteger (toInteger (digitToInt c))
```
martijnbastiaan, line 28:
```suggestion
    | otherwise = Right $ shift a 4 .|. toInteger (digitToInt c)
```

We're already in `Integer` land

bittide-extra/src/Numeric/Extra.hs
```diff
@@ -0,0 +1,28 @@
...
+import Data.Char (digitToInt, isHexDigit)
+
+-- | Parse a hexadecimal string into a 'BitPack'able type.
+--
+-- Note that this function does not handle types that do not use the full range
+-- of their bit size. For example, 'Index' will return an 'XException' if the
+-- parsed value is out of range of the @'Index' n@, but in range of
+-- @'BitVector' ('BitSize' ('Index' n))@. To fix this properly, we need a version
+-- of 'unpack' that returns a 'Maybe' value.
+parseHex :: forall a. BitPack a => String -> Either String a
```
kleinreact, line 19: This introduces a lot of unnecessary `Integer` conversions. Hence, in order to avoid the confusing `log2` function of the previous version and also handle the empty string, how about

```haskell
import Clash.Prelude (BitPack(..), BitSize, (.|.), natToNum, shift)

import Control.Monad (foldM)
import Data.Char (digitToInt, isHexDigit)

-- | Parse a hexadecimal string into a 'BitPack'able type.
--
-- Note that this function does not handle types that do not use the full range
-- of their bit size. For example, 'Index' will return an 'XException' if the
-- parsed value is out of range of the @'Index' n@, but in range of
-- @'BitVector' ('BitSize' ('Index' n))@. To fix this properly, we need a version
-- of 'unpack' that returns a 'Maybe' value.
parseHex :: forall a. BitPack a => String -> Either String a
parseHex "" = Left "Empty String"
parseHex (dropWhile (== '0') -> s) =
  if   not (null s) && isHexDigit (head s)
    && 4 * (length s - 1) + digitBits (head s) > natToNum @(BitSize a)
  then Left $ "Value is out of range: " <> s
  else unpack <$> foldM parseDigit 0 s
 where
  parseDigit !a c
    | not (isHexDigit c) = Left $ "Non-hexadecimal digit: " <> [c] <> " in " <> s
    | otherwise          = Right $ shift a 4 .|. digitToNum c

  digitBits x = case digitToInt x of
    { 0 -> 0; 1 -> 1; 2 -> 2; 3 -> 2; 4 -> 3; 5 -> 3; 6 -> 3; 7 -> 3; _ -> 4 }

  -- 'digitToInt' produces only 16 different values. Hence, it can be
  --  extended to work for any 'Num' instance.
  digitToNum x = case digitToInt x of
    { 0 -> 0; 1 -> 1; 2 -> 2; 3 -> 3; 4 -> 4; 5 -> 5; 6 -> 6; 7 -> 7; 8 -> 8
    ; 9 -> 9; 10 -> 10; 11 -> 11; 12 -> 12; 13 -> 13; 14 -> 14; 15 -> 15
    ; y | y < 0     -> error "digitToInt returned some negative value"
        | otherwise -> error "digitToInt returned some value greater than 15"
    }

```

kleinreact (approved):
Great work and very appealing to also have some tests for `parseHex` now. It's up to you whether you like to apply my proposal for further changes or not.

bittide-extra/src/Numeric/Extra.hs
martijnbastiaan: What integer conversions do you see? I only see this one inside the loop:

```haskell
Right $ shift a 4 .|. toInteger (digitToInt c)
```

(I had a superfluous `fromInteger` in there, woops.) And one outside:

```haskell
fromInteger result
```

Both are _very_ close to being "free". `Int` -> `Integer` is a matter of stripping a constructor [`I#`](https://hackage.haskell.org/package/base-4.19.1.0/docs/GHC-Int.html) and replacing it with [`IS`](https://hackage.haskell.org/package/ghc-bignum-1.3/docs/GHC-Num-Integer.html#v:IS). My guess is that GHC optimizes this to a no-op, because that should be representationally equal. Likewise for `Integer` -> `Natural`, if the integer >= 0 (which is checked), it's just a repack.

Handling the empty string is definitely a good idea, that's something I missed!

martijnbastiaan: If you're worried about `Integer` vs `Natural`: yeah, I believe there should be some overhead. Keep in mind we do use two `Naturals` to implement `BitVector` though, so whether it'll be more inefficient than that... I'm not sure.

If we want to make this as "neat" as possible I suggest we include our own `digitToNatural` (inspired by `digitToInt`):

```haskell
digitToNatural :: Char -> Either String Natural
digitToNatural c
  | dec  <= 9 = Right (dec)
  | hexl <= 5 = Right (hexl + 10)
  | hexu <= 5 = Right (hexu + 10)
  | otherwise = Left ("digitToNatural: not a digit " ++ show c)
  where
    dec  = fromIntegral (ord c - ord '0')
    hexl = fromIntegral (ord c - ord 'a')
    hexu = fromIntegral (ord c - ord 'A')
```

which would remove some conversions, make the accumulator type a `Natural` (instead of `Integer`), and eliminate the `isHexDigit` check:

```haskell
  parseDigit !a c =
    d <- digitToNatural c
    pure $ shift a 4 .|. d
```

kleinreact: But then you can also use `digitToNum`, as `Natural` has a `Num` instance as well. I actually added that in the first place, because I don't understand why the base libraries restrict this function to `Int` as the extension to any `Num` instance turns out to be quite "natural" as well :stuck_out_tongue_winking_eye:.

And we should avoid copying code, even if it is inspired by the base libraries.

martijnbastiaan: Yeah `digitToNum` is better. I also don't know why `Int` is so pervasive in `base`. 

> And we should avoid copying code, even if it is inspired by the base libraries.

I of course agree general, but this also has its limits. We're dealing with a function that has some downsides (its partial and not polymorphic enough) and is very unlikely to change in implementation. So :shrug:.

martijnbastiaan: Okay I think we've rabbit holed this discussion enough :-). I'm going to press the merge button :eyes:.

martijnbastiaan: (Ignore the accidentally broken commit :woozy_face:, fixed now)

PR #488, opened 2024-03-27
Store post processed plot data as CSV

kleinreact:
This PR adds an additional CSV writer to the post-process plotting tool, which stores the recovered data used for generating the plots in some additional CSV files.

The generated files are structured as follows:

* **Index:** The row index
* **Synchronized Time (fs):** The reference time shared along all nodes in fs (absolute since the test start).
* **Local Clock time (fs):** The local time measured by the particular node in fs (absolute since the test start).
* **Integrated FINC/FDECs:** The running sum of the FINCs and FDECs according to the sign function, i.e.,
  *  `FINC` → 1 
  *  `FDEC` → -1
* **Reframing State:** The state of the reframing state machine _(note that reframing is currently disabled for the HITL tests)_
  * `Detect`: Waiting for the elastic buffers to become stable
  * `Wait`: Waiting for the waiting time to elapse (red marking within the plots)
  * `Done`: Waiting time has elapsed
* **EB ℓ**: The occupancies of the elastic buffer corresponding to link ℓ
* **ℓ is stable**: boolean flag of the stability detector indicating whether it determines the occupancies of the elastic buffer corresponding to link ℓ to be stable
* **ℓ is settled**: boolean flag of the stability detector indicating whether it determines the occupancies of the elastic buffer corresponding to link ℓ to be stable and centered

### DIFF
--- a/bittide-tools/clockcontrol/plot/Main.hs
+++ b/bittide-tools/clockcontrol/plot/Main.hs
@@ -20,7 +20,7 @@ import Clash.Prelude
 
 import Clash.Signal.Internal (Femtoseconds(..))
 import qualified Clash.Sized.Vector as Vec
-  (unsafeFromList, toList, repeat, zip, zipWith)
+  (unsafeFromList, toList, repeat, zip, zipWith, indicesI)
 
 import GHC.TypeLits
 import Data.Type.Equality ((:~:)(..))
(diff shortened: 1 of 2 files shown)


bittide-tools/clockcontrol/plot/Main.hs
```diff
@@ -152,6 +154,35 @@ data DataPoint (n :: Nat) (m :: Nat) =
...
+      ]
+   where
+    b2bs = \case
+      False -> "0"
+      True  -> "1"
+
+    rf2bs = \case
+      RSDetect -> "0"
+      RSWait   -> "1"
+      RSDone   -> "2"
```
martijnbastiaan, line 184: Why not:

```suggestion
    rf2bs = \case
      RSDetect -> "Detect"
      RSWait   -> "Wait"
      RSDone   -> "Done"
```

?

martijnbastiaan (approved):

bittide-tools/clockcontrol/plot/Main.hs
kleinreact: The numbers are easier to parse and don't distract too much if you look at the CSV file with an editor, but I can also produce strings instead. Any preference with respect to that?

martijnbastiaan: I think I prefer strings, just because you don't need a separate (mental) mapping when looking at / parsing the CSV file.

PR #489, opened 2024-03-28
Add generated CSVs to artifacts

kleinreact:
The PR adds a new artifact `CC-HITLT Plot Sources` containing the data in CSV used for the HITLT plot generation.


martijnbastiaan (approved):
I'm assuming you checked it ends up in the artifacts properly 👀

PR #490, opened 2024-03-28
Support postproc data in HITLT configs

kleinreact:
This PR adds support for placing additional data to the HITLT configs, which is not required by the HITL test itself, but can be utilized by later post processing steps.


martijnbastiaan (commented):
This approach looks sensible to me. I'm assuming what's left is integrating this in Shake?

kleinreact:
There are no remaining shake related tasks for this PR. These tasks will be integrated with the upcoming post processing tools that utilize these configurations in a separate PR. 

I prefer to integrate the necessary steps iteratively one after another here.

martijnbastiaan (approved):

martijnbastiaan:
LGTM then!

PR #497, opened 2024-04-03
Run Docker in priviliged mode with host mappings on HITL runner

martijnbastiaan:
`--privileged` is needed to access USB devices (i.e., JTAG), while `--usersns host` is needed to get the former to work. This is done as part of a separate PR to give everyone the chance to rebase on top of it, as running CI jobs with and without the flags causes permission errors as the different docker instances will write files under different users. Also see https://stackoverflow.com/a/65964376.


martijnbastiaan:
@lmbollen I'll take the blame if this messes things up..

lmbollen (approved):

PR #500, opened 2024-04-05
Upgrade to GHC 9.4.8

christiaanb:


martijnbastiaan (approved):

martijnbastiaan:
@christiaanb How did you fix the `zlib` issue?

DigitalBrains1:
Drive-by comment: why not GHC 9.6?

martijnbastiaan:
I think the Cabal issue we saw earlier prevented us from upgrading one of the dependencies.

christiaanb:
> @christiaanb How did you fix the `zlib` issue?

It was a cabal.project.freeze issue, in the original PR it disabled the pkgconfig flag for the zlib Haskell package; I simply put it back into the state it was previously: enabled.

bittide-experiments/src/Bittide/Topology.hs
```diff
@@ -256,52 +256,55 @@ instance FromJSON a => FromJSON (TopologyType IO a) where
    where
     tmm = typeMismatch "Topology" v
 
+newtype FUN a = FUN (forall n . SNat n -> a)
```
kleinreact, line 259: That's an interesting new requirement. Also means we don't need `ImpredicativeTypes` any more.

kleinreact (commented):
LGTM

bittide-experiments/src/Bittide/Topology.hs
christiaanb: GHC was no longer happy with `Maybe (forall n . SNat n -> a)`

martijnbastiaan:
@DigitalBrains1 I looked it up, it was indeed the GHC/Cabal issue:

https://github.com/clash-lang/clash-vexriscv/pull/15:

> I've left 9.6 out, as it is affected by the issues reported in https://github.com/clash-lang/clash-compiler/pull/2665.

DigitalBrains1:
Thanks! That's a good thing to have documented here.

PR #509, opened 2024-04-19
Improve the clock reports

kleinreact:
The PR fixes several issues with the data that is presented by the clock control reports and adds some new features:

* Disabled nodes and links have been removed from the plots.
* The descriptions in the configuration table have been adapted to suite both: 
   the HITLT and the simulation report.
* A mapping to the corresponding FPGA IDs has been added to the HITLT report.
* A legend for the different markings in the elastic buffers plots has been added.
* Several minor layout issues have been fixed.

### DIFF
--- a/bittide-experiments/src/Bittide/Report/ClockControl.hs
+++ b/bittide-experiments/src/Bittide/Report/ClockControl.hs
@@ -29,10 +29,12 @@ generateReport ::
   -- ^ Document description header
   FilePath ->
   -- ^ Directory containing the intermediate plot results
+  [(Integer, String)] ->
+  -- ^ Node identifiers
   SimConf ->
   -- ^ The utilized simulation configuration
   IO ()
-generateReport (("Bittide - " <>) -> header) dir cfg =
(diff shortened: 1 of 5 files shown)


martijnbastiaan:
I have a (hopefully small) feature request: could we scale the time units depending on the max value? I.e., switch to milliseconds where it makes sense. Similarly, I feel it makes sense to make the vertical axis on the clock plot ppm (if possible). Disregard if that's too difficult to piggyback to this PR.

lmbollen (approved):
LGTM, noticed and discussed a small 1e3 factor problem with the vertical axis.
Let's discuss the PPM representation later.

kleinreact:
There was a bug in the calculation of the clock drift (thanks to @lmbollen for pointing that out), which now got fixed by the PR as well. Furthermore, there is also a fixed $3*10^6$ fs offset between the 125Mhz clock, used as the static reference clock, and the dynamic 200MHz clock, which we now just statically derive from the domains to cancel it out. This way, we end up with offsets close to zero again also fixing the aforementioned scaling problem, as large numbers are no longer involved.

PR #511, opened 2024-04-22
Smoltcp bitide instance

lmbollen:
A bittide instance containing Vex risc and all components necessary for ethernet communication


bittide/src/Bittide/Ethernet/Mac.hs
```diff
@@ -0,0 +1,451 @@
...
+    miiSel txClkEna rxClkEna txAxiData txAxiKeep txAxiValid txAxiLast txAxiUser
+    gmiiRxData' gmiiRxValid' gmiiRxError' rxAxiReady
+
+  ethStatus = makeEthStatus <$> txFifoUnderflow <*> txFifoOverflow <*> txFifoBadFrame
+    <*> txFifoGoodFrame <*> rxBadFrame <*> rxBadFcs <*> rxFifoOverflow <*> rxFifoBadFrame
+    <*> rxFifoGoodFrame
+
+  rxAxiM2S = makeAxi <$> rxAxiValid <*> rxAxiData <*> rxAxiKeep <*> rxAxiUser <*> rxAxiLast
+
+-- | Utility function to create an Axi4StreamM2S from it's components.
```
martijnbastiaan, line 188:
```suggestion
-- | Utility function to create an Axi4StreamM2S from its components.
```

martijnbastiaan (changes requested):
Could you please remove the unused code / move it to another draft PR?

bittide-shake/src/Clash/Shake/Vivado.hs
```diff
@@ -68,6 +69,8 @@ meetsTiming path = andM
   , fmap not $ inFile "Timing constraints are not met" path
   ]
 
+type GlobPattern = String
```
martijnbastiaan, line 72: Perhaps:

```haskell
type TclGlobPattern = String
```

And make sure to point out https://www.tcl.tk/man/tcl8.6/TclCmd/glob.htm

bittide-instances/src/Bittide/Instances/Pnr/Ethernet.hs
```diff
@@ -0,0 +1,192 @@
...
+        [PortName "TX_P", PortName "TX_N"]
+      ]
+  }) #-}
+
+
+-- TODO: Add more constraints?
+-- | Convert a reset signal from one domain to another domain and insert a register.
+-- This can be used to move a reset signal from a domain with a synchronous reset to
+-- a domain with an asynchronous reset.
+unsafeResetDesynchronizer ::
```
martijnbastiaan, line 181: It's still synchronizing, just to other specifications. I also think it's not quite correct; you want to sample in the target domain with multiple ffs _or_ specify that users should be very careful and make sure that these domains have the same clock. Unless there is a good reason to care about that single flipflop, just use `xpmResetSynchronizer`.

martijnbastiaan (approved):
In general LGTM, could you look at my comments?

bittide-instances/src/Bittide/Instances/Pnr/Ethernet.hs
lmbollen: Actually we don't want any logic or state in the "target" domain.
We use this for the asynchronous reset port on of `gmiiSgmiiBridge` which, according to https://docs.amd.com/v/u/en-US/gig_eth_pcs_pma_ds264 : 
| Signal | Direction | Clock Domain | Description                                          |
|--------|-----------|--------------|------------------------------------------------------|
| reset  | Input     | n/a          | Asynchronous reset for the entire core. Active High. |

The signal is not supposed to be synchronous, but we can't express that in Clash....

martijnbastiaan: `delayReset` (filter glitches) + `unsafeResetSynchronizer` (change domain)

martijnbastiaan: `unsafeResetSynchronizer = unsafeFromBla . unsafeSynchronizer . unsafeToBla`

PR #512, opened 2024-04-23
Add example of smoltcp echo server

lmbollen:
Add all rust code for the `smoltcp` echo server


firmware-support/bittide-sys/src/smoltcp/axi.rs
```diff
@@ -0,0 +1,104 @@
...
+    fn consume<R, F>(self, f: F) -> R
+    where
+        F: FnOnce(&mut [u8]) -> R,
+    {
+        let buf = self.axi_rx.get_slice();
+
+        // TODO: This is a hack to get around the fact that the buffer is not mutable
+        // in the smoltcp API. We should probably fix this in the smoltcp API.
+        #[allow(clippy::cast_ref_to_mut)]
+        let mutable_buf = unsafe { &mut *(buf as *const [u8] as *mut [u8]) };
```
cuddlefishie, line 73: That's unfortunate but fine as a workaround. This should work fine and all, but my spidey senses tingle because those are fat pointers (addr + len of slice), it would be more wordy, but maybe we could use `slice::from_raw_parts_mut(buf.as_ptr().cast_mut(), buf.len())`?
I don't fully understand this code yet either, so maybe there's some other way for this that could work too? I don't know :woozy_face:

firmware-support/bittide-sys/src/smoltcp/loopback.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    fn receive(&mut self, _timestamp: Instant) -> Option<(Self::RxToken<'_>, Self::TxToken<'_>)> {
+        self.queue.pop_front().map(move |buffer| {
+            let rx = RxToken { buffer };
+            let tx = TxToken {
+                queue: &mut self.queue,
+            };
+            (rx, tx)
+        })
```
cuddlefishie, line 47: What is the `move` moving here? Just `&mut self.queue`? Shouldn't this work without move too?

firmware-binaries/examples/smoltcp_echo/memory.x
```diff
@@ -0,0 +1,18 @@
+/*
+SPDX-FileCopyrightText: 2024 Google LLC
+
+SPDX-License-Identifier: CC0-1.0
+*/
+
+MEMORY
+{
+  IMEM : ORIGIN = 0x80000000, LENGTH = 256K
```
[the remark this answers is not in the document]
lmbollen: I can locally reduce it to 32k for both memories for the release version, but for Debug I need at least 201k for Imem.
Preferably I'd make an enhancement issue to handle this

firmware-support/bittide-sys/src/smoltcp/loopback.rs
lmbollen: I think I'm going to remove the loopback device as it has outlived its use.
(It was pretty useful for development)

firmware-binaries/.cargo/config.toml
```diff
@@ -9,4 +9,3 @@ target-dir = "../_build/cargo/firmware-binaries"
 
 [unstable]
 build-std = ["core,panic_abort"]
-build-std-features = ["panic_immediate_abort"]
```
cuddlefishie, line 12: It would be great if we could select this based on debug/release, but I don't know if that's possible :pensive:

firmware-binaries/examples/smoltcp_echo/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+use smoltcp::socket::tcp::{Socket, SocketBuffer};
+use smoltcp::wire::{EthernetAddress, IpAddress, IpCidr};
+use ufmt::uwriteln;
+
+const CLOCK_ADDR: usize = 0b0011 << 28;
+const DNA_ADDR: *const DnaValue = (0b0111 << 28) as *const DnaValue;
+const MAC_ADDR: usize = 0b1001 << 28;
+const RX_AXI_ADDR: usize = 0b0101 << 28;
+const TX_AXI_ADDR: usize = 0b0110 << 28;
+const UART_ADDR: usize = 0b0010 << 28;
```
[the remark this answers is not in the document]
cuddlefishie: There is `core::ffi::c_void` that you can use for pointers, but later on they are mostly used as `u8` or `u32` pointers, if you're immediately casting those constant numbers to pointers then you might as well make them constant pointers IMO

firmware-support/bittide-sys/src/mac.rs
```diff
@@ -2,13 +2,20 @@
...
     status: *const MacStatus,
 }
 
 impl Mac {
-    pub fn new(addr: usize) -> Mac {
+    /// Creates a new instance of `Mac`.
+    ///
+    /// # Safety
+    /// - `addr` must be the base address of the Mac status peripheral.
+    pub unsafe fn new(addr: usize) -> Mac {
```
cuddlefishie, line 18: Please use a `*const MacStatus` instead a `usize` here

```suggestion
    pub unsafe fn new(addr: *const MacStatus) -> Mac {
```

firmware-binaries/examples/smoltcp_echo/memory.x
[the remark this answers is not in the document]
lmbollen: https://github.com/bittide/bittide-hardware/issues/625

firmware-binaries/.cargo/config.toml
lmbollen: Yeahh, that would be nice. But first we'd need to figure out if we can reduce the Debug size I think

firmware-binaries/examples/smoltcp_echo/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+        if !socket.is_active() && !socket.is_listening() {
+            mac_status = unsafe { *MAC_ADDR };
+            uwriteln!(uart, "listening").unwrap();
+            uwriteln!(uart, "{} ", clock.elapsed()).unwrap();
+            socket.listen(7).unwrap();
+        }
+        if socket.is_open() && socket.may_send() && !socket.may_recv() {
+            socket.close();
+            uwriteln!(uart, "DNA: {:X}", dna).unwrap();
+            let new_mac_status = unsafe { *MAC_ADDR };
```
cuddlefishie, line 99: This is reading `MAC_ADDR` multiple times, those should probably be `MAC_ADDR.read_volatile()`, otherwise the compiler might see those reads, conclude there's nobody writing to it (because it doesn't know about hardware) and then collapse all of those into one read and save the result.

firmware-support/bittide-sys/src/axi.rs
```diff
@@ -116,42 +117,46 @@ impl AxiRx {
...
     pub fn send(&mut self, packet: &[u8]) {
         // Split packet into a slice of words and a slice of bytes
         let len = packet.len();
         let words = len / 4;
-        let (words_slice, bytes_slice) = packet.split_at(words * 4);
+        let (init_slice, bytes_slice) = packet.split_at(words * 4);
 
+        // Create a slice of u32 from first multiple of 4 bytes
+        let words_slice =
+            unsafe { core::slice::from_raw_parts(init_slice.as_ptr() as *const u32, words) };
```
cuddlefishie, line 141: This is not safe, `packet` can be any arbitrary slice, so it doesn't have any alignment guarantees, but for a `&[u32]` it needs to be aligned to 4 bytes. So either this whole function must be marked as `unsafe` and this constraint must be added to the `# Safety` section, or the alignment needs to be checked at runtime to make sure constructing a `&[32]` from the `&[u8]` is safe to do.

Maybe you can use [`.align_to()`](https://doc.rust-lang.org/stable/core/primitive.slice.html#method.align_to) here?

cuddlefishie (changes requested):

firmware-support/bittide-sys/src/axi.rs
lmbollen: Could I instead check for aligned with https://doc.rust-lang.org/std/primitive.pointer.html#method.is_aligned and otherwise panic? I'm afraid this might introduce significant overhead which I'd like to avoid on embedded platforms.

firmware-binaries/examples/smoltcp_echo/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+authors = ["Google LLC"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+bittide-sys = { path = "../../../firmware-support/bittide-sys" }
+riscv-rt = "0.11.0"
+riscv = "^0.10"
+heapless = { version = "0.8", default-features = false}
+smoltcp = {default-features = false, features = ["medium-ethernet", "proto-ipv4", "socket-tcp"] }
```
rslawson, line 19: `smoltcp` should have a version specified, as per the warning in `cargo build`. It appears to be using `0.11.0` currently.

rslawson (commented):
No real complaints about the actual code itself, more just some style/readability things. Except for the one note on lines 96-97 of `firmware-support/bittide-sys/src/axi.rs`. I'm happy with basically everything else otherwise.

firmware-support/bittide-sys/src/axi.rs
```diff
@@ -85,23 +99,24 @@ impl AxiRx {
...
         // We can check by using the utility function read_status
-        if unsafe { self.status.read_volatile() == 0 } {
+        if self.read_status_raw() == 0 {
             return None;
         }
 
         // Get length of the incoming data
         let len = self.packet_length();
+
+        debug_assert!(len > buffer.len(), "Buffer too small to receive packet");
```
rslawson, line 109: Comparison should be the other way around here - if the boolean expression in `debug_assert!` evaluates to `false`, then a panic is triggered. So it should be `debug_assert!(len <= buffer.len(), "Buffer too small to receive packet");`.

firmware-support/bittide-sys/src/axi/self_test.rs
```diff
@@ -13,43 +13,22 @@ const STRING_SIZE: usize = 1024;
...
-            read_rx_packet_length as fn(AxiTx, AxiRx, &mut String<STRING_SIZE>) -> bool,
-            "read_rx_packet_length",
-        ),
-        (
-            send_receive_random_packet as fn(AxiTx, AxiRx, &mut String<STRING_SIZE>) -> bool,
-            "send_receive_random_packet",
-        ),
+    type TestFn<const BUF_SIZE: usize> =
+        fn(AxiTx, AxiRx<BUF_SIZE>, &mut String<STRING_SIZE>) -> bool;
+    // Construct a list of tests with their names.
```
rslawson, line 20: This comment is a duplicate of the one two lines above.

rslawson (commented):
Looks good apart from the previous two comments I left. My only other request would be to run a `./cargo.sh fmt` from the base of the repo before resolving the PR.

rslawson (approved):

PR #515, opened 2024-04-26
Align exported EB data with the selected topology

kleinreact:
The PR reduces the elastic buffer entries of the CSV exports to the available links according to the selected topology.

To this end, the adjacency matrix of the selected topology now matches with the indices used to identify the incoming links of a node via the column headers, i.e,  
* the numbers before the `_` in the files name of the exported CSVs index the rows of the adjacency matrix and
* the numbers in the header descriptions of the `n > 5` columns in the CSV index the column corresponding to the edge in the adjacency matrix.

---

Moreover, the legend plotting for the elastic buffers has been disabled if the number of entries exceeds the available space for placing the legend.

### DIFF
--- a/bittide-experiments/src/Bittide/Plot.hs
+++ b/bittide-experiments/src/Bittide/Plot.hs
@@ -14,6 +14,7 @@ module Bittide.Plot
 import Clash.Prelude (KnownNat, Vec)
 import Clash.Sized.Vector qualified as Vec
 
+import Data.Graph (edges)
 import Data.List (foldl', transpose)
 import Data.Aeson (ToJSON)
 import Data.Bifunctor (bimap)
@@ -61,6 +62,8 @@ plot ::
 plot dir graph =
   uncurry (matplotWrite dir) . Vec.unzip . Vec.imap plotDats
  where
+  edgeCount = length $ edges $ topologyGraph graph
+
   plotDats i =
       bimap
         ( withLegend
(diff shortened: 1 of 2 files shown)


bittide-experiments/src/Bittide/Plot.hs
```diff
@@ -69,8 +72,10 @@ plot dir graph =
         . unzip
         )
         ( foldPlots
-        . fmap (\(j, p) -> withLegend
-                   (p @@ [o2 "label" $ show i <> " ← " <> show j])
+        . fmap ( if edgeCount <= 20
```
martijnbastiaan, line 75:
```suggestion
        -- Too many legend entries don't fit. We picked 20 by simply observing
        -- when legend entries wouldn't fit anymore.
        . fmap ( if edgeCount <= 20
```

bittide-tools/clockcontrol/plot/Main.hs
```diff
@@ -224,13 +232,24 @@ instance Exception CsvParseError
 
 -- | Data post processor.
 postProcess ::
-  forall n k c m .
-  (KnownNat n, KnownNat k, KnownNat c, Monad m, c <= k) =>
-  ConduitT (Capture n c) (DataPoint n k) m ()
-postProcess = scanlC process initDummy
+  forall n k n' k' m.
```
martijnbastiaan, line 235: Not a fan of `n'` and `k'`, the `'` is meaningless to me. Can you rename them such that they are descriptive? Could you also add documentation on what these are / why they differ?

martijnbastiaan (changes requested):

martijnbastiaan (approved):
yesss!

PR #525, opened 2024-05-21
Add user data transmission to transceivers logic

martijnbastiaan:
# TODO
- [x] Test freeze signal in WordAlign
- [x] Process Felix's review comments
- [x] Re-read documentation and fix out-of-date sections
- [x] Add PRBS simulation tests
- [x] Add handshake simulation tests
- [x] ~~Send different data over different links and verify~~ Nice getting-to-know-the-project task?


bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs
```diff
@@ -84,14 +112,43 @@ goTransceiversUpTest refClk sysClk rst rxns rxps miso =
...
+      expectCounter
+      transceivers.rxClocks
+      transceivers.rxResets
+      transceivers.rxDatas
+
+  expectCounterErrorSys =
+      fmap and
+    $ bundle
+    $ zipWith (.&&.) transceivers.linkUps
+    $ zipWith (`xpmCdcSingle` sysClk) transceivers.rxClocks expectCounterError
```
martijnbastiaan, line 134: Not only should we check whether every digit is as expected, we also need to verify that we see user data in the first place. I.e., make the test fail if we didn't count to at least N.

bittide/src/Bittide/Transceiver.hs
```diff
@@ -7,453 +7,550 @@
...
+prettifyMetaBits :: BitVector 8 -> BitVector 12
+prettifyMetaBits bv = pack $
+  let meta = unpack @Meta bv in
+  ( low, low, meta.ready
+  , meta.lastPrbsWord
+  , low, meta.fpgaIndex
+  , low, meta.transceiverIndex
+  )
+
+data TransceiverOptions dom = TransceiverOptions
```
martijnbastiaan, line 143:
```suggestion
data Config dom = Config
```

People should use qualified imports

bittide/src/Bittide/Arithmetic/Time.hs
```diff
@@ -71,3 +79,35 @@ trueFor _ clk rst =
   transF counter = \case
     True -> satSucc SatBound counter
     _    -> 0
+
+-- | Rises after the incoming signal has been 'True' for the specified amount of
+-- time. The time is specified in milliseconds. Example invocation:
```
kleinreact, line 84: The documentation here still can be improved. It took me a bit to realize that the specified time is just `stepSize` * `someLimit`. Why not adding that fact to the documentation as well.

And it should also be noted that this introduces an error if `stepSize` can't  be captured precisely by the selected clock domain, as the formula mentioned above does not take this into account.

bittide/src/Bittide/Arithmetic/Time.hs
```diff
@@ -71,3 +79,35 @@ trueFor _ clk rst =
...
+  forall (stepSize :: Nat) (dom :: Domain) (a :: Type) .
+  ( HasCallStack
+  , KnownDomain dom
+  , KnownNat stepSize
+  , NFDataX a, Bounded a, Counter a, Eq a, Num a
+  ) =>
+  Proxy stepSize ->
+  -- ^ Step size. Use the type aliases of 'Bittide.Arithmetic.Time' for time span
+  -- specification.
+  a ->
```
kleinreact, line 100: Why not naming the argument `counter` here. There is no need for first letting the user find that this is a counter because it's constrained by the `Counter` class.

bittide/src/Bittide/Transceiver.hs
```diff
@@ -7,453 +7,550 @@
...
+defConfig = Config
+  { debugIla = False
+  , debugFpgaIndex = pure 0
+  , resetManagerConfig = ResetManager.defConfig
+  }
 
 -- | Careful: the domains for each transceiver are different, even if their
 -- types say otherwise.
-data TransceiverOutputs n tx rx txS freeclk = TransceiverOutputs
+data Outputs n tx rx txS free = Outputs
```
kleinreact, line 161: Is there a good reasons to package `Input` and `Output` vectors this way? It looks to me like it would just introduce conversion overhead for switching between the two representations.

bittide/src/Bittide/Transceiver/Comma.hs
```diff
@@ -0,0 +1,39 @@
...
+  SNat nCycles ->
+  Clock dom ->
+  Reset dom ->
+  -- | (comma, tx control)
+  ( Signal dom (Maybe (Bytes nBytes))
+  , Signal dom (BitVector nBytes)
+  )
+generator _nCycles@SNat clk rst =
+  ( mux sendCommas (pure $ Just commas) (pure Nothing)
+  , mux sendCommas (pure maxBound)      (pure 0)
```
kleinreact, line 29: Why not
```haskell
-- send commas until the counter reaches it's maximum
mux ((/= maxBound) <$> counter) 
  (pure (Just commas), pure maxBound) 
  (pure Nothing      , pure 0       )
```

bittide/src/Bittide/Transceiver/Prbs.hs
```diff
@@ -0,0 +1,137 @@
...
+  = Down (Index 8096)
+  -- ^ Link is considered down. Needs 127 cycles of \"good\" input to transition
+  -- to 'Up'.
+  | Up
+  -- ^ Link has not seen errors in at least 127 cycles.
+  deriving (Eq, Show, Generic, NFDataX)
+
+isUp :: TrackerState -> Bool
+isUp Up = True
+isUp _ = False
```
kleinreact, line 113: `isUp` is never used.

bittide/src/Bittide/Transceiver/Prbs.hs
```diff
@@ -0,0 +1,137 @@
...
+     prbs_out :: Vec nBits Bit
+     prbs_state :: BitVector polyLength
+     (prbs_state, prbs_out) = mapAccumL goPrbs prbs_reg (reverse $ unpack prbsIn)
+
+     goPrbs :: BitVector polyLength -> Bit -> (BitVector polyLength, Bit)
+     goPrbs bv newBit = (o, bitErr)
+      where
+       o = newBit +>>. bv
+       tap = SNat @(polyLength - polyTap)
+       bitErr = xor newBit (xor (lsb bv) (unpack $ slice tap tap bv))
```
kleinreact, line 101: Is this code replication with respect to

https://github.com/bittide/bittide-hardware/blob/8e9fd44862ac38e641e2511276907b3651d74109/bittide/src/Bittide/Transceiver/Prbs.hs#L68-L70

 intentional?

bittide/src/Bittide/Transceiver/Prbs.hs
```diff
@@ -0,0 +1,137 @@
...
+    BitVector nBits ->
+    ((BitVector polyLength, BitVector nBits), BitVector nBits)
+  go (prbs_reg, prbs_out_prev) prbsIn =
+    ( (prbs_state, pack $ reverse prbs_out)
+    , prbs_out_prev
+    )
+   where
+     prbs_out :: Vec nBits Bit
+     prbs_state :: BitVector polyLength
+     (prbs_state, prbs_out) = mapAccumL goPrbs prbs_reg (reverse $ unpack prbsIn)
```
kleinreact, line 94: Minor: camelCase (`prbsIn`) vs. snake_case (`prbs_out`)

bittide/src/Bittide/Transceiver/ResetManager.hs
```diff
@@ -0,0 +1,186 @@
...
+  | RxWait Statistics (MaxRxTimeoutMs, IndexMs dom 1)
+  -- ^ Wait for the receive side to report it is done _and_ that it can predict
+  -- the data coming from the other side. After /n/ milliseconds (see type) it
+  -- times out. Depending on the value of 'ResetStat's 'rxRetries' it will
+  -- either reset both the receive and the transmit side, or just the receive
+  -- side. If all is well though, move on to 'Monitor'.
+  | Monitor Statistics
+  -- ^ Wait till the end of the universe, or until a link goes down - whichever
+  -- comes first. In case of the latter, the state machine moves to 'StartTx'.
+  deriving (Generic, NFDataX)
```
kleinreact, line 101: If `Statistics` is part of every state anyway, then it is more convenient to leave it from the type an run the state machine on `(Statistics, State)` instead.

bittide/src/Bittide/Transceiver/ResetManager.hs
```diff
@@ -0,0 +1,186 @@
...
+      , rx_data_good
+      )
+
+  initSt :: State dom
+  initSt = StartTx (Statistics
+    { rxRetries=0
+    , rxFullRetries=0
+    , txRetries=0
+    , failAfterUps=0
+    })
```
kleinreact, line 142: No parenthesis need here.

bittide/src/Bittide/Transceiver/ResetManager.hs
```diff
@@ -0,0 +1,186 @@
...
+        | otherwise ->
+          RxWait stats (countSucc cntr)
+
+      -- Monitor link. Move all the way back to 'StartTx' if the link goes down
+      -- for some reason.
+      Monitor stats@Statistics{failAfterUps}
+        | rx_done && rx_good -> Monitor stats
+        | otherwise -> StartTx stats{failAfterUps=satSucc SatBound failAfterUps}
+
+  extractOutput st = case st of
```
kleinreact, line 180: With externalizing `Statistics` from `State` and deriving `Eq`, we could for example reduce this to
```haskell
extractOutput (st, stats) = (st == StartTx, st == StartRx, stats)
```
making it much more readable IMO.

bittide/src/Bittide/Transceiver/Prbs.hs
```diff
@@ -0,0 +1,137 @@
...
+   where
+     prbs :: Vec nBits (BitVector polyLength)
+     prbs = unfoldrI goPrbs prbs_reg
+
+     goPrbs :: BitVector polyLength -> (BitVector polyLength, BitVector polyLength)
+     goPrbs bv = (o,o)
+      where
+       o = newBit +>>. bv
+       tap = SNat @(polyLength - polyTap)
+       newBit = xor (lsb bv) (unpack $ slice tap tap bv)
```
kleinreact, line 70: Slicing a single bit looks a bit cumbersome, why not
```haskell
xor (lsb bv) $ boolToBit $ testBit bv $ natToNum @(polyLength - polyTap)
```

bittide/src/Bittide/Transceiver/WordAlign.hs
```diff
@@ -0,0 +1,178 @@
...
+import Data.Tuple.Extra (curry3)
+
+data State n m = State { prev :: BitVector (n * m), offset :: Index n }
+  deriving (Show, Generic, ShowX, NFDataX)
+
+defState :: (KnownNat n, KnownNat m) => State n m
+defState = State{prev=0, offset=0}
+
+alignSymbol :: forall n . (KnownNat n, 1 <= n) => BitVector n
+alignSymbol = bit (natToNum @(n - 1))
```
kleinreact, line 61: `1 +>>. 0` may be a bit more straightforward to read (and you can remove the explicit forall).

bittide/src/Bittide/Transceiver/WordAlign.hs
```diff
@@ -0,0 +1,178 @@
...
+  , KnownNat n
+  , 1 <= n
+  ) =>
+  -- | Freeze alignment
+  Signal dom Bool ->
+  -- | Data with alignment bits in the byte's MSBs
+  Signal dom (Bytes n) ->
+  -- | 'Nothing' until an alignment symbol has been observed. 'Just' if an alignment
+  -- symbol has been observed.
+  Signal dom (Bytes n)
```
kleinreact, line 109: According to the documentation it seems that some `Maybe a` gets returned, but that's actually not the case. Is the documentation maybe not up to date?

bittide/src/Bittide/Transceiver/WordAlign.hs
```diff
@@ -0,0 +1,178 @@
...
+  -- | Freeze alignment
+  Signal dom Bool ->
+  -- | One hot encoded alignment signal. Behavior is undefined when multiple
+  -- bits or none are asserted.
+  Signal dom (Vec n Bool) ->
+  -- | Data
+  Signal dom (BitVector (n * m)) ->
+  -- | 'Nothing' until an alignment symbol has been observed. 'Just' if an
+  -- alignment symbol has been observed.
+  Signal dom (BitVector (n * m))
```
kleinreact, line 139: Same mismatch in the documentation here.

bittide/src/Bittide/Transceiver/WordAlign.hs
```diff
@@ -0,0 +1,178 @@
...
+    )
+   where
+    mAsIndex = maxBound :: Index (m + 1)
+
+    newOffset
+      | freeze = offset
+      | otherwise = fromMaybe 0 (elemIndex True alignment)
+
+  takeMsbs :: forall i . KnownNat i => BitVector (i * 2) -> BitVector i
+  takeMsbs bv = truncateB (rotateR bv (natToNum @i))
```
kleinreact, line 158: Is there a particular reason that we rotate here? Wouldn't `shiftR` be sufficient already?

bittide/src/Bittide/Transceiver/WordAlign.hs
```diff
@@ -0,0 +1,178 @@
...
+
+-- | Opposite of 'align'. Useful for testing. The first and last word are padded
+-- with zeros. If the given offset is zero, the last word is all zeros.
+-- This function mimics the behavior of the transceiver IP, i.e., it mimics reading
+-- at a certain offset in the stream. (Also see this module's documentation.)
+dealignBytes ::
```
kleinreact, line 164: As this is not synthesizeable and only required for testing, I propose moving it to the test module instead.

bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs
```diff
@@ -125,45 +193,34 @@ transceiversUpTest refClkDiff sysClkDiff syncIn rxns rxps miso =
...
-  (txns, txps, allUp, _stats, spiDone, spiOut) =
-    goTransceiversUpTest refClk sysClk testRst rxns rxps miso
+  (txns, txps, allUp, anyErrors, _stats, spiDone, spiOut) =
+    goTransceiversUpTest fpgaIndex refClk sysClk testRst rxns rxps miso
 
-  startTest :: Signal Basic125 Bool
-  startTest =
-    hitlVioBool
+  failAfterUp = isFalling sysClk testRst enableGen False allUp
+  failAfterUpSticky = register sysClk testRst enableGen False (failAfterUp .||. failAfterUpSticky)
```
kleinreact, line 200: Why not use `sticky` instead?

kleinreact (approved):
These are some great new features, where I only have minor suggestions for improvements (_primarily given in separate comments_).

Beside those, there are only two further observations, which come into mind:
1. The data transmitted as part of the HITL test currently is the same for all nodes in the network. Hence, if the transmitted data is mixed between nodes, we may not observe that with this test. We already have the FPGA indices available, so we could add them to `Bittide.Instances.Hitl.Transceivers.counterStart` introducing just some slight differences in the data being transferred. Nodes know their neighbors according to `Bittide.Instances.Hitl.Setup.fpgaSetup`. Hence, they can use that to check that the received data is correct. This way, we would also get better link configuration check, than it is currently done in `Bittide.Instances.Hitl.LinkConfiguration` (where we disable nodes and check that links don't come up to check that the setup is right).
2. `Bittide.Transceiver.Prbs` claims that `checker` is "self synchronizing", but as far as I can see this is not tested explicitly. Hence, some more tests in that direction an regarding the implemented PRBS31 in general would be desirable.

None of that is critical, however, and can be considered "nice-to-have" features for the moment. I'll keep it to the authors of this PR to decide, whether they wanna skip these, move them to a separate PR, or piggyback them with this one.

Nice work overall!

martijnbastiaan:
Thank you @kleinreact! I do want to add simulation tests for PRBS and the handshake logic before merging this. I'll reflag it for review then.

bittide/src/Bittide/Transceiver/Comma.hs
martijnbastiaan: It needs to be slightly different, due to bundling:

```haskell
generator _nCycles@SNat clk rst = unbundle $
  mux
    (counter .==. pure maxBound)
    (pure (Nothing, 0))
    (pure (Just commas, maxBound))
```

but yes!

bittide/src/Bittide/Arithmetic/Time.hs
martijnbastiaan: The documentation was a mess in other ways too. How do you feel about the current state?

bittide/src/Bittide/Arithmetic/Time.hs
martijnbastiaan: Agreed

bittide/src/Bittide/Transceiver.hs
martijnbastiaan: I'll let you be the judge of whether it's a _good_ reason, but the idea is that one can do something like:

```haskell
outputs = transceiversPrbsN [..]

allUp = and <$> outputs.linkUps
```

..instead of having to manually pick out the `linkUp` signals from vectors.

bittide/src/Bittide/Transceiver/Prbs.hs
martijnbastiaan: Removed

bittide/src/Bittide/Transceiver/Prbs.hs
martijnbastiaan: Not intentional, but I also don't know of a great way to deduplicate it. I can't think of a sensible name for a function computing this bit.

bittide/src/Bittide/Transceiver/WordAlign.hs
martijnbastiaan: Ah clever, I like it.

bittide/src/Bittide/Transceiver/WordAlign.hs
martijnbastiaan: Yeah.. that documentation was a mess. What do you think of the current state?

bittide/src/Bittide/Transceiver/WordAlign.hs
martijnbastiaan: Yes. See other comment.

bittide/src/Bittide/Transceiver/WordAlign.hs
martijnbastiaan: Yep, doesn't make sense for sure!

bittide/src/Bittide/Transceiver/ResetManager.hs
martijnbastiaan: Agreed, done

bittide/src/Bittide/Arithmetic/Time.hs
kleinreact: The current state delivers a more precise description in that sense.

bittide/src/Bittide/Transceiver/WordAlign.hs
kleinreact: Makes more sense now.

bittide/src/Bittide/Transceiver/Prbs.hs
martijnbastiaan: I did this, but it broke the transceivers :melting_face:. I don't really have the time to debug this, so if you don't mind I'll let it be.

bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs
```diff
@@ -28,20 +30,53 @@ import Clash.Explicit.Prelude
...
+  -- ^ Error
+expectCounter clk rst = sticky clk rst . mealy clk rst enableGen go counterStart
+ where
+  go c (Just e) = (c + 1, c /= e)
+  go c Nothing  = (c, False)
 
 -- | Worker function for 'transceiversUpTest'. See module documentation for more
 -- information.
 goTransceiversUpTest ::
+  Signal Basic125 FpgaIndex ->
```
kleinreact, line 79: Better use https://github.com/bittide/bittide-hardware/blob/703b881e8e468e7916dc01b2f8c49ba559ac4649/bittide-instances/src/Bittide/Instances/Hitl/Setup.hs#L22 instead of `FpgaIndex`, to avoid redundancies. The number of FPGAs in use should also not be fixed in `bittide-experiments` IMO.

bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs
```diff
@@ -84,14 +123,43 @@ goTransceiversUpTest refClk sysClk rst rxns rxps miso =
...
+      transceivers.txClocks
+      transceivers.txResets
+      transceivers.txSamplings
+
+  expectCounterError =
+    zipWith3
+      expectCounter
+      transceivers.rxClocks
+      transceivers.rxResets
+      transceivers.rxDatas
```
kleinreact, line 139: This looks redundant.

bittide/src/Bittide/Transceiver/Comma.hs
```diff
@@ -0,0 +1,49 @@
...
+
+import Clash.Explicit.Prelude
+
+import Bittide.Arithmetic.Time (IndexMs)
+import Bittide.SharedTypes (Bytes, Byte)
+import Clash.Class.Counter (Counter(countSuccOverflow))
+
+-- | Number of milliseconds to generate commas for
+defTimeout :: SNat 1
+defTimeout = SNat
```
kleinreact, line 14: I don't see the benefit of offering a default value at this point, as fixing a particular value has no benefit against any other value here. Passing just `d1` to `generator` at every usage looks more reusable and shorter to me. If a value shared among multiple instances is required it still can be declared and knowing that the natural corresponds to milliseconds still requires looking at the documentation anyway. Also the detour via the term level feels kind of strange.

bittide/src/Bittide/Transceiver/Comma.hs
```diff
@@ -0,0 +1,49 @@
...
+  comma :: Byte
+  comma = 0xbc
+
+  commas :: Bytes nBytes
+  commas = pack (repeat comma)
+
+  counter =
+    register
+      clk rst enableGen
+      (0 :: Index ms, 0 :: IndexMs dom 1)
```
kleinreact, line 44: Is there any particular reason for this split, or in other words, can't we simply offer the user the same time interface as for https://github.com/bittide/bittide-hardware/blob/703b881e8e468e7916dc01b2f8c49ba559ac4649/bittide/src/Bittide/Arithmetic/Time.hs#L57-L62 which would be much more intuitive IMO, cf. e.g., https://github.com/bittide/bittide-hardware/blob/703b881e8e468e7916dc01b2f8c49ba559ac4649/bittide-instances/src/Bittide/Instances/Hitl/HwCcTopologies.hs#L335

bittide/src/Bittide/Transceiver/Prbs.hs
kleinreact: Can you document this in a comment then, so we don't mess up with it any time in the future soon?

bittide/src/Bittide/Transceiver/Prbs.hs
kleinreact: `prbsBitStep` maybe? As it is basically the main underlying arithmetic operation executed here.

bittide/src/Bittide/Transceiver/Prbs.hs
kleinreact: And there it is again :upside_down_face:.

bittide/src/Bittide/Transceiver/Comma.hs
martijnbastiaan: The split is there to facilitate runtime reconfiguration of waiting times, because we want to do a parameter sweep at some point to see what the right values for the timeouts / comma generation are. Right now they're chosen (very) arbitrarily.

Similar to: https://github.com/bittide/bittide-hardware/blob/e27f60e3fd216f0643f09dbbe4acdb81f9a5f770/bittide/src/Bittide/Transceiver/ResetManager.hs#L48-L51

bittide/src/Bittide/Transceiver.hs
kleinreact: I don't see any particular overhead in using the following
```haskell
  transceivers =
    transceiverPrbs
      @GthTx @GthRx @Ext200 @Basic125 @GthTx @GthRx
      Gth.gthCore defConfig{debugIla=True, debugFpgaIndex=bitCoerce <$> fpgaIndex}
      <$> ( Input sysClk gthAllReset refClk
              <$> iterateI (+1) 0
              <*> clockPaths
              <*> channelNames
              <*> unbundle (unpack <$> rxNs)
              <*> unbundle (unpack <$> rxPs)
              <*> counters
              <*> repeat (pure True)
              <*> repeat (pure True)
          )

  allUp = and . linkUp <$> transceivers
```
compared to
https://github.com/bittide/bittide-hardware/blob/e27f60e3fd216f0643f09dbbe4acdb81f9a5f770/bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs#L147-L162 + `allUp = and <$> outputs.linkUps`

except that you can't use record syntax in the former case. However, if this is all just for introducing names then it definitely is boilerplate overhead. Especially since you enable `NoFieldSelectors`, which means using for example `linkUp` from other module as a function requires nasty re-introductions, like
```haskell
  linkUp' :: Output a b c dom d -> Signal dom Bool
  linkUp' x = x.linkUp
```
---
We shouldn't introduce tons of definitions just to support a particular coding style. Remember that others may like to use other styles and this way of implementation looks all quite restrictive and cumbersome to me in that sense.

bittide/src/Bittide/Transceiver.hs
```diff
@@ -7,453 +7,556 @@
...
+  , KnownDomain txS
+  , KnownDomain ref
+  , KnownDomain free
+  ) =>
+  GthCore tx rx ref free txS rxS serializedData ->
+  Config free ->
+  Input tx rx ref free rxS serializedData ->
+  Output tx rx txS free serializedData
+transceiverPrbs gthCore opts args@Input{clock, reset} =
+  when opts.debugIla debugIla `hwSeqX` result
```
kleinreact, line 367: Have you checked that this works for `transceiverPrbsN` with `n > 1`, because I had issues with shared names in case of introducing multiple ILAs requiring the introduction of individual name sets to work around: cf. https://github.com/bittide/bittide-hardware/blob/c81b2d06ffc99bf1798421a871f61003c6f89714/bittide-instances/src/Bittide/Instances/Hitl/IlaPlot.hs#L537-L554 _(The case switch is needed here as composing names based on a type level parameter does not work with the blackbox backend at the moment.)_

Hence, if you run into the same issues here, it might be good advise to restrict the ILA debugging feature to `transceiverPrbs` only and remove it from `transceiverPrbsN`.

bittide/src/Bittide/Transceiver.hs
```diff
@@ -7,453 +7,556 @@
...
+         fmap (truncateB @_ @8) rxCtrl0 ./=. pure 0
+    .||. fmap (truncateB @_ @8) rxCtrl1 ./=. pure 0
+    .||.                        rxCtrl2 ./=. pure 0
+    .||.                        rxCtrl3 ./=. pure 0
+
+  prbsOk =
+    mux
+      rxUserData
+      (pure True)
+      (Prbs.tracker rxClock rxReset (anyPrbsErrors .||. alignError .||. rxCtrlOrError))
```
kleinreact, line 512: How about
```haskell
prbsOk = rxUserData 
  .||. (Prbs.tracker rxClock rxReset (anyPrbsErrors .||. alignError .||. rxCtrlOrError))
```

bittide/src/Bittide/Transceiver.hs
```diff
@@ -7,453 +7,556 @@
...
+  , HasDefinedInitialValues free
 
-  Clock refclk ->
-  Clock freeclk ->
+  , KnownDomain rxS
+  , KnownDomain txS
+  , KnownDomain ref
+  , KnownDomain free
+  ) =>
+  GthCore tx rx ref free txS rxS serializedData ->
```
kleinreact, line 362: Why do we need to provide a core for `transceiverPrbs` but not for `transceiverPrbsN`? I would expect a uniform interface here.

bittide/src/Bittide/Transceiver.hs
martijnbastiaan: Yes, it works. Though it is kind of a happy coincidence. See:

```haskell
  -- XXX: Replacing 'zipWithN' with '<$>' and '<*>' triggers a combination of:
  --
  --       * https://github.com/clash-lang/clash-compiler/issues/2723
  --       * https://github.com/clash-lang/clash-compiler/issues/2722
  --
```

bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs
martijnbastiaan: Not sure what to do about it.

bittide/src/Bittide/Transceiver.hs
```diff
@@ -7,453 +7,556 @@
...
   }
  where
-  outputs = zipWith4 (transceiverPrbs refclk freeclk rst) chanNms clkPaths rxns rxps
+  -- XXX: Replacing 'zipWithN' with '<$>' and '<*>' triggers a combination of:
+  --
+  --       * https://github.com/clash-lang/clash-compiler/issues/2723
+  --       * https://github.com/clash-lang/clash-compiler/issues/2722
+  --
+  outputs = zipWith8 go
+    (iterateI (+1) 0)
```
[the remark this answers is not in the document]
martijnbastiaan: Almost! That's actually what I tried to do initially, but

```
indicesI :: KnownNat n => Vec n (Index n)
```

..so it will require the number of FPGAs instantiated to be the same as the debug Index. I.e., that will mean extra constraints and conversion functions which seems a bit much for a debugging feature.

bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs
martijnbastiaan: Though I agree, this is already part of `bittide-experiments`. Can we do this cleanup in another PR?

bittide/src/Bittide/Transceiver/Comma.hs
martijnbastiaan: That's fair. The thing I was doing here was purely moving it to _milliseconds_, which makes the trick to run simulations run faster actually work. I'll add a TODO comment.

bittide/src/Bittide/Transceiver/Prbs.hs
martijnbastiaan: The heck :see_no_evil:

bittide/src/Bittide/Transceiver.hs
martijnbastiaan: > except that you can't use record syntax in the former case. However, if this is all just for introducing names then it definitely is boilerplate overhead.

Yes, this is all for introducing names! Positional arguments aren't that great, especially when there is many of them. It's easy to swap arguments (Does P come before N? Does clock path come before channel name?), you can't see at the instantiation site what the argument names are, and error messages are cryptic. I.e., compare:

```haskell
    <*> repeat (pure True)
    <*> repeat (pure True)
```

with

```haskell
    , txReadys = repeat (pure True) 
    , rxReadys = repeat (pure True) 
```

The latter is much more intelligible, I hope you agree. 

I intentionally implement this because I was getting tired of the bad error messages and counting arguments after introducing new ones. And yes, I wish we had something like anonymous records / row polymorphism to make this a whole bunch easier.

martijnbastiaan:
@kleinreact Adding the PRBS tests actually uncovered an issue with `WordAlign`: I had assumed data was being sent MSB first, instead of LSB first. So I assumed:

msb_vs_lsb_2 drawio

while in reality:

msb_vs_lsb drawio

In other words, to dealign we needed to do `takeMsbs (shiftL (new ++# old) offset)` instead of `takeMsbs (shiftL (old ++# new) offset)`. The reason this worked in practice though was that the PRBS checker would trip after erroneous alignment, making the receive side reset again, and eventually the dice would land on perfect alignment. If you get the VCD data from before this PR you can actually see that: all the incoming data directly from the Xilinx IP ends up being `0xDEAD_BEEF_CA55_E77E`. This also explains the erratic number of retries we saw before these fixes.

In any case; there are some remaining tasks I'd like to move to issues with your blessing:

- [ ] Sending over IDs as data - i.e., doing a (part of) the other link test.
- [ ] Finding out why the logic in the comment https://github.com/bittide/bittide-hardware/blob/73a3a2184a824818d4bc4270787a7c485997d57f/bittide/tests/Tests/Transceiver/Prbs.hs#L80 doesn't work
- [ ] Finding a way of specifying _failing_ properties. For some settings (i.e., mixing LSB first and MSB first), I'd expect properties to fail. In fact, it's pretty crucial that they do - though not for the full parameter sweep.
- [ ] Creating a convenience module with utility functions for `Byte`. We use them fairly often, and its pretty cumbersome to keep writing `8 * n` everywhere - especially in combination with `Index`s.

Finally, there is some test-only code that is part of the library again:

https://github.com/bittide/bittide-hardware/blob/73a3a2184a824818d4bc4270787a7c485997d57f/bittide/src/Bittide/Transceiver/WordAlign.hs#L162-L164

Mostly because:

 - They're tiny definitions
 - I neither wanted to create a full-fledged module for the helper functions they use, nor did I feel like duplicating the helper functions

So it's not perfectly "as it should be", but :shrug:.

bittide/src/Bittide/Transceiver.hs
kleinreact: I still don't get your point. If you don't like positional arguments then use `ApplicativeDo`, e.g.,
```haskell
transceivers =
  transceiverPrbs @GthTx @GthRx @Ext200 @Basic125 @GthTx @GthRx
    defConfig{debugIla=True, debugFpgaIndex=bitCoerce <$> fpgaIndex}
      <$> do transceiverIndex <- iterateI (+1) 0
             channelName <- channelNames
             clockPath <- clockPaths
             rxN <- unbundle (unpack <$> rxNs)
             rxP <- unbundle (unpack <$> rxPs)
             txData <- counters
             pure Input
               { clock = sysClk
               , reset = gthAllReset
               , refClock = refClk
               , txReady = pure True
               , rxReady = pure True
               , ..
               }
```
This way of writing offers the same readability as your approach (IMO), just without the necessity of all the overhead. And users that prefer using positional arguments are also not restricted in using them.

_The only flaw with this extension, I don't like, is that it does not support `in`-free `let`-bindings as the normal do notation does. You can use such let bindings, but GHC then always enforces the context to be monadic instead of applicative, which makes no sense somehow._

bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs
kleinreact: That's fine for me. I added #539 to keep track of it.

bittide/src/Bittide/Transceiver.hs
martijnbastiaan: I used `<$>`/`<*>` before but that triggers the ILA/VIO Clash bug mentioned elsewhere..

Other than the bug, I didn't think of using `ApplicativeDo` like this before, this is interesting! I'm still a bit on the fence, because you do push the boiler plate around at bit. But instead of doing it once, you have to do it at every call site:

```haskell
      <$> do transceiverIndex <- iterateI (+1) 0
             channelName <- channelNames
             clockPath <- clockPaths
             rxN <- unbundle (unpack <$> rxNs)
             rxP <- unbundle (unpack <$> rxPs)
             txData <- counters
```

I.e., for _N_ arguments you add _N_ lines of boilerplate on call sites. Not that that's necessarily bad, boilerplate is much less of a concern than hard-to-inspect code.

Initially I tried getting it to work using [barbies](https://hackage.haskell.org/package/barbies), preventing the overhead of having to define both `Input` and `Inputs`. Unfortunately, I never got it t work for reasons I've forgotten..

In any case, named arguments for functions this big are definitely the way to go, there is no doubt in my mind. (I even think named arguments is the biggest advantage Python has over Haskell in terms of readability!) Given that the `ApplicativeDo` approach is currently being blocked by Clash bugs, I think we should just leave it.

bittide-instances/src/Bittide/Instances/Hitl/Transceivers.hs
kleinreact: Me neither. I initially thought there may be an easy improvement, but that doesn't seem to be the case.

bittide/src/Bittide/Transceiver/Comma.hs
kleinreact: I think you did not get my point. My proposal is to completely remove `defTimeout` from this module and replace every usage of it in the remaining code with `d1`. Is there any particular argument against this change?

bittide/src/Bittide/Transceiver.hs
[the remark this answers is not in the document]
kleinreact: That's fair. I didn't consider that point. 

Nevertheless, we should add a note somewhere then, as I am a bit worried that we simply oversee the `Unsigned 3` once we change the setup to more than 8 nodes and run into issues then.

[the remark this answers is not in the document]
martijnbastiaan: I can also just removed the feature entirely. It's just for debugging / sanity checking, the actual code does not rely on it.

bittide/src/Bittide/Transceiver/Comma.hs
martijnbastiaan: It's only used in one place, so no, I can remove it. I could even make it entirely unconfigurable to begin with.

bittide/src/Bittide/Transceiver.hs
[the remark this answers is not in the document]
kleinreact: No. I would keep it, as you only introduce extra work in case you need it again in the feature otherwise. Just add a note to the comment.

bittide/src/Bittide/Transceiver.hs
kleinreact: Can you please add another note to the code then. The above comment only legitimates the use of `zipWithN` instead of Applicative primitives, but does not mention that the debug ILAs start breaking as well, if not used or the two mentioned issues get fixed. 

We need to inform the person removing that `XXX` note some time in the future, that the ILAs need to be fixed as well as part of the change introduced then. Let's be some kind people of past :wink:.

bittide/tests/Tests/Transceiver/Prbs.hs
```diff
@@ -0,0 +1,132 @@
...
+      cover 20
+        "nNoiseCycles <= expectAlignmentAfterNCycles"
+        (nNoiseCycles <= expectAlignmentAfterNCycles)
+
+      cover 70 "any error detected after 3 cycles" anyNotOk
+      cover 2 "no errors detected after 3 cycles" allOk -- detect "always pass"
+
+implies :: Bool -> Bool -> Bool
+implies a b = not a || b
+
```
kleinreact, line 124: Please remove these lines. They are not used.

kleinreact (changes requested):
> Adding the PRBS tests actually uncovered an issue with WordAlign.

Nice to hear that the extra work also payed off in the end.

> In any case; there are some remaining tasks I'd like to move to issues with your blessing.

I'm fine with that. Let's go with a formal change request then until all issues have been created and linked here.

> So it's not perfectly "as it should be", but :shrug:

Yeah, but the PR introduces a lot of great new features, nevertheless, which also should be taken into account. It's a great work overall.

kleinreact (approved):
Thanks.

PR #538, opened 2024-06-05
Use transceiver clocks in FINC/FDEC tests

hiddemoll:
The HITL test `fincFdecTests` is no longer run when a branch is pushed, so I've added a debug.json to explicitly test it. The result of that test can be seen [here](https://github.com/bittide/bittide-hardware/actions/runs/9761995422).

The main goal of this PR is to enable us to run the FINC/FDEC tests on all FPGAs in the demo rig. To do this, we need access to the external clock, which is connected to `SMA_MGT_REFCLK_C` and which can only be used by the transceivers. We therefore modified `gthCore` to allow access to this clock before it goes through the internal CPLL (see this [diagram in the GTH Transceiver user guide](https://docs.amd.com/api/khub/documents/X8hVhAx~JVBkxiZRAjsdRg/content?Ft-Calling-App=ft%2Fturnkey-portal&Ft-Calling-App-Version=4.3.48#G5.290097). 

To select `TXPLLREFCLK_DIV1` as `TXOUTCLK`, we need a different ratio between the reference clock and 125 MHz TX clock. We therefore changed the external clock from 200 MHz to 250 MHz for all bittide instances. Note that this does not interfere with internals of any bittide instance, it only changes the reference clock of the transceivers.


bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
```diff
@@ -36,13 +42,32 @@ data Test
   | FDecInc
   -- | 'FInc' test followed by an 'FDec' one
   | FIncDec
-  deriving (Enum, Generic, NFDataX, Bounded, BitPack, ShowX, Show)
+  deriving (Enum, Eq, Generic, NFDataX, Bounded, BitPack, ShowX, Show)
+
+gthResetManagerWithoutRx ::
```
martijnbastiaan, line 47: Please add a comment explaining the function and its purpose

martijnbastiaan (approved):

PR #551, opened 2024-06-17
Improve reset behaviour of transceiver with no receive clock or data

hiddemoll:
Up until now we have only used the transceivers in cases where they receive data. We now also want to use the internal `gthCore` to only extract the tx clock from the transceiver PLL. The reset bahaviour of the transceivers was incorrect, but worked because we keep resetting the whole component when the word alignment of the received data failed. This PR improves the reset behaviour of the transceivers when no receive channel is connected.

Note that no specific test is added in this PR, for a test which uses these changes see #538 .

### DIFF
--- a/bittide/src/Bittide/Transceiver/ResetManager.hs
+++ b/bittide/src/Bittide/Transceiver/ResetManager.hs
@@ -143,7 +143,9 @@ resetManager config clk rst tx_init_done rx_init_done rx_data_good =
   update :: (State dom, Statistics) -> (Bool, Bool, Bool) -> (State dom, Statistics)
   update st (tx_done, rx_done, rx_good) =
     case st of
-      -- Reset everything:
+      -- The reset of the PLLs and data directions is triggerd by the fallig
+      -- edge of the active-high reset_all signal. So the actual reset is
+      -- initialized one cycle later.
       (StartTx, stats) -> (TxWait minBound, stats)
 
       -- Wait for transceiver to indicate it is done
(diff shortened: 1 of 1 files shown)


martijnbastiaan (approved):
```haskell
      -- The reset of the PLLs and data directions is triggerd by the fallig
      -- edge of the active-high reset_all signal.
```

Amazing!

For the rebase: the logic has moved to `ResetManager.hs`, but should otherwise be the same.

bittide/src/Bittide/Transceiver/ResetManager.hs
```diff
@@ -171,8 +176,13 @@ resetManager config clk rst tx_init_done rx_init_done rx_data_good =
...
       -- Monitor link. Move all the way back to 'StartTx' if the link goes down
-      -- for some reason.
-      (Monitor, stats@Statistics{failAfterUps})
+      -- for some reason. On startup `tx_done` can be True while it should not
+      -- be. When the receiver is not connected, this causes the reset manager
+      -- to claim it is done. We therefore also check `tx_done` in `Monitor`,
+      -- going back to `StartTx` when it goes down. We expect this to only
+      -- happen once.
+      (Monitor, stats@Statistics{txRetries, failAfterUps})
+        | not tx_done -> (StartTx, stats{txRetries=satSucc SatBound txRetries})
```
martijnbastiaan, line 185: Isn't this a `failAfterUp`?

hiddemoll: Well, it would be if it properly came here through `RxWait`. But when the Rx side is not connected, and both `rx_done` and `rx_good` are `pure True`, did it really fail after being up? Or was it never really up and should it retry Tx?

martijnbastiaan:
Turns out to not be needed anymore after talking to @hiddemoll

PR #573, opened 2024-07-08
`check_copyright_years` check ranges instead of exact match

lmbollen:
Currently the copyright year check will fail on `2023-2024` if the commit was created in 2023, but altered in 2024 because it looks for an exact match on the creation date.

This change changes the check to check if the creation date is within the and the highest year instead of exactly matching on the creation year.


martijnbastiaan:
Though I guess I should talk to @buzzardpeaceful before we go through with this.. Please stand by.

martijnbastiaan:
Let's go with creation date :heavy_check_mark:

martijnbastiaan (approved):

PR #575, opened 2024-07-08
Update register-runners.sh

leonschoorl:
Update to v2.317.0
And add labels automatically.

### DIFF
--- a/.github/scripts/register_runners.sh
+++ b/.github/scripts/register_runners.sh
@@ -1,15 +1,21 @@
 #!/bin/bash
-# SPDX-FileCopyrightText: 2022 Google LLC
+# SPDX-FileCopyrightText: 2022,2024 Google LLC
 #
 # SPDX-License-Identifier: Apache-2.0
 #
 # Usage:
 #
-#    register_runners <n_runners> <github_token>
+#    register_runners <n_runners> <labelA,labelB,...> <github_token>
 #
 # Example:
 #
-#    register_runners 8 jbg2PSMoQLfqvj0ZeBUW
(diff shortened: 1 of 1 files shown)


.github/scripts/register_runners.sh
```diff
@@ -1,15 +1,21 @@
...
 #
 # Usage:
 #
-#    register_runners <n_runners> <github_token>
+#    register_runners <n_runners> <labelA,labelB,...> <github_token>
 #
 # Example:
 #
-#    register_runners 8 jbg2PSMoQLfqvj0ZeBUW
+#    register_runners 8 self-hosted,compute jbg2PSMoQLfqvj0ZeBUW
```
martijnbastiaan, line 12:
```suggestion
#    register_runners 8 compute jbg2PSMoQLfqvj0ZeBUW
```

`self-hosted` is applied automatically. I'm not sure what happens if you apply it manually too.

martijnbastiaan: Nonetheless, having an example on how to apply multiple labels is nice.

.github/scripts/register_runners.sh
```diff
@@ -1,15 +1,21 @@
...
+#
+# Currently we're labelling runners with either "self-hosted,compute"
+# or "self-hosted,hardware-access".
```
martijnbastiaan, line 15:
```suggestion
# Currently we're labelling runners with either "compute" or "hardware-access".
# Note that a label "self-hosted" gets applied automatically.
```

martijnbastiaan (approved):
That should prevent the issues we saw before the weekend in the future yeah :+1:

(No need to re-review after changes, so :heavy_check_mark:)

PR #577, opened 2024-07-16
Add `readDnaPortE2Wb`

lmbollen:
A wishbone wrapper around [readDnaPortE2](https://github.com/clash-lang/clash-compiler/blob/master/clash-cores/src/Clash/Cores/Xilinx/Unisim/DnaPortE2/Internal.hs#L35).

We need this identifier to create unique MAC addresses for our network applications.


firmware-support/bittide-sys/src/dna_port_e2.rs
```diff
@@ -0,0 +1,33 @@
...
+        let mut result = 0;
+        for _ in 0..3 {
+            unsafe {
+                result <<= 32;
+                result |= tmp.read_volatile() as u128;
+                tmp = tmp.sub(1);
+            }
+        }
+        result
+    }
```
cuddlefishie, line 32:
```suggestion
    /// Reads the DNA value from the DNA port E2 peripheral.
    /// The DNA value is a 96-bit unique identifier, the upper bits from the 128-bit value are zero.
    ///
    /// # Safety
    /// The `addr` pointer must point to valid memory field of at least size 12 bytes.
    pub unsafe fn read(addr: *const u32) -> u128 {
        let mut tmp = addr.add(2);
  
        let mut result = 0;
        for _ in 0..3 {
            result <<= 32;
            result |= tmp.read_volatile() as u128;
            tmp = tmp.sub(1);
        }
        result
    }
```

This function cannot be marked as safe, it must be `unsafe` because
- an arbitrary value is passed in, that's interpreted as a pointer
- that pointer gets dereferenced without any safety checks

So it's really easy to cause undefined behaviour with *safe Rust*, which is the whole point that safe Rust is trying to prevent.

(It would be nice if we could work more with array here instead of `u128` and loops and shifts, but Rust's const generics aren't there yet for that :pensive: )

firmware-support/bittide-sys/src/dna_port_e2.rs
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+/// The DnaPortE2 struct is a hardware abstraction for the DnaPortE2 peripheral,
+/// a unqique identifier for the FPGA device.
+pub struct DnaPortE2 {}
+
+impl DnaPortE2 {
+    /// Reads the DNA value from the DNA port E2 peripheral.
+    /// The DNA value is a 96-bit unique identifier, the upper bits from the 128-bit value are zero.
+    pub fn read(addr: usize) -> u128 {
```
cuddlefishie, line 18: Could this be a freestanding function? Does this need the struct?

cuddlefishie (changes requested):

firmware-support/bittide-sys/src/dna_port_e2.rs
lmbollen: We could indeed just make a freestanding function.

firmware-support/bittide-sys/src/dna_port_e2.rs
lmbollen: I'm also not very happy with this code tbh. I did look at different solutions and found some packages that could represent `u96`, but I myself found that a bit much to add new dependencies just to read a single unsigned number...

bittide/src/Bittide/DoubleBufferedRam.hs
```diff
@@ -386,6 +387,31 @@ blockRamByteAddressableU readAddr newEntry byteSelect =
...
 
 data RegisterWritePriority = CircuitPriority | WishbonePriority
+ deriving Eq
+
+-- | Register with additional wishbone interface, this component has a configurable
+-- priority that determines which value gets stored in the register during a write conflict.
+-- With 'CircuitPriority', the incoming value in the fourth argument gets stored on a
+-- collision and the wishbone bus gets acknowledged, but the value is silently ignored.
+-- With 'WishbonePriority', the incoming wishbone write gets accepted and the value in the
+-- fourth argument gets ignored.
```
cuddlefishie, line 397: Nitpick, there's no "fourth argument", this function only has two arguments :nerd_face: 

I guess you mean of the `Circuit` inner function?

firmware-binaries/test-cases/dna_port_e2_test/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+authors = ["Google LLC"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+riscv-rt = "0.11.0"
+bittide-sys = { path = "../../../firmware-support/bittide-sys", features = ["panic-handler"] }
```
cuddlefishie, line 16: the `panic-handler` feature doesn't actually do anything anymore. It wasn't removed from the `Cargo.toml` file of `bittide-sys`, so it's not an error, but I'll do a pass removing it and all references to it soon, would be nice if it wasn't introduced if it doesn't do anything :sweat_smile:

firmware-binaries/test-cases/dna_port_e2_test/rust-toolchain.toml
cuddlefishie, line 1: I want to say "we don't even need that file, why do we even have it?" (it's handled by `.cargo/config`) I think I added it once while doing the Rust crate reorganisation and then it was just copied over :face_with_peeking_eye: 

So not really your fault, but note for me: get rid of all of those unnecessary files :sweat_smile:

PR #588, opened 2024-08-05
small Rust cleanups

cuddlefishie:
This PR
- removes unnecessary `rust-toolchain.toml` files that were left over after refactors
- remove the no-longer used `panic-handler` feature flag in `bittide-sys`
- removes unused toplevel `Cargo.lock` (which also caused dependabot warnings, even though it has no effect on anything)


martijnbastiaan (approved):
Perhaps a dumb question, but why doesn't the top level `Cargo.lock` have any effect?

cuddlefishie:
> Perhaps a dumb question, but why doesn't the top level `Cargo.lock` have any effect?

We don't actually have a single Rust workspace in this repository, we have three (firmware-binaries, firmware-support, host-tools), and some of them build on different compiler versions and targets, that's why we have the `cargo.sh` script :woozy_face:

PR #590, opened 2024-08-05
Don't hardcode IP addresses in CI scripts

leonschoorl:
Apparently the comment about the container being unable to resolve names isn't true (anymore), as it just works.
And this is less fragile.


martijnbastiaan (approved):
Wonderful, not sure what changed!

PR #594, opened 2024-08-07
Move `hwCcTopologyTest` to nightly 

martijnbastiaan:
It takes ~10 minutes to execute, which is too long for our single demo rig

# TODO
- [ ] Let's see if CI survives


hiddemoll (approved):
LGTM. We also get the added effect that report generation is not performed when a debug.json is used!

Issue #598, opened 2024-08-12
Clock reports should use more sensible time units


martijnbastiaan:
We currently produce plots like these:

![Screenshot from 2024-08-12 20-09-13](https://github.com/user-attachments/assets/eeb0ed60-b9c6-40bc-803e-22f3863dedc4)

Note the bottom bar mentions 1.2e16 fs. Or in human terms: about 10 seconds. Our plots should use the latter, not the first, time unit.

PR #599, opened 2024-08-13
Use `ms` instead of `fs` on x-axes

martijnbastiaan:
Fixes https://github.com/bittide/bittide-hardware/issues/598

---------

I couldn't understand the plot code because of all the zipping / uncurrying / anonymity / polymorphism, so I've done a bit of restructuring (in a separate commit). Hopefully this makes it more clear what's going on.

--------

Example: https://github.com/bittide/bittide-hardware/actions/runs/10365635367/job/28695237992


hiddemoll (approved):
LGTM

PR #613, opened 2024-08-21
Rewrite `smoltcp_echo` into `smoltcp_client`

lmbollen:
This pull request changes the `smoltp_echo` example from a TCP echo server to a TCP client that connects to a server and stress tests it by continuously writing for 30 seconds. The client now obtains an IP using DHCP.

Furthermore it contains the following changes:
* Rename `smoltcp_echo` to `smoltcp_client` 
* Write panic info in `smoltcp_echo`
* Add `end_of_time()` to our rust `time` module, which is the largest time we can represent
* Add `Display` traits for `Instant` and `Duration`
* Add `display_source`, `display_level` and `display_time` options to `UartLogger`
* The smoltcp ethernet driver only processes one received packet at a time
   The packet is processed directly from the buffer to prevent us from copying it over first. So we have to keep track of whether or not we've
   already seen that this packet exists.
* `Instant` now stores time in microseconds rather than keeping track of the absolute number of clock cycles and frequency.
* `ilaWb` does not assume anymore when you want to trigger and capture


bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -80,51 +97,62 @@ case_testGdbProgram = do
...
+            assertFailure "Timeout waiting for \"Starting TCP Client\""
+
+          let numberOfClients = 1
+          putStrLn $ "Waiting for " <> show numberOfClients <> " clients to connect to TCP server."
+          maybeClients <- timeout 10_000_000 $ waitForClients numberOfClients serverSock
+          clients <- case maybeClients of
+            Just clients' -> pure clients'
+            Nothing -> do
+              loggingSequence
+              assertFailure "Timeout waiting for clients to connect"
```
leonschoorl, line 140: You're doing almost this three times already.
Maybe factor out something like:
```haskell
foo :: Int -> String -> IO a -> IO a
foo n msg f = do
  res <- timemout n f
  case res of
    Just x -> pure x
    Nothing -> do
      loggingSequence
      assertFailure $ "Timeout waiting for" <> msg
```

And then you can do:
```suggestion
          clients <- foo 10_000_000 "clients to connect" (waitForClients numberOfClients serverSock)
```

Note that `foo` should be locally bound within `case_testGdbProgram` so it can access `loggingSequence`.

bittide-instances/src/Bittide/Instances/Hitl/Post/TcpServer.hs
```diff
@@ -0,0 +1,52 @@
...
+    )
+    [1 .. numberOfClients]
+
+clientDataToFile :: Handle -> Int -> NS.Socket -> IO ()
+clientDataToFile handle bufferSize sock = do
+  start <- getPOSIXTime
+  SBS.toHandle handle $ SBS.reread (const $ NS.recv sock bufferSize) ()
+  stop <- getPOSIXTime
+  size <- hFileSize handle
+  let speed = fromIntegral size / (stop - start)
```
leonschoorl, line 42: You need some extra type conversion here.
Because now `speed :: NominalDiffTime`, and it will be shown with a "s" suffix, because it is supposed to represent seconds.

firmware-support/bittide-sys/src/uart/log.rs
```diff
@@ -44,15 +52,19 @@ impl log::Log for UartLogger {
...
-                            l,
-                            "{} | {}:{} - {}",
-                            record.level(),
-                            record.file().unwrap(),
-                            record.line().unwrap(),
-                            record.args()
-                        )
-                        .ok();
+                        if record.level() <= self.display_level {
+                            write!(l, "{} | ", record.level()).ok();
```
rslawson, line 56: I'm unsure what the purpose of the `.ok()`s here is. All it's doing is converting a `Result<T, E>` into an `Option<T>` by dropping the `E` value. If the use case here is to ignore the result, then I would suggest instead explicitly ignoring it by writing:
```
let _ = write!(l, "{} | ", record.level());
```

firmware-support/bittide-sys/src/uart/log.rs
```diff
@@ -44,15 +52,19 @@ impl log::Log for UartLogger {
...
+                        }
+                        if record.level() <= self.display_source {
+                            write!(
+                                l,
+                                "{}:{} - ",
+                                record.file().unwrap(),
+                                record.line().unwrap()
+                            )
+                            .ok();
```
rslawson, line 65: See comment on line 56.

firmware-support/bittide-sys/src/uart/log.rs
```diff
@@ -44,15 +52,19 @@ impl log::Log for UartLogger {
...
+                        }
+                        writeln!(l, "{}", record.args()).ok();
```
rslawson, line 67: See comment on line 56.

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -80,51 +97,62 @@ case_testGdbProgram = do
...
+
+          putStrLn "Writing client data to files"
+          createDirectoryIfMissing True tcpDataDir
+          mapConcurrently_
+            ( \(sock, addr) -> do
+                (path, handle) <- openTempFile tcpDataDir (show addr)
```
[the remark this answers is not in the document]
lmbollen: Good question, at the moment it's useless garbage data I'll rewrite it and just check if I receive what I expect instead.

firmware-support/bittide-sys/src/axi.rs
```diff
@@ -110,7 +110,8 @@ impl<const BUF_SIZE: usize> AxiRx<BUF_SIZE> {
         debug_assert!(len <= buffer.len(), "Buffer too small to receive packet");
 
         unsafe {
-            core::ptr::copy_nonoverlapping(self.base_addr, buffer.as_mut_ptr(), len);
+            let dst = self.base_addr as *mut u8;
+            core::ptr::copy_nonoverlapping(dst, buffer.as_mut_ptr(), len);
```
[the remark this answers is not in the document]
lmbollen: Nice catch :)

firmware-support/bittide-sys/src/time.rs
```diff
@@ -415,3 +424,26 @@ impl uDisplay for Duration {
...
+
+impl fmt::Display for Duration {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let hours = self.to_hours();
+        let mins = self.to_mins() % 60;
+        let secs = self.to_secs() % 60;
+        let millis = self.to_millis() % 1000;
+        let micros = self.to_micros() % 1000;
+
+        write!(f, "{}:{}:{}.{}.{}", hours, mins, secs, millis, micros)
```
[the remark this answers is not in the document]
lmbollen: Didn't know that! Nice :)

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -80,51 +100,78 @@ case_testGdbProgram = do
...
         hSetBuffering picocomStdOutHandle LineBuffering
 
-        putStr "Waiting for Picocom to be ready..."
-        hFlush stdout
-        waitForLine picocomStdOutHandle "Terminal ready"
-        putStrLn " Done"
+        putStrLn "Waiting for Picocom to be ready..."
+        _ <-
+          tryWithTimeout "Timeout waiting for Picocom to be ready" 10_000_000 $
+            waitForLine picocomStdOutHandle "Terminal ready"
```
leonschoorl, line 130: You shouldn't need to explicitly ignore the result as it's `()`
```suggestion
        tryWithTimeout "Timeout waiting for Picocom to be ready" 10_000_000 $
          waitForLine picocomStdOutHandle "Terminal ready"
```

rslawson (approved):
No issues that I can see in the Rust code (:

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -80,51 +100,78 @@ case_testGdbProgram = do
...
+          loggingSequence
+
+runTcpTest :: (NS.Socket, NS.SockAddr) -> Assertion
+runTcpTest (sock, sockAddr) = do
+  putStrLn $ "Testing TCP connection from " <> show sockAddr
+  start <- getPOSIXTime
+  bs <- SBS.toLazy_ $ SBS.reread (const $ NS.recv sock (1024 * 1024)) ()
+  stop <- getPOSIXTime
+  let size = BS.length bs
+  let speed = fromIntegral size / (stop - start)
```
leonschoorl, line 161: You need some extra type conversion here.
Because now `speed :: NominalDiffTime`, and it will be shown with a "s" suffix, because it is supposed to represent seconds.

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -0,0 +1,177 @@
...
+
+          putStrLn "Waiting for \"Starting TCP Client\""
+          _ <-
+            tryWithTimeout "Timeout waiting for \"Starting TCP Client\"" 10_000_000 $
+              waitForLine picocomStdOutHandle "Starting TCP Client"
+
+          let numberOfClients = 1
+          putStrLn $ "Waiting for " <> show numberOfClients <> " clients to connect to TCP server."
+          clients <-
+            tryWithTimeout "Timeout waiting for clients to connect" 10_000_000 $
```
leonschoorl, line 144: You could put `"Timeout waiting for clients "` part inside `tryWithTimeout` so you don't have to repeat it

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -0,0 +1,177 @@
+-- SPDX-FileCopyrightText: 2024 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE NumericUnderscores #-}
+{-# OPTIONS_GHC -Wno-unused-do-bind #-}
```
leonschoorl, line 5: I don't think you need to disable this warning

bittide-instances/src/Bittide/Instances/Hitl/Post/TcpServer.hs
leonschoorl: If you delete `clientDataToFile` then you can ignore this ^

firmware-binaries/examples/smoltcp_echo/memory.x
```diff
@@ -6,8 +6,8 @@ SPDX-License-Identifier: CC0-1.0
 
 MEMORY
 {
-  IMEM : ORIGIN = 0x80000000, LENGTH = 256K
-  DMEM : ORIGIN = 0x10000000, LENGTH = 64K
+  IMEM : ORIGIN = 0x80000000, LENGTH = 1024K
+  DMEM : ORIGIN = 0x10000000, LENGTH = 256K
```
hiddemoll, line 10: Do you really need 4 times more memory? We already use a lot IMO.

hiddemoll (approved):
I've not looked at the Rust code, as Ryan has already given an extensive review of that. I have no remarks on the other changes, other than that the last few commits look a bit WIP, which they are.  I'll approve now if you clean up the last commits when you're done debugging.

firmware-binaries/examples/smoltcp_echo/memory.x
lmbollen: We already have 4 times more memory, because currently we define
1024k and 256k "elements" of `BitVector 32`:
https://github.com/bittide/bittide-hardware/blob/staging/bittide/src/Bittide/DoubleBufferedRam.hs#L42-L45
https://github.com/bittide/bittide-hardware/blob/staging/bittide/src/Bittide/DoubleBufferedRam.hs#L190

It's a bit misleading, but not trivial to fix since we do need `Initialcontents d (Bytes 4)`

Issue #603, opened 2024-08-14
Clock reports should have cover page explaining how to interpret the results


martijnbastiaan:
In particular, it should explain:

* How relative frequencies are determined
  * Explain (potentially) different domains
  * Mention how clock plots always end up <= -100
* What reframing (approximately) is
* What the stability detector does / how it works
* What clock offsets are
* What startup delays are

PR #619, opened 2024-08-26
Add label to Y-axis for buffer occupancy plots

martijnbastiaan:
Bikeshed away! Mostly because I want to use it in context where the title of the plot disappears.

Example of how it looks:

Screenshot from 2024-08-26 16-13-57

Current proposal:

 - **Occupancy count**


lmbollen:
Why would the occupancies be virtual?

hiddemoll:
I think you don't need to add '0 = half full' in the y-axis label, as this is something you'd add to the cover page to the clock report (#603).

lmbollen (approved):

PR #620, opened 2024-08-27
Add scripts for setting up hoeve

lmbollen:
To make the setup process reproducible

For ethernet connectivity we need to do a few things:
* Connect all ethernet interfaces to a virtual bridge
* Assign the virtual bridge a static IP
* Set up DHCP for the ethernet interfaces


leonschoorl (commented):
How do you see this script?

#### A:
Is it something you intend to run on each CI run within the (docker) container?
Then we should consider adding `isc-dhcp-server` to the Docker image, instead of installing it each time.
Also do you even have access to the outside network interfaces to setup a bridge like this?
A vanilla docker container definitely does not. But maybe the ones setup by the CI infra are configured differently.


#### B:
Or is it something run once on a system to set it up for talking ethernet with the FPGAs?
Then it should have a more persistent way of making these changes, possibly with a NetworkManager configuration or with some systemd service or something.

hitl-setup/dhcpd.conf
```diff
@@ -0,0 +1,11 @@
...
+#
+# SPDX-License-Identifier: CC0-1.0
+
+default-lease-time 600;
+max-lease-time 7200;
+ddns-update-style none;
+subnet 100.100.100.0 netmask 255.255.255.0 {
+  range 100.100.100.2 100.100.100.254;
+  option routers 100.100.100.1;
+}
```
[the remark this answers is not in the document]
DigitalBrains1: Also, is the `routers` necessary? Isn't it just about communication on a subnet?

lmbollen:
> How do you see this script?
> #### A:
> 
> Is it something you intend to run on each CI run within the (docker) container? Then we should consider adding `isc-dhcp-server` to the Docker image, instead of installing it each time. Also do you even have access to the outside network interfaces to setup a bridge like this? A vanilla docker container definitely does not. But maybe the ones setup by the CI infra are configured differently.
> #### B:
> 
> Or is it something run once on a system to set it up for talking ethernet with the FPGAs? Then it should have a more persistent way of making these changes, possibly with a NetworkManager configuration or with some systemd service or something.

The latter. I'm not very familiar with these types of configurations, so this is currently my best effort (within reasonable time)

hitl-setup/dhcpd.conf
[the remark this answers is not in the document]
leonschoorl: A router shouldn't be necessary

hitl-setup/fpganet.service
```diff
@@ -0,0 +1,27 @@
...
+ExecStart=/usr/sbin/ip link set dev $BRIDGE_NAME up
+ExecStart=/usr/sbin/ip address add $INTERFACE_IP dev $BRIDGE_NAME
+ExecStart=/usr/bin/sh -c 'for INTERFACE in $INTERFACES; do /usr/sbin/ip link set dev $INTERFACE up; done'
+ExecStart=/usr/bin/sh -c 'for INTERFACE in $INTERFACES; do /usr/sbin/ip link set dev $INTERFACE master $BRIDGE_NAME; done'
+
+ExecStop=/usr/sbin/ip link delete name $BRIDGE_NAME type bridge
+
+[Install]
+WantedBy=multi-user.target
+WantedBy=network-online.target
```
martijnbastiaan, line 27:
```suggestion
WantedBy=multi-user.target
WantedBy=network-online.target
WantedBy=$whatever-the-github-actions-runner-is
```

?

hitl-setup/fpganet.service
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: 2024 Google LLC
+#
+# SPDX-License-Identifier: CC0-1.0
+
+[Unit]
+Description=Set up bridge network with FPGAs
```
martijnbastiaan, line 6:
```suggestion
Description=Set up bridge network with FPGAs - used for .......
```

hitl-setup/fpganet.service
DigitalBrains1: Do you want to prevent the runner from running if this script failed? Because `Wanted` is not strong enough for that.

If it's just about ordering, that's already taken care of by `Before=network.target`.

martijnbastiaan: > Do you want to prevent the runner from running if this script failed?

Yes.

How would you specify this?

DigitalBrains1: I would add this to the runner:
```
[Unit]
Requires=fpganet.service
After=fpganet.service
```

but if you want the runner to be able to run if you `systemctl disable fpganet.service`, you could add this to `fpganet.service`:
```
[Unit]
Before=fpganet.service

[Install]
RequiredBy=actions.runner.bittide-bittide-hardware.hoeve-1.service
```

But this seems weird. Why would it suddenly no longer require the service just because you `systemctl disable`d it? On the other hand, if you do that, you might have good reason to? I dunno :-).

hitl-setup/setup.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: 2024 Google LLC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FORCE="$1"
+set -exuo pipefail
```
DigitalBrains1, line 7: This is a script run by a user. It is my strong belief that `set -e` only really has a place in either simple commands (and no invocations of shell functions) or in CI scripts. Everything else is asking for trouble.

I can rewrite it to not use `set -e` unwisely.

DigitalBrains1: Although frankly this example is also a reason why we maybe shouldn't even do it in CI. In CI, we think, "well if it exits with non-zero exit code when it should have run fine that's no big problem, we'll just fix the script there and then". But it can also happen the other way around: you think it will exit but it continues running.

```bash
set -e
f() { false; echo "This won't run, right?"; }
if f; then  # same with: f && echo survived
    echo survived
fi
```
prints
```
This won't run, right?
survived
```

hitl-setup/fpganet.service
leonschoorl: I was under the (possibly naive) impression that just doing:
```
systemctl restart fpganet.service
systemctl enable fpganet.service
[...]
systemctl add-wants isc-dhcp-server.service fpganet.service
systemctl restart isc-dhcp-server.service
systemctl enable isc-dhcp-server.service
```

would be good enough to make sure that first the `fpganet.service` gets started, and then `isc-dhcp-server.service`, somewhere during boot up of the machine.

If you want to do it "really properly" you should probably add a dependency between the runner and the dhcp server.
Just depending on the setting up of the bridge isn't strictly enough.

hitl-setup/setup.sh
leonschoorl: If you think is should be rewritten, please do so.
I did the `set -e` mostly because I didn't want to put ` || exit 1` behind all the calls to systemctl, and I think it should work ok for that as is.

There are no functions, and the `for INTERFACE` and `if` block end with explicit `exit` statement.
Which I assumed would be ok, but I definitely don't know all the intricacies involved.

hitl-setup/fpganet.service
DigitalBrains1: I don't really understand what you mean, but I don't see any ordering in the commands you give. Those commands will certainly not start the `fpganet.service` before `isc-dhcp-server.service`.

> Note that requirement dependencies do not influence the order in which services are started or stopped. This has to be configured independently with the `After=` or `Before=` options. If unit foo.service pulls in unit bar.service as configured with `Wants=` and no ordering is configured with `After=` or `Before=`, then both units will be started simultaneously and without any delay between them if foo.service is activated.

DigitalBrains1: > If you want to do it "really properly" you should probably add a dependency between the runner and the dhcp server.

Doing that (if I understand what you mean correctly) means that during bootup, it will wait for the DHCP server to be started before launching the runners. But the DHCP server doesn't need to be available until well into a CI run, not for the runner to be able to start. So I think this is overkill and not really the systemd "we want boot to be quick and ordering dependencies to be few" philosophy. I think we needed `fpganet.service` to be well and truly before the runner because otherwise Docker would not be able to bind the bridge interface into its runtime; in other words, the runner truly does depend on `fpganet.service` to be available when it is starting.

But I may be misinterpreting, there are so many ways to interpret "add a dependency" in systemd land.

hitl-setup/setup.sh
```diff
@@ -0,0 +1,47 @@
...
+
+systemctl restart fpganet.service
+systemctl enable fpganet.service
+
+cp "$HERE/dhcpd.conf" /etc/dhcp/
+echo "INTERFACESv4=\"$BRIDGE_NAME\"" > /etc/default/isc-dhcp-server
+
+systemctl add-wants isc-dhcp-server.service fpganet.service
+systemctl restart isc-dhcp-server.service
+systemctl enable isc-dhcp-server.service
```
DigitalBrains1, line 47: This contains some redundancies:
* `systemctl enable fpganet.service`:
    Start `fpganet.service` on every boot
* `systemctl add-wants isc-dhcp-server.service fpganet.service`:
    If `isc-dhcp-serve.service` starts, also start `fpganet.service`
    `systemctl enable isc-dhcp-server.service`:
    Start `isc-dhcp-server.service` on every boot

Either point makes the other redundant in regard to the starting of `fpganet.service`. I'll delete the add-wants.

hitl-setup/fpganet.service
DigitalBrains1: >  I think we needed fpganet.service to be well and truly before the runner because otherwise Docker would not be able to bind the bridge interface into its runtime;

This was an earlier experiment, for the current version, the runner can start fine without the bridge. Obviously CI jobs that need the bridge will fail without it, but the runner starts fine and other jobs run fine.

martijnbastiaan (approved):

PR #621, opened 2024-08-28
Make wishbone addressing word based

lmbollen:
Byte based addressing has up to now only added complexity for both designs and testing


martijnbastiaan (approved):
LGTM, is there anything that needs to be done here still?

PR #631, opened 2024-09-03
 Align value of `k_p` with Julia implementation of Callisto 

martijnbastiaan:
If we want to fully match the Julia code we should build a clock config for 500 PPB. **Edit**: Added in https://github.com/bittide/bittide-hardware/pull/633. Switching to 500 PPB makes clock synchronize very fast though, making for uninteresting plots.

See `freq_step` at https://github.com/bittide/Callisto.jl/blob/e47139fca128995e2e64b2be935ad588f6d4f9fb/demo/pulsecontrol.jl#L41

# TODO
- [x] Wait for https://github.com/bittide/bittide-hardware/actions/runs/10690296339. View report by clicking on "Generate clock control report" and downloading the report uploaded in step "Upload reports".

### DIFF
--- a/bittide/src/Bittide/ClockControl/Callisto.hs
+++ b/bittide/src/Bittide/ClockControl/Callisto.hs
@@ -163,12 +163,16 @@ callisto ControlConfig{..} mask scs dataCounts state =
   steadyStateTarget = D.fromSignal (_steadyStateTarget <$> state)
 
   -- see clock control algorithm simulation here:
+  --
   -- https://github.com/bittide/Callisto.jl/blob/e47139fca128995e2e64b2be935ad588f6d4f9fb/demo/pulsecontrol.jl#L24
   --
-  -- the constants here are chosen to match the above code.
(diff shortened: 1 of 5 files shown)


bittide/src/Bittide/ClockControl/Callisto.hs
```diff
@@ -167,8 +167,8 @@ callisto ControlConfig{..} mask scs dataCounts state =
   --
   -- the constants here are chosen to match the above code.
```
christiaanb, line 168: We should update this comment that we copied the proportional gain _k_ from the Julia code, but that we set the step size to match the step size for which the PLL is configured. Perhaps commending on why we kick 100e-9 instead of 5e-7; I assume this might have something to do with the PLLs in the CDRs on the receive side losing track if we were to make the step size that large?

christiaanb (approved):

bittide/src/Bittide/ClockControl/Callisto.hs
martijnbastiaan: Agreed

PR #642, opened 2024-09-11
FincFdec test: change the extra clock to use out1 instead of out0

leonschoorl:
We want to use this extra clock for other things in other tests.
Namely as a better reference clock in the HwCcTopologies test.
But out0 and out0a share their divider and that won't work there.

I've made this change to the hardware setup already, so without this PR CI should now fail the FincFdecTest.


leonschoorl:
This patch only affects the fincfdec test and that test is only tested in the [nightly CI run](https://github.com/bittide/bittide-hardware/actions/runs/10818105132), not on PRs.
This patch was tested in (this CI run](https://github.com/bittide/bittide-hardware/actions/runs/10814931948).

martijnbastiaan (approved):

PR #643, opened 2024-09-24
Switch from the 300MHz onboard sysclk to the 125MHz one

leonschoorl:
As discovered by @martijnbastiaan the period of the 300MHz can't be represented by clash, causing rounding errors.
When converted back into a frequency to as input to the clock wizard, it gets rendered as a `300.030` MHz.
Then from that the clock wizard is asked to create an 125MHz clock, but can't do that exactly, creating 124.991 MHz instead.

These together account for 172ppm out of ~180ppm of difference we used to measure between the onboard clocks and the external clocks.

By using instead using the onboard 125MHz clock, which can be represented exactly, we don't have these issues.
And the both come from the same clock generator and crystal so they should have the same accuracy.

### DIFF
--- a/bittide-instances/data/constraints/fullMeshHwCcTest.xdc
+++ b/bittide-instances/data/constraints/fullMeshHwCcTest.xdc
@@ -2,8 +2,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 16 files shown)


martijnbastiaan (approved):
Yessssss, I didn't fully realize yet that this was purely a Clash issue (i.e., the PLL can do 300 => 125 exactly). LGTM.

PR #578, opened 2024-07-16
Add smoltcp hitl test

lmbollen:
Creates a HITL test containing Vexrisc with ethernet connectivity.

This test simply programs the core with a tcp echo server and tries to connect to it using `tcpspray`

PR #647, opened 2024-10-07
Identify all serial devices

lmbollen:
Before this pull request we had no idea which serial port connects to which FPGA.

This PR introduces:
- A listing of each FPGA ID and the static path to its serial port (This stays consistent as long as the setup does not change)
- A listing of each FPGA ID with its corresponding DNA. (Verified by hand, there is no check for this and we will not introduce it in this PR)
- A Hardware In The Loop test that verifies that we receive the correct DNA over from the serial ports. 

If one of the serial ports stop working or is connected to the wrong FPGA, we have a way to verify.

### DIFF
--- a/bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
+++ b/bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
@@ -164,9 +164,9 @@ runTcpTest (sock, sockAddr) = do
       <> " sent "
       <> show size
       <> " bytes in "
-      <> show diff
+      <> show (round diff :: Integer)
       <> " seconds ("
-      <> show speed
+      <> show (round speed :: Integer)
       <> " bytes/s)"
   assertEqual
     "Expected and actual bytestring are not equal"
--- a/bittide-instances/src/Bittide/Instances/Hitl/DnaOverSerial.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/DnaOverSerial.hs
@@ -1,0 +1,103 @@
+-- SPDX-FileCopyrightText: 2024 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+{-# HLINT ignore "Use head" #-}
+
+module Bittide.Instances.Hitl.DnaOverSerial where
+
+import Bittide.Hitl
+import Bittide.Instances.Domains
+import Bittide.Instances.Hitl.Setup
+import Clash.Annotations.TH (makeTopEntity)
+import Clash.Cores.UART
+import Clash.Cores.Xilinx.Extra (ibufds)
+import Clash.Cores.Xilinx.Unisim.DnaPortE2
+import Clash.Explicit.Reset
+import Clash.Functor.Extra
+import Clash.Prelude
+import Data.Char
+import qualified Data.List as L
+import Data.Maybe
+
+dnaOverSerial ::
+  "CLK_125MHZ" ::: DiffClock Ext125 ->
+  "USB_UART_TXD" ::: Signal Ext125 Bit ->
+  "USB_UART_RXD" ::: Signal Ext125 Bit
+dnaOverSerial diffClk serialIn = serialOut
+ where
+  clk = ibufds diffClk
+  (_rxData, serialOut, ack) = withClockResetEnable clk testRst enableGen $ uart baud serialIn txData
+  baud = SNat @9600
+
+  dna = readDnaPortE2 clk testRst enableGen simDna2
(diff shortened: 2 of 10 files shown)


bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -48,15 +67,15 @@ case_testGdbProgram :: Assertion
 case_testGdbProgram = do
```
hiddemoll, line 67: I have 2 remarks on things you didn't touch in this PR, but since you do touch the files you can fix them here. Please update the test documentation in `bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs`, which you copy-pasted in #578. Also rename the test case to `case_testTcpProgram` or something like that.

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -48,15 +67,15 @@ case_testGdbProgram :: Assertion
...
   startOpenOcdPath <- getOpenOcdStartPath
   startPicocomPath <- getPicocomStartPath
-  startTcpPrayhPath <- getTcpSprayPath
   uartDev <- getUartDev
   gdbScriptPath <- getGdbScriptPath
 
+  putStrLn "Starting TCP Server"
+  (serverSock, _) <- startServer
   putStrLn "Starting Gdb"
```
hiddemoll, line 75: Do you start GDB here, or do you start it in line 130? You now print it twice.

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -80,51 +99,78 @@ case_testGdbProgram = do
...
+    speed = fromIntegral size / diff
+  putStrLn $
+    show sockAddr
+      <> " sent "
+      <> show size
+      <> " bytes in "
+      <> show diff
+      <> " seconds ("
+      <> show speed
+      <> " bytes/s)"
```
hiddemoll, line 169: Can you ease up on the precision of `diff` and `speed`? In [this run](https://github.com/bittide/bittide-hardware/actions/runs/11290288485/job/31406920217#step:9:138) the following line is printed:
`10.0.0.2:1234 sent 20294504 bytes in -28.312381858000 seconds (-716806.664369905236 bytes/s)`

bittide-instances/src/Bittide/Instances/Hitl/DnaOverSerial.hs
```diff
@@ -0,0 +1,103 @@
...
+  go (vecOld, nElementsOld) (dataIn, ack) = ((vecNew, nElementsNew), dataOut)
+   where
+    (vecNew, nElementsNew)
+      | nElementsOld == 0 && isJust dataIn = (bitCoerce $ fromJust dataIn, maxBound)
+      | ack = (vecOld <<+ 0, satPred SatZero nElementsOld)
+      | otherwise = (vecOld, nElementsOld)
+
+    dataOut
+      | nElementsOld == 0 = Nothing
+      | otherwise = Just $ vecOld !! (0 :: Int)
```
hiddemoll, line 85:
```suggestion
      | otherwise = Just $ head vecOld
```

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
```diff
@@ -80,51 +99,78 @@ case_testGdbProgram = do
...
-          putStrLn stdOutContents
-
-          if not (null stdErrContents)
-            then assertFailure $ "tcpspray failed with stderr:\n" <> stdErrContents
-            else pure ()
+          putStrLn "Waiting for \"Starting TCP Client\""
+          tryWithTimeout "Handshake softcore" 10_000_000 $
+            waitForLine picocomStdOutHandle "Starting TCP Client"
+
+          let numberOfClients = 1
```
hiddemoll, line 138: Have you also tested this with more clients?

hiddemoll (changes requested):

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
lmbollen: Not without introducing a new package, which seems a bit overkill tbh

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
lmbollen: Fixed in #613

bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
lmbollen: Fixed in #613

hiddemoll (approved):
Nice work!

PR #650, opened 2024-10-14
Add memory-mapped component for UGN/IGN calculation

hiddemoll:
I'd like to start a discussion on the naming of components and signals related to UGN/IGN calculation. The 3 names of which I'm not sure if they could be improved are:
- Local sequence counter, or local counter for short. @leonschoorl  called this `txCounters` in `FullMeshSwCc.hs`
- Remote sequence counter, or remote counter for short. @leonschoorl  called this `rxCounters` in `FullMeshSwCc.hs`
- Circuit `captureUgn`, which actually does not capture the UGN but rather captures the local and remote sequence counters.


lmbollen:
>I'd like to start a discussion on the naming of components and signals related to UGN/IGN calculation
We should definitely standardize our naming conventions, I'd like to coordinate that with Martijn since I think it should be part of our new design documentation ambitions

bittide/src/Bittide/CaptureUgn.hs
```diff
@@ -0,0 +1,52 @@
...
+import Protocols.Wishbone
+
+import Bittide.SharedTypes (Bytes)
+import Bittide.Wishbone (wbToVec)
+
+import qualified Clash.Prelude as C
+
+{- | Capture the UGN from the received data stream. The first `Just` value in the
+elastic buffer contains the rxCounter. Subtracting that from the txCounter results
+in the UGN. All subsequent values in the elastic buffer are bittide data.
```
lmbollen, line 22: Change to something along the lines of:

```
Captures the remote counter from a bittide link and pairs it with the corresponding local counter.
All frames except for the first frame will be forwarded to the output.

Assumes that:
* When a link is not up, it produces `Nothing` consistently
* When a link comes down, it will at some point switch from only producing `Nothing`, to only producing `Just data`
* When a link comes up, the first frame will contain the remote counter value.
* When a link goes down, it will start producing consistently `Nothing`
```

firmware-support/bittide-sys/src/ugn.rs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use ufmt::derive::uDebug;
+
+#[repr(C)]
+#[derive(uDebug, PartialEq, Eq, Copy, Clone)]
+pub struct Ugn {
+    pub tx_counter: u64,
+    pub rx_counter: u64,
+}
```
lmbollen, line 12: Can you add some utility functions that:
* Detect if we have captured the UGNs (Both counters are 0 => Not captured). This could also be used to see if the link is up.
* Subtract the UGNs correctly to calculate the IGN.


Furthermore, change names to `local` and `remote` to be consistent with the Clash code

bittide/src/Bittide/CaptureUgn.hs
```diff
@@ -0,0 +1,52 @@
...
+  ) =>
+  Signal dom (Unsigned 64) ->
+  Circuit
+    ( Wishbone dom 'Standard addrW (Bytes 4)
+    , CSignal dom (Maybe (BitVector 64))
+    )
+    (CSignal dom (Maybe (BitVector 64)))
+captureUgn cntr = Circuit go
+ where
+  go ((wbM2S, eb), _) = ((wbS2M, pure ()), bittideData)
```
lmbollen, line 42: Change `cntr` to `localCounter` and `eb` to `linkIn`. The name elastic buffer is not inherently related to ugns

bittide/src/Bittide/CaptureUgn.hs
```diff
@@ -0,0 +1,52 @@
...
+  forall dom addrW.
+  ( C.HiddenClockResetEnable dom
+  , KnownNat addrW
+  ) =>
+  Signal dom (Unsigned 64) ->
+  Circuit
+    ( Wishbone dom 'Standard addrW (Bytes 4)
+    , CSignal dom (Maybe (BitVector 64))
+    )
+    (CSignal dom (Maybe (BitVector 64)))
```
lmbollen, line 39: I actually expect that we can (and should) remove this `Maybe` since all communication within the network is scheduled.
You can simply apply `fromMaybe 0` to the output

bittide-instances/tests/Wishbone/CaptureUgn.hs
```diff
@@ -0,0 +1,146 @@
...
+    PeConfig
+      (0b00 :> 0b01 :> 0b10 :> 0b11 :> Nil)
+      (Reloadable $ Blob iMem)
+      (Reloadable $ Blob dMem)
+
+getSequenceCounters ::
+  Signal dom (Unsigned 64, Maybe (BitVector 64)) ->
+  (Unsigned 64, BitVector 64)
+getSequenceCounters ((a, Just b) :- _) = (a, b)
+getSequenceCounters ((_, Nothing) :- xs) = getSequenceCounters xs
```
lmbollen, line 140: Can you add a small comment that explains what this function does

bittide-instances/tests/Wishbone/CaptureUgn.hs
```diff
@@ -0,0 +1,146 @@
...
+  -}
+  localCounter = register clk rst ena 0xaabbccddeeff1234 (localCounter + 1)
+
+  eb = regEn clk rst ena Nothing remoteStarted (Just . pack <$> remoteCounter)
+   where
+    remoteCounter :: Signal Basic50 (Unsigned 64)
+    remoteCounter = register clk rst ena 0x1122334411223344 (remoteCounter - 1)
+    remoteStarted = counter .==. pure maxBound
+    counter :: Signal Basic50 (Index 1000)
+    counter = register clk rst ena 0 (satSucc SatBound <$> counter)
```
lmbollen, line 82: You can move the `Index 1000` annotation to the initial value and remove the `counter` type annotation.

bittide-instances/tests/Wishbone/CaptureUgn.hs
```diff
@@ -0,0 +1,146 @@
...
+
+  {- The local counter starts counting up from 0x1122334411223344. The elastic buffer
+  outputs Nothing for 1000 cycles, after which it will  start outputting a decreasing
+  counter starting at 0xaabbccddeeff1234.
+  -}
+  localCounter = register clk rst ena 0xaabbccddeeff1234 (localCounter + 1)
+
+  eb = regEn clk rst ena Nothing remoteStarted (Just . pack <$> remoteCounter)
+   where
+    remoteCounter :: Signal Basic50 (Unsigned 64)
```
lmbollen, line 78: You should not need this type annotation and I think it's trivial

bittide-instances/tests/Wishbone/CaptureUgn.hs
```diff
@@ -0,0 +1,146 @@
...
+  Signal dom (Maybe (BitVector 64)) ->
+  -- | Local sequence counter
+  Signal dom (Unsigned 64) ->
+  -- | USB_UART_TX
+  Signal dom Bit ->
+  -- | USB_UART_RX
+  Signal dom Bit
+dut baud clk rst eb localCounter usbUartTx = usbUartRx
+ where
+  (_, usbUartRx) = go ((usbUartTx, pure $ JtagIn low low low), pure ())
```
lmbollen, line 104: Could you try removing the `Jtag` circuit and use `jtag <- idleSource -< ()` instead in the `circuit` itself?

lmbollen (approved):
lgtm modulo some minor comments

bittide-instances/tests/Wishbone/CaptureUgn.hs
hiddemoll: We do need the type annotation, but I've moved it to the initial value.

Issue #587, opened 2024-07-31
Implement fast/efficient execution deadlines in software


martijnbastiaan:
We currently maintain two instances for clock control: hardware and software. We use the former because it both consistently and reliably hits a (maximum) clock control update frequency of 1 MHz. While the latter does what we want in practice (clock convergence), we don't have a good idea on how long it takes to do a single clock control update. I.e.:

![Untitled Diagram drawio(1)](https://github.com/user-attachments/assets/7ec6b454-c28f-47f8-b5f8-24984608c0c9)

We don't really mind fewer updates (say, 100 KHz) but the updates should be timed reliably to make sure we can make our theory match our practice. We currently do have some timing logic implemented in Rust (`firmware-support/bittide-sys/src/time.rs`) backed by a hardware component `bittide/src/Bittide/Wishbone.hs` => `timeWb` but this is too slow to reliably schedule microseconds.

We should investigate what is needed to set a reliable update rate. There is no one valid solution that comes to mind, but in the end we should have a solution that:

* Can reliably hit an update mark precise to say 100 KHz.
* Gives feedback in such a way that it is immediately noticeable if the mechanism fails / the computation between updates takes too long

Relevant code:

* `bittide/src/Bittide/Wishbone.hs` => `timeWb`
* `firmware-support/bittide-sys/src/time.rs`: HAL for timing component
* `firmware-support/bittide-sys/src/clock_control.rs`: HAL for FINC/FDEC buttons
* `firmware-support/bittide-sys/src/callisto.rs`: Implementation of clock control algorithm ("Callisto")

leonschoorl:
I added some [measurement hardware](https://github.com/bittide/bittide-hardware/commit/9927aaebcbf373c78094679b1083d11883415ffb) and ran [a test](https://github.com/bittide/bittide-hardware/actions/runs/10600359274/job/29416763045).

With the current setup the period between clock updates fluctuates between 405 and 414 cycles @ 125MHz.
Except for some reason on node 2, where it fluctuates between 403 and 412 cycles.
So the update frequency is `125 MHz / 414 = 0.301 MHz` or 301 kHz.
And the period is `414 / 125 MHz = 3.31e-6 s` or 3.31 μs

rslawson:
I didn't realise that this would get closed when even just one of the linked PRs got closed, oops.

rslawson:
Oops, still not done with #654.

PR #654, opened 2024-10-15
Add in a software clock control DUT

rslawson:
This PR adds in a new SwCc DUT based on the work done in my two other PRs. As of the current PR it mostly passes CI except for linting, but that's okay since there's still work to be done.

As for what work is still yet to be done:
- @lmbollen suggested that I add in UGN stability testing to this test as well. It was in some earlier versions of my work, which can be seen in the (awful) history of my `rs/measure-sw-cc` branch. The only other DUT this appears in is `FullMeshSwCc`.
- Some work must be done to figure out why I need the weird hack around masking `syncIn` to `IlaPlotSetup` (see line 908 and lines 937-947 as of time of writing) in order to avoid two `syncRst` asserts, which then causes `captureCond` to become `Calibrate :: CaptureCondition` for a second time, which is problematic.
- Commented-out legacy code needs to be removed.
- Formatting needs to be fixed (thanks CI).


rslawson:
I should also probably report that somewhere in the combined total of the work from the three PRs (between this one and the two it depends on) the clock update frequency variability mentioned in issue #587 seems to have disappeared. Looking at the `callistoSwIla` ILA data, every test reports a minimum update period of 307 cycles/2.456 μs and a maximum update period of 309 cycles/2.472 μs for a variability of 16 ns. This is consistent across _all_ boards within _all_ topologies. The only big unknown left is software reframing timing, but that should be easy enough to get data on with just a single change to the DUT in another branch.

rslawson:
As per the discussion in the meeting earlier today, the `syncIn` masking hack will stay in place until the testing framework changes in the `pre-process` branch are merged.

The issue is, as far as I can tell, that the `syncOut` signal from FPGA 7 continues to propagate to the other boards until it is reset by the HITL VIO providing it with new test data. The other boards, are still receiving this on `syncIn`, causing them to begin the test early. This results in the plotting ILA entering the calibrating state twice per test, which is a problem for `cc-plot`.

The current ~~hack~~ solution in this branch looks for the `syncIn` signal to be unchanging for 25ms or longer, which is interpreted as FPGA 7 being reset. After this the `syncIn` signal is unmasked and allowed to propagate through to the test synchronisation circuitry for the duration of the test. However, through proper usage of the pre-test, monitor, and post-test functions in the HITL tests it should be possible to force all of the synchronisers into reset in the post-test phase. This would ensure there would be no sync signal on `syncIn` until all of the boards have been given test data in the following pre-test phase.

rslawson:
Also yes, I promise I'll clean up this commit history before asking for review.

rslawson:
As mentioned, reframing is currently not working and I don't believe there's a quick path to getting it to work. As such, I'll make another PR at a later time to fix that, but the testing framework for enabling it will remain in place for the time being.

lmbollen (changes requested):

bittide/src/Bittide/ClockControl/CallistoSw.hs
```diff
@@ -0,0 +1,149 @@
...
+import Bittide.SharedTypes (ByteOrder (BigEndian))
+
+import Project.FilePath
+
+import Protocols
+import Protocols.Idle
+
+import Clash.Cores.Xilinx.Ila (Depth (..), IlaConfig (..), ila, ilaConfig)
+import Data.Maybe (fromMaybe, isJust)
+import VexRiscv
```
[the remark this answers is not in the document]
martijnbastiaan: "Qualified deps" => we can do this with the latest fourmolu https://github.com/fourmolu/fourmolu/pull/403. _But_, these changes haven't made it into a release yet. Something to keep in mind.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
```diff
@@ -196,6 +183,8 @@ data TestConfig = TestConfig
   -- cycles of the stable clock.
   , mask :: BitVector LinkCount
   -- ^ The link mask depending on the selected topology.
+  , reframingEnabled :: Bool
+  -- ^ Whether or not to run this test with reframing enabled.
```
martijnbastiaan, line 187: Create an issue we need to figure out how to properly test reframing and link it here.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
```diff
@@ -570,154 +477,217 @@ topologyTest refClk sysClk sysRst IlaControl{syncRst = rst, ..} rxNs rxPs miso c
...
+        rxCounterStartUgn
+        enaSig
+        . xpmCdcArraySingle txClk sysClk
+
+  ugnsStable :: Vec LinkCount (Signal Basic125 Bool)
+  ugnsStable = go <$> ugns'
+   where
+    go ugn = ugnStable
+     where
+      ugn' = bitCoerce <$> ugn
```
martijnbastiaan, line 595:
```suggestion
      ugn1 = bitCoerce <$> ugn
```

Use number postfixes please. Ticks don't really scale and are easy to miss.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
```diff
@@ -570,154 +477,217 @@ topologyTest refClk sysClk sysRst IlaControl{syncRst = rst, ..} rxNs rxPs miso c
...
+    , dDiff2
+    , dDiff3
+    , dDiff4
+    , dDiff5
+    , dDiff6
+    ) = vecToTuple domainDiffs
+
+type WaitCycles dom hold prd = PeriodToCycles dom (prd - hold)
+
+data ToFincFdecState' dom hold min
```
martijnbastiaan, line 659: It isn't clear what the `'` is supposed to communicate.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
```diff
@@ -570,154 +477,217 @@ topologyTest refClk sysClk sysRst IlaControl{syncRst = rst, ..} rxNs rxPs miso c
...
+  = Wait (Index (WaitCycles dom hold min))
+  | Pulse (Index (PeriodToCycles dom hold)) SpeedChange
+  | Idle
+  deriving (Generic, NFDataX, Eq)
+
+speedChangeToFincFdec' ::
```
martijnbastiaan, line 665: Same here. Typically `'` is used for "strict" versions of a function, but that's not the case here.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
```diff
@@ -738,19 +708,53 @@ hwCcTopologyTest ::
...
+  syncNodeEnteredReset =
+    trueFor
+      (SNat @(Milliseconds 25))
+      sysClk
+      testReset
+      (not <$> syncNodeProbablyWorking)
+  syncNodePrevEnteredReset =
+    sticky sysClk testReset syncNodeEnteredReset
+
+  syncIn' = mux syncNodePrevEnteredReset syncIn (pure True :: Signal Basic125 Bool)
```
martijnbastiaan, line 743:
```suggestion
  syncIn1 = mux syncNodePrevEnteredReset syncIn (pure True :: Signal Basic125 Bool)
```

martijnbastiaan (changes requested):
LGTM in general. I'm really looking forward to ditching all the different instances.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
```diff
@@ -439,56 +330,85 @@ topologyTest refClk sysClk sysRst IlaControl{syncRst = rst, ..} rxNs rxPs miso c
...
   fincFdecIla =
     setName @"fincFdecIla"
       ila
       ( ilaConfig
-          $ "trigger_0"
-          :> "capture_0"
-          :> "probe_milliseconds"
+          $ "trigger_fdi_0"
+          :> "capture_fdi_0"
+          :> "probe_fdi_milliseconds"
```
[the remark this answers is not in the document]
rslawson: I actually have no idea. I swear I copied that from another test but now I can't find where.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
```diff
@@ -416,6 +298,15 @@ topologyTest refClk sysClk sysRst IlaControl{syncRst = rst, ..} rxNs rxPs miso c
   transceiversFailedAfterUp =
     sticky sysClk syncRst (isFalling sysClk syncRst enableGen False allReady)
 
+  othersNotInCCReset = go <<$>> transceivers.rxDatas
+   where
+    go (Just val) = msb val == high
+    go _ = False
```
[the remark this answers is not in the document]
rslawson: I still need to do `<<$>>`, but otherwise yeah I like how you wrote that.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
```diff
@@ -0,0 +1,1057 @@
...
+  -- Check that tests are not synchronously started before all transceivers are up
+  startBeforeAllReady =
+    sticky
+      sysClk
+      syncRst
+      (syncStart .&&. ((not <$> allReady) .||. transceiversFailedAfterUp))
+
+  calibratedClockShift = capturedOnce
+   where
+    -- Possibly overkill.
```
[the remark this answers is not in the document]
rslawson: So actually I think it's worth bringing this up here since I wanted to ask if there was a good way to latch a value exactly once on a given signal. I think what I wrote there might be a bit over-engineered, and I'm also not sure if there's just a utility function somewhere that I overlooked.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
rslawson: Made #693, will link in the comment.

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
rslawson: It's there because I forgot you can `import Foo hiding (bar)`. Will fix.

bittide/src/Bittide/ClockControl/CallistoSw.hs
```diff
@@ -0,0 +1,163 @@
...
+          elfPath = elfDir </> "clock-control"
+          iSize = 64 * 1024 -- 64 KB
+          dSize = 64 * 1024 -- 64 KB
+        memBlobsFromElf BigEndian (Just iSize, Just dSize) elfPath Nothing
+     )
+  peConfig =
+    PeConfig
+      (0b100 :> 0b010 :> 0b110 :> 0b111 :> 0b001 :> Nil)
+      (Reloadable $ Blob iMem)
+      (Reloadable $ Blob dMem)
```
lmbollen, line 163: Since there is only one clock domain here, can we try to get away with using `exposeClockResetEnable` combined with `hasClock`, `hasReset`, `hasEnable`? ping me if it's not trivial for you to do :)

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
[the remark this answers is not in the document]
lmbollen: If there's a reason to not name them `trigger` and `capture`, could you add it in a comment?

bittide-instances/src/Bittide/Instances/Hitl/SwCcTopologies.hs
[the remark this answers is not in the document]
lmbollen: How about:
```hs
capturedOnce = 
  regEn clk rst ena Nothing 
  (isNothing <$> capturedOnce) 
  (orNothing <$> captureCond <*> capturedValue )
```
This samples `capturedValue` once, you can then use `fromMaybe defaultValue` to output `defaultValue` to set it to `defaultValue` untill you capture.

You then get
```hs
capturedOnce = fromMaybe defaultValue <$>
  regEn clk rst ena Nothing 
  (isNothing <$> capturedOnce) 
  (orNothing <$> captureCond <*> capturedValue )
  ```

lmbollen (changes requested):
Some small things ...
Could you resolve the conversations whose changed you adapter?

lmbollen (approved):
Ryan is going to apply the small remaining changes and is then free to merge when CI passes

PR #669, opened 2024-11-12
Tee Picocom stdout and stderr to files and archive them

rslawson:
CI seems to fail spuriously on timeouts, almost exclusively on the Vex RISC-V TCP test and in invoking Picocom. However, when certain functions time out, their previously read input is discarded, often along with what's left in the buffer being read from. These should instead probably be printed out on the program's stdout so that CI logs show what occurred prior to the timeout.

### DIFF
--- a/bittide-instances/data/picocom/start.sh
+++ b/bittide-instances/data/picocom/start.sh
@@ -8,4 +8,18 @@
 #
 #    /dev/serial/by-id/usb-FTDI_Dual_RS232-if01-port0
 #
-picocom  --baud 921600 --imap lfcrlf --omap lfcrlf $@
+set -e
+
+# Default stdout to /dev/null
+PICOCOM_STDOUT_LOG="${PICOCOM_STDOUT_LOG:-/dev/null}"
+stdout_dir=$(dirname "${PICOCOM_STDOUT_LOG}")
+mkdir -p "${stdout_dir}"
+
+# Default stderr to /dev/null
+PICOCOM_STDERR_LOG="${PICOCOM_STDERR_LOG:-/dev/null}"
+stderr_dir=$(dirname "${PICOCOM_STDERR_LOG}")
+mkdir -p "${stderr_dir}"
+
+picocom  --baud 921600 --imap lfcrlf --omap lfcrlf $@ \
+  > >(tee "${PICOCOM_STDOUT_LOG}") \
+  2> >(tee "${PICOCOM_STDERR_LOG}" >&2)
--- a/bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
+++ b/bittide-instances/exe/post-vex-riscv-tcp-test/Main.hs
@@ -79,6 +79,13 @@ case_testTcpClient = do
   (serverSock, _) <- startServer
   withAnnotatedGdbScriptPath gdbScriptPath $ \gdbProgPath -> do
     currentEnv <- getEnvironment
+    projectDir <- findParentContaining "cabal.project"
+    let
+      hitlDir = projectDir </> "_build" </> "hitl"
+      stdoutLog = hitlDir </> "picocom-stdout.log"
+      stderrLog = hitlDir </> "picocom-stderr.log"
+    putStrLn $ "logging stdout to `" <> stdoutLog <> "`"
+    putStrLn $ "logging stderr to `" <> stderrLog <> "`"
     let
       openOcdProc =
         (proc startOpenOcdPath [])
@@ -95,6 +102,9 @@ case_testTcpClient = do
           { std_out = CreatePipe
           , std_in = CreatePipe
           , new_session = True
+          , env =
+              Just
+                (currentEnv <> [("PICOCOM_STDOUT_LOG", stdoutLog), ("PICOCOM_STDERR_LOG", stderrLog)])
           }
 
     putStrLn "Starting OpenOcd..."


bittide-instances/src/Project/Handle.hs
```diff
@@ -37,15 +40,45 @@ expectLine = expectLine' ""
...
+            Stop (Error msg') -> do
+              rest <- readRemainingChars h'
+              putStrLn $ "handle buffer dump:\n" <> rest
+              errorIO $
+                "expect line failed on `"
+                  <> msg
+                  <> "`\nmessage: `"
+                  <> msg'
+                  <> "`\npreviously read:\n"
+                  <> s1
```
lmbollen, line 72: Would it make sense to create a custom return type to differentiate between `Success`, `Timeout` and `Closed`?

bittide-instances/src/Project/Handle.hs
```diff
@@ -37,15 +40,45 @@ expectLine = expectLine' ""
...
+  timeout' = toRational timeout
+  expectLine' t0 s0 h' f' = do
+    now <- getCurrentTime
+    if toRational (diffUTCTime t0 now) >= timeout'
+      then errorIO $ "expect line timed out on: `" <> msg <> "`. previously read: \n" <> s0
+      else do
+        line <- hGetLine h
+        let
+          trimmed = trim line
+          s1 = s0 <> "\n" <> line
```
lmbollen, line 56: I think I'd rather see a list of lines  of type `[String]` rather than reading on a per line basis and then manually re-inserting new lines. You can easily convert `[String]` to `String` using `unlines`.

bittide-instances/src/Project/Handle.hs
```diff
@@ -37,15 +40,45 @@ expectLine = expectLine' ""
...
+          cont = expectLine' t0 s1 h' f'
+        if null trimmed
+          then cont
+          else case f trimmed of
+            Continue -> cont
+            Stop Ok -> pure ()
+            Stop (Error msg') -> do
+              rest <- readRemainingChars h'
+              putStrLn $ "handle buffer dump:\n" <> rest
+              errorIO $
```
lmbollen, line 66: You seem to trade `assertFailure` for `errorIO`, is there a reason for that?

bittide-instances/exe/post-vex-riscv-test/Main.hs
```diff
@@ -64,59 +62,34 @@ case_testGdbProgram = do
...
-          loggingSequence = do
-            threadDelay 1_000_000 -- Wait 1 second for data loggers to catch up
-            putStrLn "Picocom stdout"
-            picocomOut <- readRemainingChars picocomStdOut
-            putStrLn picocomOut
-            case maybePicocomStdErr of
-              Nothing -> pure ()
-              Just h -> do
-                putStrLn "Picocom StdErr"
-                readRemainingChars h >>= putStrLn
```
lmbollen, line 88: I see you are removing picocom stderr printing from the logging infra, do you think we don't need it?

bittide-instances/src/Project/Handle.hs
rslawson: That I'm not sure of - I'm not entirely certain whether that info is actually useful to the postproc functions since I haven't read over them properly, just updated the call locations for the functions I changed.

bittide-instances/exe/post-vex-riscv-test/Main.hs
rslawson: Mostly just wasn't sure how to integrate it into the new stuff. I would probably have to include another argument to the function, something like `onErr` which would run in the error branches. Not sure if that would be overcomplicating things, though.

bittide-instances/data/picocom/start.sh
```diff
@@ -8,4 +8,6 @@
 #
 #    /dev/serial/by-id/usb-FTDI_Dual_RS232-if01-port0
 #
-picocom  --baud 921600 --imap lfcrlf --omap lfcrlf $@
+picocom  --baud 921600 --imap lfcrlf --omap lfcrlf $@ \
+  > >(tee -a $PICOCOM_STDOUTLOG) \
+  2> >(tee -a $PICOCOM_STDERRLOG >&2)
```
martijnbastiaan, line 13:
```suggestion
set -u

picocom  --baud 921600 --imap lfcrlf --omap lfcrlf $@ \
  > >(tee -a "${PICOCOM_STDOUTLOG}") \
  2> >(tee -a "${PICOCOM_STDERRLOG}" >&2)
```

martijnbastiaan: `set -u`: error if variables are undefined
`${}`: good practice to avoid things like `$FOO_bar`, where you _think_ you're accessing `$FOO`, but you're accessing `$FOO_bar`.
`""`: account for spaces

martijnbastiaan (approved):
Please change the PR title too :)

martijnbastiaan (changes requested):
Wait no, we need artifacts on CI too for this PR to make sense :upside_down_face:

martijnbastiaan (approved):

martijnbastiaan:
Sorry for the spam. LGTM if we get the artifacts on CI.

bittide-instances/data/picocom/start.sh
```diff
@@ -8,4 +8,34 @@
...
+  PICOCOM_STDERRLOG=/dev/null
+else
+  >&2 echo "envvar \$PICOCOM_STDERRLOG set to '$PICOCOM_STDERRLOG'."
+  outdir=$(dirname "$PICOCOM_STDERRLOG")
+  if [[ ! -d "$outdir" ]]
+  then
+    >&2 echo "directory $outdir does not exist. creating it."
+    mkdir -p "$outdir"
+  fi
+fi
```
martijnbastiaan, line 37: I think I'd keep it much simpler, maybe:

```suggestion
set -e

# Default stdout to /dev/null
PICOCOM_STDOUT_LOG="${PICOCOM_STDOUT_LOG:-/dev/null}"
stdout_dir=$(dirname "${PICOCOM_STDOUT_LOG}")
mkdir -p "${stdout_dir}"

# Default stderr to /dev/null
PICOCOM_STDERR_LOG="${PICOCOM_STDERR_LOG:-/dev/null}"
stderr_dir=$(dirname "${PICOCOM_STDERR_LOG}")
mkdir -p "${stderr_dir}"
```

I can see that there is value in warning for undefined variables, but all the other things are way too noisy IMO.

```suggestion
if [[ -z "${PICOCOM_STDOUTLOG}" ]]
then
  >&2 echo "envvar \$PICOCOM_STDOUTLOG not set. defaulting to /dev/null."
  PICOCOM_STDOUTLOG=/dev/null
fi

outdir=$(dirname "$PICOCOM_STDOUTLOG")
mkdir -p "$outdir"

if [[ -z "${PICOCOM_STDERRLOG}" ]]
then
  >&2 echo "envvar \$PICOCOM_STDERRLOG not set. defaulting to /dev/null."
  PICOCOM_STDERRLOG=/dev/null
fi

mkdir -p "$outdir"
```

In any case: 

1. `set -e` prevents silent errors, use it :). 
2. `PICOCOM_STDERRLOG` -> `PICOCOM_STDERR_LOG`
3. I think `$a` -> `${a}` is good practice.

PR #675, opened 2024-11-15
Test the GDB programming of all devices

hiddemoll:
The demo rig had a wrongly connected jumper cable, causing the issue @lmbollen decribed in #649 . This has been fixed. 

I rewrote the post-processing function for the `vexRiscvTest` so it runs on all devices. @lmbollen also requested I do that for `vexRiscvTcpTest`, but since we still have some issues with that test on a single device I think it is best to postpone extending it to all devices.

### DIFF
--- a/bittide-instances/exe/post-vex-riscv-test/Main.hs
+++ b/bittide-instances/exe/post-vex-riscv-test/Main.hs
@@ -7,6 +7,7 @@
 import Prelude
 
 import Control.Concurrent (threadDelay)
+import Control.Exception (catch)
 import Data.Maybe (fromJust)
 import System.Environment (withArgs)
 import System.IO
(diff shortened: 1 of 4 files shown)


bittide-instances/exe/post-vex-riscv-test/Main.hs
```diff
@@ -42,81 +43,97 @@ GDB: GNU Debugger. This program will connect to the OpenOCD server and is able
...
-            waitForLine picocomStdOut "Hello, UART!"
+  results <- mapM (\d -> catch (testGdb d) catchHUnitFailure) demoRigInfo
+  print results
+  assertBool "Not all FPGAs were correctly programmed" (and results)
+ where
+  -- We use this solution so the test stops when it encounters a problem
+  catchHUnitFailure :: HUnitFailure -> IO Bool
+  catchHUnitFailure (HUnitFailure _ msg) = do
+    putStrLn $ "Test failed with: " <> msg
+    pure False
```
lmbollen, line 54: Perhaps a good idea to also pass `d` to `catchHUnitFailure` and print some more info regarding the device for which the test failed.

bittide-instances/exe/post-vex-riscv-test/Main.hs
```diff
@@ -42,81 +43,97 @@ GDB: GNU Debugger. This program will connect to the OpenOCD server and is able
...
+            -- Wait for GDB to reach its last command - where it will wait indefinitely
+            tryWithTimeout "Waiting for \"> continue\"" 10_000_000 $
+              waitForLine gdbStdOut "> continue"
+
+            -- Test UART echo
+            hPutStrLn picocomStdIn "Hello, UART!"
+            tryWithTimeout "Waiting for \"Hello, UART!\"" 10_000_000 $
+              waitForLine picocomStdOut "Hello, UART!"
+
+            pure True
```
lmbollen, line 136: It seems you can just return `IO ()` and use `mapM_`

lmbollen (approved):
lgtm modulo two minor comments

martijnbastiaan (approved):
Nice work. Agreed with @lmbollen's comments.

PR #679, opened 2024-11-20
Remove `Maybe` from the `switch` links

lmbollen:
Our internal bittide links have have notion of transactions, thus a valid bit is not needed

### DIFF
--- a/bittide/src/Bittide/Node.hs
+++ b/bittide/src/Bittide/Node.hs
@@ -11,6 +11,7 @@ module Bittide.Node where
 import Clash.Prelude
 import Clash.Sized.Vector.ToTuple (vecToTuple)
 
+import Data.Maybe
 import Protocols
 import Protocols.Idle
 import Protocols.Internal (vecCircuits)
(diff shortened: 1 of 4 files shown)


bittide/src/Bittide/Node.hs
```diff
@@ -319,3 +320,17 @@ unconcatC ::
...
+ where
+  go (fwds, bwds) = unzip $ zipWith (curry f) fwds bwds
+
+-- | Map a function over a Circuit of CSignals
+cSigMap ::
+  forall dom a b.
+  (KnownDomain dom) =>
+  (a -> b) ->
+  Circuit (CSignal dom a) (CSignal dom b)
+cSigMap fn = Circuit $ \(m, _) -> (pure (), fn <$> m)
```
hiddemoll, line 336: This function is also exists in [HwCcTopologies.hs](https://github.com/bittide/bittide-hardware/blob/8f6a4ddcb0477ec7f5f33b636b4c96d82bf64af1/bittide-instances/src/Bittide/Instances/Hitl/HwCcTopologies.hs#L219). Is it worth it to deduplicate them?

hiddemoll (commented):
Nice work. Shouldn't the `DataLink` be replaced with `BitVector fw` everywhere after the `captureUgn` component? If so, it should also be removed from the scatter and gather unit. 

(Bonus points for creating a function named `funzip`, that made me smile)

bittide/src/Bittide/Node.hs
lmbollen: The reason why I did not remove `DataLink` yet entirely is because that would leave the scatter unit and gather unit with not tests at the moment. I'll get that in the next PR :)

lmbollen: Regarding the deduplication of `cSigMap`, we should add some form of mapping upstream to `clash-protocols`. 
Though I do not know yet if that should be a function per protocol or whether we can make a typeclass.
I would not worry right now about a function that's duplicated once

hiddemoll (approved):

PR #680, opened 2024-11-20
Upgrade SmolTCP

martijnbastiaan:
Trying to get rid of unsafe blocks.

For reference, I've updated all the `Cargo.lock` files (twice) by doing:

```bash
find . -name Cargo.lock -not -path "./clash-vexriscv/*" | xargs rm
./cargo.sh clean
./cargo.sh update
```

Note that upgrading smoltcp forced a Rust upgrade too, which in turn required some clippy fixes and warning suppression.

### DIFF
--- a/firmware-binaries/examples/smoltcp_client/src/main.rs
+++ b/firmware-binaries/examples/smoltcp_client/src/main.rs
@@ -42,6 +42,8 @@ const CHUNK_SIZE: usize = 4096;
 const SERVER_IP: IpAddress = IpAddress::v4(10, 0, 0, 1);
 const SERVER_PORT: u16 = 1234;
 
+// See https://github.com/bittide/bittide-hardware/issues/681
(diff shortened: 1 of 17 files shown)


firmware-binaries/examples/smoltcp_client/src/main.rs
```diff
@@ -42,6 +42,8 @@ const CHUNK_SIZE: usize = 4096;
 const SERVER_IP: IpAddress = IpAddress::v4(10, 0, 0, 1);
 const SERVER_PORT: u16 = 1234;
 
+// See https://github.com/bittide/bittide-hardware/issues/681
+#[allow(static_mut_refs)]
```
rslawson, line 46: I have Opinions™ on how this should be removed, but for now at least this can stay.

rslawson (approved):

PR #682, opened 2024-11-22
Filter NULL bytes in `expectLine`

martijnbastiaan:
Picocom sometimes produces NULL bytes. Attempts to debug this behavior have been futile, so this patch simply filters them out. This does remove the ability to send over binary data, but I don't think we want to do that over UART anyway..

Note: this removes a call to `trim`, in favor of `trimEnd`. We don't want to strip start of line whitespace..

### DIFF
--- a/bittide-instances/src/Project/Handle.hs
+++ b/bittide-instances/src/Project/Handle.hs
@@ -6,7 +6,7 @@ module Project.Handle where
 
 import Prelude
 
-import Data.List.Extra (trim)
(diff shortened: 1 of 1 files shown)


lmbollen (approved):

leonschoorl (approved):
The `filter (/= '\NUL')` doesn't break the ability to ability to send over arbitrary raw bytes anymore then `trim` and `hGetLine` already did.

PR #683, opened 2024-11-22
Use `bracket`-style socket setup

martijnbastiaan:
This makes sure resources get cleaned up in case of errors. This helps running multiple tests, which is for now mainly motivated for debugging purposes.


lmbollen (approved):
Lgtm

Issue #663, opened 2024-10-28
Update HAL for scatter and gather engines


lmbollen:
There is an existing Hardware Abstraction Layer for the scatter and gather engines, but these have not been used yet. 
The hardware abstraction layers should be updated to not use the flattened device tree initialization.

Furthermore we should create tests that check the HAL and component combination.

Note that the scatter and gather engines currently operate on `Signal dom (Maybe (BitVector fw))` which should change to `Signal dom (BitVector fw)` at some point in the future #662

PR #684, opened 2024-11-26
Remove `DataLink` from internal bittide links

lmbollen:
These links have no notion of transactions, they simply send over `BitVector`s.

For now I've removed the existing unit tests which will be replaced by integration tests as part of #663


hiddemoll (approved):
LGTM

PR #685, opened 2024-11-27
Remove `firmware-binaries/examples/.cargo/Cargo.toml`

lmbollen:
This causes rust code to trap instead of panic, leading to undebuggable errors


rslawson (commented):
Looks good to me.

rslawson (approved):

PR #687, opened 2024-11-29
Ignore all non printable UART bytes

martijnbastiaan:
Maybe we want a `waitForLineLenient` instead..


martijnbastiaan:
This is a "real" issue

PR #690, opened 2024-12-03
Attempt to capture panics with GDB

rslawson:
There's still some issues here, e.g. it seems like the maximum stack trace depth on a panic is 2 no matter what I try, but otherwise it seems to have reduced the number of times the TCP test fails (likely by replacing `start` with `continue` in the GDB script). Otherwise, this PR:

- Introduces a new crate to `firmware-support` that provides a method that GDB can `break` on (though it should preferentially break on `core::panicking::panic` or `rust_begin_unwind` for the time being) in order to grab a backtrace of the program and exit.
- Changes how the TCP test is started.
- Attempts to force Rust to build the TCP test with frame pointers preserved and unwind tables (unsure if this is actually working).


firmware-binaries/.cargo/config.toml
```diff
@@ -4,7 +4,11 @@
 
 [build]
 target = "riscv32imc-unknown-none-elf"
-rustflags = ["-C", "target-feature=+f"]
+rustflags = [
+  "-C", "target-feature=+f",
+  "-C", "force-frame-pointers=yes",
+  "-C", "force-unwind-tables=yes"
```
martijnbastiaan, line 10: Could you add an `XXX: Seem to be ignored`?

firmware-support/gdb-trace/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+
+use core::{fmt::Write, panic::PanicInfo};
+
+/// Internal function for a GDB panic handler. It disables interrupts and attemps to write
+/// the panic info to the provided write-able interface. It's expected that the user will
+/// be running a GDB process and set a breakpoint to the panic handler so that the output
+/// can be read.
```
martijnbastiaan, line 12: Will we always have GDB though? Couldn't we _also_ write to UART here?

rslawson: Right, yes, it could just do that. That's totally an option - nothing in this function _requires_ that you have GDB hook a breakpoint on this function, it's just there for convenience so that there's a fixed name to use. It takes some generic writer in order to enable that.

firmware-binaries/.cargo/config.toml
rslawson: Yep, can do.

firmware-support/gdb-trace/src/lib.rs
martijnbastiaan: Ahhh okay, I get how to read this now. So in the current configuration nothing really changes. If we _don't_ trap with GDB it'll happily print to UART, because we pass that as `$writer`?

firmware-binaries/.cargo/config.toml
martijnbastiaan: Wait the other flags _also_ seem to be ignored write? When we add gooblygook there, we don't get any warnings or errors.

firmware-support/gdb-trace/src/lib.rs
rslawson: Yeah, if you define your panic handler using the provided convenience macro. If you need something else then the `gdb_panic_internal` function is there to handle writing the panic info to whatever is provided, even if that's a null writer.

martijnbastiaan (approved):

PR #692, opened 2024-12-04
RISCy `sim` functions

lmbollen:
Debug some designs with a risc processor more easily by simulating it in your terminal


bittide-instances/src/Bittide/Instances/Pnr/Ethernet.hs
```diff
@@ -26,11 +28,34 @@ import Bittide.ProcessingElement
 import Bittide.Wishbone
 import Protocols.Idle
 
+-- Simulation imports
```
hydrolarus, line 31: Maybe it's super overkill, but this sounds like `LANGUAGE CPP` could be nice for conditional imports and also later down in the file for the memory contents?

lmbollen: I added two flags, one to include the binaries in the compilation named `cpu-include-binaries`, and `sim-baud-rate` that sets the baud rate to the maximum allowed by the domain for faster simulation.

martijnbastiaan: I know I'm on holiday :eyes:, but couldn't we use `clashSimulation` for this? CPP is somewhat terrible in the sense that it's very hard to keep all combinations of CPP compiling / testable.

lmbollen: It would not be trivial, since we would then always depend on the existence of the rust binaries that we simulate.

An alternative to this would be to take the same approach as `clash-vexriscv-sim` and use a simulated jtag bridge + gdb + openocd. That would also be closer to the truth. The downside is that that costs more engineering time than using the existing solution.

martijnbastiaan: I think it should be relatively easy:

```haskell
{- | Given the path to an elf file, return a tuple containing @(instruction memory, data memory)@. -}
vecsFromElf ::
  forall nData nInstr .
  (HasCallStack) =>
  -- | How the words should be ordered
  ByteOrder ->
  FilePath ->
  -- | Optional tuple of starting address and filepath to a device tree.
  Maybe (I.Key, FilePath) ->
  -- | (instruction memBlob, data memBlob)
  IO ( Vec nData (BitVector 32)
     , Vec nInstr (BitVector 32)
     )
vecsFromElf byteOrder elfPath maybeDeviceTree = do
  (iMemIntMap, dMemIntMap) <- getBytesMems elfPath maybeDeviceTree
  let
    dSize = natToNum @nData
    iSize = natToNum @nInstr
    (_iStartAddr, _, iList) = extractIntMapData byteOrder iMemIntMap
    (_dStartAddr, _, dList) = extractIntMapData byteOrder dMemIntMap
    iListPadded = padToSize "Instruction memory" (fmap ((`div` 4) . (+ 3)) iSize) 0 iList
    dListPadded = padToSize "Data memory" (fmap ((`div` 4) . (+ 3)) dSize) 0 dList
    iBlob = memBlobTH Nothing iListPadded
    dBlob = memBlobTH Nothing dListPadded

  pure (fromList iListPadded, fromList dListPadded)
```

Then use that function:

```haskell
  peConfig = PeConfig memMap (Vec iMem) (Vec dMem)

  (iMem, dMem) =
    unsafePerformIO $ do
      root <- runIO $ findParentContaining "cabal.project"
      let
        elfDir = root </> firmwareBinariesDir "riscv32imc-unknown-none-elf" Release
        elfPath = elfDir </> "smoltcp_client"

      vecsFromElf
        @(64 * 1024) -- 64 KiB data memory
        @(256 * 1024) -- 256 KiB instruction memory
        BigEndian
        elfPath
        Nothing
```

hydrolarus: We were using MemBlobs to start with because constructing the `Vec`s for the programs in TemplateHaskell was really resource intensive, maybe at runtime it's better? But just something to be aware of, at least it bit us in the past

martijnbastiaan: Yeah, it's a Clash-the-compiler thing. Not a simulation thing. Blockram simulation will "normalize" everything to an `Array` anyway.

martijnbastiaan (changes requested):
I general LGTM, but let's remove the `CPU_INCLUDE_BINARIES` (if it is as easy as I think it is).

PR #694, opened 2024-12-04
Run nightly 10 times

martijnbastiaan:
We think https://github.com/bittide/bittide-hardware/pull/690 fixed our nightly issues. At the very least, we cannot reproduce it locally anymore. However, we've also seen that nightly fails more often than local runs, so we'll run nightly a bunch of times in the coming days to see if we've actually fixed things.


hiddemoll (approved):
This is quality programming. A single nightly CI run takes ~1h5m, so 10 runs should stay under the job queue time limit of 24h.

PR #697, opened 2024-12-09
Update HAL for scatter and gather engines

hiddemoll:
This PR updates the outdated Hardware Abstraction Layer for the scatter and gather engines. It also adds a simulation to test the updated HALs. 

Since we use `DataLink` in bittide internals, we cannot use address 0. This makes the test a bit more complex to read, mostly because of all the conversions from bytes to 32-bit and 64-bit words you have to make in your head. I think it may be better to wait on #684 before merging this PR.


hiddemoll:
@rslawson could you review the Rust code in this PR?
@lmbollen I'm assigning you as reviewer for the rest, since you created issue #663

bittide-instances/tests/Wishbone/ScatterGather.hs
```diff
@@ -0,0 +1,111 @@
...
+  ( HiddenClockResetEnable dom
+  , KnownNat nBytes
+  , 1 <= nBytes
+  , KnownNat addrW
+  ) =>
+  ScatterConfig nBytes addrW ->
+  GatherConfig nBytes addrW ->
+  Circuit (Df dom (BitVector 8)) (Df dom (BitVector 8))
+dut scatterConfig gatherConfig = circuit $ \uartRx -> do
+  jtagIdle <- idleSource -< ()
```
lmbollen, line 85: We have `IdleCircuit` instances for tuples, so you can write `(jtagIdle, wbGuCal, wbSuCal) <- idleSource -< ()`

bittide-instances/tests/Wishbone/ScatterGather.hs
```diff
@@ -0,0 +1,111 @@
...
+  forall dom nBytes addrW.
+  ( HiddenClockResetEnable dom
+  , KnownNat nBytes
+  , 1 <= nBytes
+  , KnownNat addrW
+  ) =>
+  ScatterConfig nBytes addrW ->
+  GatherConfig nBytes addrW ->
+  Circuit (Df dom (BitVector 8)) (Df dom (BitVector 8))
+dut scatterConfig gatherConfig = circuit $ \uartRx -> do
```
lmbollen, line 84: You are also idle driving `uartRx` in the declaration of `uartStream` above.

The code would be cleaner if you remove that incoming circuit from `dut` and drive it similar to the other unconnected circuits.

bittide-instances/tests/Wishbone/ScatterGather.hs
```diff
@@ -0,0 +1,111 @@
...
+  jtagIdle <- idleSource -< ()
+  [uartBus, wbSu, wbGu] <- processingElement NoDumpVcd peConfig -< jtagIdle
+  (uartTx, _uartStatus) <- uartInterfaceWb d2 d2 uartSim -< (uartBus, uartRx)
```
lmbollen, line 87: You might want to increase the tx buffer size slightly to prevent the CPU from waiting during the prints.
`d8` or `d16` should be fine.

bittide-instances/tests/Wishbone/ScatterGather.hs
```diff
@@ -0,0 +1,111 @@
...
+
+case_scatter_gather_echo_test :: Assertion
+case_scatter_gather_echo_test = do
+  assertBool
+    msg
+    (P.head (lines simResult) == "Written data was read back correctly")
+ where
+  msg =
+    "Received the following from the CPU over UART:\n"
+      <> unlines (P.take 5 (lines simResult))
```
lmbollen, line 43: Why take 5? If you want to limit the maximum execution time of the test or prevent timeout, you should probably do so by manipulating the configuration of `sampleC` instead.

lmbollen (changes requested):
Clash part looks good modulo some minor comments and one question

bittide-instances/tests/Wishbone/ScatterGather.hs
hiddemoll: I added a separate commit doing the same for the other bittide-instances unittests which also had an idle uartRx.

firmware-binaries/test-cases/scatter_gather_test/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    unsafe { gather_unit.copy_from_slice(&source, 0) };
+
+    // Second metacycle
+    gather_unit.wait_for_new_metacycle();
+
+    // Third metacycle
+    gather_unit.wait_for_new_metacycle();
+    let mut destination: [u8; MEM_SIZE] = [0; MEM_SIZE];
+    unsafe { scatter_unit.copy_to_slice(destination.as_mut(), 0) };
+
```
rslawson, line 50: By implementing `Deref` (with `Target = [u8; MEM_SIZE]`, [link](https://doc.rust-lang.org/std/ops/trait.Deref.html)) and `DerefMut` ([link](https://doc.rust-lang.org/std/ops/trait.DerefMut.html)) for `GatherUnit` and `ScatterUnit`, you can instead use the safe slice method `slice::copy_from_slice` and remove `GatherUnit::copy_from_slice` and `ScatterUnit::copy_to_slice`. Note that line 49 would have to change to `destination.copy_from_slice(&scatter_unit[..])` with this change.

firmware-support/bittide-sys/src/gather_unit.rs
```diff
@@ -1,82 +1,33 @@
...
 // SPDX-License-Identifier: Apache-2.0
-
-use fdt::node::FdtNode;
-
-use crate::{utils::matches_fdt_name, ComponentLoadError};
-
-pub struct GatherUnit<const FRAME_SIZE: usize> {
+pub struct GatherUnit<const MEM_SIZE: usize> {
     memory: *mut u8,
     metacycle_register: *const u8,
```
rslawson, line 6: I would prefer instead to have just the base address stored and create an associated constant for any other registers for the peripheral, like is done for `ClockControl` (`firmware-support/bittide-sys/src/clock_control.rs`) and `Clock` (`firmware-support/bittide-sys/src/time.rs`). Should be a free memory footprint decrease with no performance hit (or sometimes even a notable performance increase).

firmware-support/bittide-sys/src/gather_unit.rs
```diff
@@ -1,82 +1,33 @@
...
-                self.memory.add(i).write_volatile(d);
-            }
-        }
+    /// # Safety
+    ///
+    /// The source memory size must be smaller or equal to the size of the
+    /// `GatherUnit` memory.
+    pub unsafe fn copy_from_slice(&self, src: &[u8], offset: usize) {
+        assert!(src.len() + offset <= MEM_SIZE * 4);
+        core::ptr::copy_nonoverlapping(src.as_ptr(), self.memory.add(offset), src.len());
```
rslawson, line 30: Can be removed if `Deref` and `DerefMut` are implemented.

firmware-support/bittide-sys/src/scatter_unit.rs
```diff
@@ -1,82 +1,34 @@
...
-            }
-        }
+    /// # Safety
+    ///
+    /// The destination memory size must be smaller or equal to the size of the
+    /// `ScatterUnit`.
+    pub unsafe fn copy_to_slice(&self, dst: &mut [u8], offset: usize) {
+        let src = self.memory.add(offset);
+        assert!(dst.len() + offset <= MEM_SIZE * 4);
+        core::ptr::copy_nonoverlapping(src, dst.as_mut_ptr(), dst.len());
```
rslawson, line 31: Can be removed if `Deref` and `DerefMut` are implemented.

firmware-support/bittide-sys/src/scatter_unit.rs
```diff
@@ -1,82 +1,34 @@
...
-
-use fdt::node::FdtNode;
-
-use crate::{utils::matches_fdt_name, ComponentLoadError};
-
-pub struct ScatterUnit<const FRAME_SIZE: usize> {
+pub struct ScatterUnit<const MEM_SIZE: usize> {
     memory: *const u8,
     metacycle_register: *const u8,
 }
```
rslawson, line 7: See comment on the definition of `GatherUnit<const MEM_SIZE: usize> { ... }`.

firmware-binaries/test-cases/scatter_gather_test/src/main.rs
rslawson: Sorry, took someone reminding me that this is MMIO so it needs to be handled a bit differently. I'll write the relevant adjustments in the comment on `copy_from_slice` just below, but this can still have a safe interface.

firmware-support/bittide-sys/src/gather_unit.rs
rslawson: So this shouldn't be removed, but we can present a safe interface if you write the necessary asserts. If the `assert!` that already exists in this function is enough to guarantee that no out-of-bounds writes can occur, then it would be valid here to move the `unsafe` from the function to a block inside it. Also here, since this is MMIO, we mostly want to be using volatile operations to ensure the compiler doesn't elide them. What the HAL had previously is close to what it seems like you want, except for the ability to copy something smaller than a frame and have an offset. Something like
```rs
pub fn copy_from_slice(&self, src: &[u8], offset: usize) {
    assert!(src.len() + offset <= MEM_SIZE * 4);
    let mut off = offset;
    for &byte in src {
        unsafe {
            self.base_addr.add(off).write_volatile(byte);
        }
        off += 1;
    }
}
```
should probably get the job done.

firmware-support/bittide-sys/src/scatter_unit.rs
rslawson: Ignore, see comment on `copy_from_slice`, but replace the `write_volatile` with a `read_volatile` and assign that to the appropriate index in `dst`.

firmware-support/bittide-sys/src/gather_unit.rs
hiddemoll: I've implemented your change request, but sadly this does make reading/writing roughtly 4 times slower than it was with `copy_nonoverlapping`. I've increased the padding of the calendar for the test by a factor of 4.
Do you think this change is worth it?
Do you agree with the new function names of the read/write functions for the scatter and gather units?

rslawson (approved):
Rust side looks good to me (:

lmbollen (approved):

PR #701, opened 2025-01-08
Bump `clash-vexriscv` for new openocd

lmbollen:
Previously `vexriscv` used a custom debug plugin with a custom openocd version `openocd-vexriscv`.

The new version of `clash-vexriscv` uses the official `EmbeddedJtag` plugin that is compatible with `openocd-riscv`

PR #703, opened 2025-01-10
Add `compare-sections` to gdb scripts

lmbollen:
`compare-sections` verifies that the sections on the target and local are the same.

Depends on #701


bittide-instances/data/gdb/smoltcp-hitl-prog.gdb
```diff
@@ -14,6 +14,9 @@ break smoltcp_client::gdb_panic
 target extended-remote :3333
 load
 
+#Verify whether we transfered the binary correctly
```
martijnbastiaan, line 17:
```suggestion
# Verify whether we transfered the binary correctly
```

We always do a space after a `#`. Keep it neat :).

bittide-instances/data/gdb/test-gdb-prog.gdb
```diff
@@ -7,6 +7,9 @@ target extended-remote :3333
 # Load binary on CPU
 load
 
+#Verify whether we transfered the binary correctly
+compare-sections
```
martijnbastiaan, line 11:
```suggestion
# Verify whether we transfered the binary correctly
compare-sections
```

martijnbastiaan (approved):

PR #705, opened 2025-01-15
add a VivadoM monad to make writing HITL drivers nicer

hydrolarus:
First implementation of a `VivadoM` monad for common Vivado operations.

This is only geared towards making ""user facing"" drivers nicer to write. This could be expanded later to
- add parallelism primitives
- include enough operations/features to move all the Vivado test infrastructure to this monad?


martijnbastiaan:
Not a proper review, but one of my hopes of moving from TCL to Haskell (the whole `vivado-hs` stuff) was that using multiple Vivados in parallel would become easy. Do you have any thoughts on whether or not this abstraction would make that more or less difficult?

PR #706, opened 2025-01-16
CI: Update freeze file and make sure it stays up to date

leonschoorl:
cabal.project had two lines different index-state lines.
And the one we've been update wasn't the one cabal was actually using.
For good measure I've also updated the index-state to now.

Also freeze file to include updated index-state and dependencies to make sure the cache works properly, as the freeze file is part of the cache key.
This fixes the constant rebuilding of the word8 library CI was doing everywhere.

And I've added a check on CI that makes sure the freeze file is up to date.
If the cabal.project.freeze file it out of date it will look like [this](https://github.com/bittide/bittide-hardware/actions/runs/12810442006/job/35717513357).


leonschoorl:
If this CI check every fails on you, just run `cabal freeze` and commit the updated `cabal.project.freeze` file.

If you still have issues, make sure you don't have a `cabal.project.local` file changing things the CI doesn't know about.

martijnbastiaan (approved):
Great

PR #710, opened 2025-01-24
Only test one topology on PRs

martijnbastiaan:
Considerably speeds up CI. We still run all the topologies at night.

Unfortunately I had to copy+paste the top level type signature, as `makeTopEntity` requires it.


lmbollen (approved):
I'm gonna go ahead and approve but can you try to write the type signature using a type synonym or will you lose the names?

martijnbastiaan:
> or will you lose the names?

I thought so, but I've heard that adding `"" ::: TheTypeSynonym` should work!

PR #712, opened 2025-01-24
Only test one topology on PRs

lmbollen:
Duplicate of #710 to prevent a new CI run:

>Considerably speeds up CI. We still run all the topologies at night.
Unfortunately I had to copy+paste the top level type signature, as makeTopEntity requires it.


martijnbastiaan (approved):
Danku

PR #713, opened 2025-01-24
Mb/no werror when debug is present

martijnbastiaan:


lmbollen (approved):
You would not merge untested code, would you?

leonschoorl (approved):

PR #714, opened 2025-01-29
Refactor GDB related code

lmbollen:
This refactor improves reusability of the code surrounding GDB by putting it in its own module and reducing the number of hardcoded strings

### DIFF
--- a/bittide-instances/src/Bittide/Instances/Hitl/Driver/VexRiscvTcp.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/Driver/VexRiscvTcp.hs
@@ -122,7 +122,6 @@ driverFunc _name [(hwT, dI)] = do
             gdb
             [ "core::panicking::panic"
             , "ExceptionHandler"
-            , "rust_begin_unwind"
             , "smoltcp_client::gdb_panic"
             ]
           Gdb.setBreakpointHook gdb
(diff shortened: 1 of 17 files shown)


bittide-instances/src/Bittide/Instances/Hitl/Driver/VexRiscvTcp.hs
```diff
@@ -98,65 +88,45 @@ driverFunc _name [(hwT, dI)] = do
...
+          Gdb.setLogging gdb gdbOutLog
+          Gdb.setFile gdb $ firmwareBinariesDir "riscv32imc" Release </> "smoltcp_client"
+          Gdb.setTarget gdb 3333
+          errorToException =<< Gdb.loadBinary gdb
+          -- errorToException =<< Gdb.compareSections gdb
+          Gdb.setBreakpoints
+            gdb
+            [ "core::panicking::panic"
+            , "ExceptionHandler"
+            , "rust_begin_unwind"
```
rslawson, line 125: So as I've found out through my troubles with getting backtraces working, `rust_begin_unwind` doesn't exist outside of `std` (and we're in `#![no_std]`), and also shouldn't exist if `panic = "abort"`, as we have set. This breakpoint can actually be removed.

rslawson (approved):
Generally looks good to me (:

Issue #696, opened 2024-12-04
Remove redundant logic dealing with multiple TX domains


martijnbastiaan:
After merging https://github.com/bittide/bittide-hardware/pull/688, we can remove all code dealing with multiple TX domains. The easiest way is probably to change:

https://github.com/bittide/bittide-hardware/blob/c92c648744ca98129f646923585b43f97903072d/bittide/src/Bittide/Transceiver.hs#L175-L176

to

```haskell
 { txClock :: Clock tx
 -- ^ See 'Output.txClock' 
```

and see where the type errors lead.

PR #716, opened 2025-02-05
Move all instances using transceivers to one TX domain (+whole bunch of bug fixes)

martijnbastiaan:
In https://github.com/bittide/bittide-hardware/pull/688 we modified the transceiver logic to operate on a single TX domain (but multiple RX domains). To keep the changes contained, we still exported multiple clocks still (though just as a `repeat`). This PR cleans this up and removes all superfluous synchronization. During this process we discovered a number of bugs:

1. Transceiver up test would only fail if _all_ links would fail. This is of course an error: it should fail if _any_ link fails.
2. We had two fields called `txReady` (both on `Input` and `Output`). This caused confusion, that lead to the input's `txReady` to be looped back to the `Output`. This was clearly incorrect: `Output.txReady` should clearly depend on whether a neighbor says it is ready to receive data!

Fixes https://github.com/bittide/bittide-hardware/issues/696

# TODO
- [x] See if we can replicate handhshake failure in simulation test

### DIFF
--- a/bittide-instances/src/Bittide/Instances/Hitl/LinkConfiguration.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/LinkConfiguration.hs
@@ -179,7 +179,7 @@ transceiversStartAndObserve refClk sysClk rst myIndex rxNs rxPs miso =
         , rxPs
         , txDatas = repeat myIndexTx
         , rxReadys = repeat $ pure True
-        , txReadys = repeat $ pure True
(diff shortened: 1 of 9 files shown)


martijnbastiaan:
Thanks :). Virtually all code was written by @leonschoorl, I just made the PR. It also has my seal of approval.

> If you found any potential issues in the transceiver code now, or other things you'd like to fix in the future, make issues now that they are fresh in your heads.

We haven't found anything we didn't fix.

PR #717, opened 2025-02-05
Fix documentation and remove problematic start probe deassertion.

rslawson:
Title says it.


martijnbastiaan:
For context, we were seeing this on the nightlies:

Screenshot from 2025-02-05 13-14-59

Which is caused by the reset issue described in the commit.

martijnbastiaan (approved):
Makes sense. Let's wait for the plots before merging.

martijnbastiaan:
Plots look good!

martijnbastiaan:
Let's go bouldering? :-)

PR #731, opened 2025-03-04
Add user words to transceiver wires

martijnbastiaan:
Allows for accessing data at top entity boundaries. Prepares for a multi-node simulation including transceiver data (not just clock signals).


hiddemoll (approved):
No remarks other than a typo

PR #732, opened 2025-03-05
Remove Rust dependency from all instances

hiddemoll:
This is the first step to synthesis caching.


martijnbastiaan (approved):
> This is the first step to synthesis caching.

Very much looking forward to this

PR #581, opened 2024-07-25
Solve the constant crisis by adding memory map description to circuits

hydrolarus:
This implements #481 

There's a few noteworthy parts to this PR:
- the description of memory maps and their validation (`Protocols.MemoryMap`/`.Check`)
- the `BitPackC` type class
- the `ToFieldType` type class
- a generator for Rust wrappers based on the memory map `firmware-support/memmap-generate`

PR #735, opened 2025-03-17
Solving the Constant Crisis one step at a time

hydrolarus:
After the first try in #581 this is another attempt which carries around memory maps more explicitly.

This PR adds:
- a way to attach memory maps to circuits and build them from the bottom up
- a new typeclass `BitPackC` which allows for packing and unpacking with a C compatible layout
- a new typeclass `ToFieldType` and the description `FieldType` to represent data types
- memory map implementations for a number of components used but NOT all *yet*
- a way to check memory maps for validity
- a way to serialise memory maps into JSON
- some currently unused Rust code that still needs to be updated but we want to add since it's a good skeleton

This PR does NOT add:
- composable building blocks for defining device definitions
- HAL generation in Rust, this is a TODO for another PR, hence "one step at a time"
  - this should be a much smaller step though
- great memory maps for all possible circuits **yet**, either this PR gets updated or they get filled at a later step


clash-protocols-memmap/src/Protocols/MemoryMap.hs
```diff
@@ -0,0 +1,340 @@
...
+    def'.registers
+
+data Access
+  = -- | Managers should only read from this register
+    ReadOnly
+  | -- | Managers should only write to this register
+    WriteOnly
+  | -- | Managers can read from and write to this register
+    ReadWrite
+  deriving (Show)
```
martijnbastiaan, line 183: Perhaps add a TODO that one day we want to support all of this (see "bus access modes"):

https://corsair.readthedocs.io/en/latest/regmap.html#bit-field

clash-protocols-memmap/src/Protocols/MemoryMap.hs
```diff
@@ -0,0 +1,340 @@
...
+  (ToFieldType b, BitPackC a) =>
+  (1 <= AlignmentC a) =>
+  RegisterType
+regTypeSplit = RegisterType (Proxy @(RegTypeSplit a b))
+
+data RegTypeSplit a b
+
+instance (ToFieldType b) => ToFieldType (RegTypeSplit a b) where
+  type Generics (RegTypeSplit a b) = Generics b
+  type WithVars (RegTypeSplit a b) = WithVars b
```
martijnbastiaan, line 196: What's all this? Could you document it?

clash-protocols-memmap/src/Protocols/MemoryMap/Check.hs
```diff
@@ -0,0 +1,12 @@
+-- SPDX-FileCopyrightText: 2025 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+-- | Validity checks performed on memory maps
+module Protocols.MemoryMap.Check (
+  module Protocols.MemoryMap.Check.AbsAddress,
+  module Protocols.MemoryMap.Check.Normalised,
```
martijnbastiaan, line 8: I'm going to be annoying and say that we decided to do everything in US English a couple of years ago, sorry :(

clash-protocols-memmap/src/Protocols/MemoryMap/Check/Normalised.hs
```diff
@@ -0,0 +1,126 @@
...
+  | Interconnect SrcLoc [(RelAddress, MemoryMapTree)]
+  | DeviceInstance SrcLoc DeviceName
+  deriving (Show)
+
+type MemoryMapTreeRelNorm =
+  MemoryMapTreeAnn ([(SrcLoc, String)], Path, Maybe (SrcLoc, Address)) 'Normalised
+
+type AbsAddressList = [(Path, Address)]
+
+data Normalised = Normalised | NotNormalised
```
martijnbastiaan, line 46:
```suggestion
data Normalized = Normalized | NotNormalized
```

bittide/src/Bittide/SwitchDemoProcessingElement.hs
```diff
@@ -217,3 +221,83 @@ switchDemoPeWb SNat localCounter = Circuit go
         <$> writeVec
 
     (writeVec, wbS2M) = unbundle $ wbToVec <$> bundle readVec <*> wbM2S
+
+  mm =
+    MemoryMap
+      { tree = DeviceInstance locCaller "SwitchDemoPE"
+      , deviceDefs = deviceSingleton deviceDef
+      }
+  deviceDef =
```
martijnbastiaan, line 230: This function now both defines the hardware and the memory map. I'm assuming it's future work to define the hardware in a way that generates the memory map. Is that correct? If so, could you add that to the "This PR does NOT add:" list in the PR description?

martijnbastiaan (commented):
Lookin' good :D

clash-protocols-memmap/src/Protocols/MemoryMap.hs
```diff
@@ -0,0 +1,340 @@
...
+data MemoryMap = MemoryMap
+  { deviceDefs :: DeviceDefinitions
+  , tree :: MemoryMapTree
+  }
+  deriving (Show)
+
+data DeviceDefinition = DeviceDefinition
+  { deviceName :: Name
+  , registers :: [NamedLoc Register]
+  , defLocation :: SrcLoc
```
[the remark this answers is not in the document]
hydrolarus: This is meant as "definition location", not "default location" :sweat_smile:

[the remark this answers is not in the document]
martijnbastiaan: Ah ok, well let's write it out then :-)

clash-protocols-memmap/src/Protocols/MemoryMap.hs
```diff
@@ -0,0 +1,340 @@
...
+  , tags :: [String]
+  }
+  deriving (Show)
+
+deviceSize :: DeviceDefinition -> Integer
+deviceSize def' =
```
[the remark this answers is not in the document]
martijnbastiaan: Yeah let's write it out then.

bittide/src/Bittide/SwitchDemoProcessingElement.hs
hydrolarus: Yes, at the moment I just added the memory maps separately to the implementation. There's already building blocks that build it up automatically, like `singleMasterInterconnect` or `processingElement`, but a lot of different circuits use different ways to implement them, so there wasn't a single place I could have added a memory map implementation to help build up those device definitions :cry: Maybe one day though!

clash-protocols-memmap/src/Protocols/MemoryMap.hs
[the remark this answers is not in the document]
hydrolarus: other places also use `dev` so I changed it here too.

bittide/src/Bittide/Node.hs
```diff
@@ -31,21 +31,22 @@ The 'scatterUnitWb's and 'gatherUnitWb's are initialized with 'CalendarConfig's
...
+--     (repeat (GppeConfig (ScatterConfig sgConfig) (GatherConfig sgConfig) peConfig NoDumpVcd))
+--  where
+--   switchCal = CalendarConfig (SNat @1024) (switchEntry :> Nil) (switchEntry :> Nil)
+--   sgConfig = CalendarConfig (SNat @1024) (sgEntry :> Nil) (sgEntry :> Nil)
+--   peConfig = PeConfig memMapPe (Undefined @8192) (Undefined @8192) d0 d0
+--   nmuConfig = PeConfig memMapNmu (Undefined @8192) (Undefined @8192) d0 d0
+--   memMapPe = iterateI (+ 0x1000) 0
+--   memMapNmu = iterateI (+ 0x1000) 0
+--   switchEntry = ValidEntry{veEntry = repeat 0, veRepeat = 0 :: Unsigned 0}
+--   sgEntry = ValidEntry{veEntry = 0 :: Index 1024, veRepeat = 0 :: Unsigned 0}
```
hiddemoll, line 49: I guess you can remove this entirely, it is not used anywhere

bittide/src/Bittide/Wishbone.hs
```diff
@@ -28,11 +28,17 @@ import Data.Bool (bool)
...
 import Protocols
 import Protocols.Wishbone
 
+import BitPackC
 import Clash.Sized.Vector.ToTuple (VecToTuple (vecToTuple))
+import GHC.Stack (HasCallStack)
 import qualified Protocols.Df as Df
+import qualified Protocols.MemoryMap as MM
+import Protocols.MemoryMap.FieldType (ToFieldType)
 import qualified Protocols.Wishbone as Wishbone
```
hiddemoll, line 42: Can you sort the imports (here and in some other modules where you add qualified imports). We try to put qualified imports in a separate category. Sadly this is not fixed by fourmolu (until v0.17.0.0)... 
The best example is in [`Tests.ClockControlWb`](https://github.com/bittide/bittide-hardware/blob/staging/bittide-instances/tests/Tests/ClockControlWb.hs).

hiddemoll (commented):
Nice work! I basically only have some formatting remarks. Other than that I trust the successful tests.

firmware-support/memmap-generate/src/generators/types.rs
```diff
@@ -0,0 +1,233 @@
...
+    } = &ty.definition
+    {
+        match variants.as_slice() {
+            [] => quote! {},
+            [x] => {
+                if &x.name == &ty.name {
+                    quote! { #[repr(C)] }
+                } else {
+                    panic!("Single-constructor data types are required to have the constructor name match the type name.")
+                    // quote! { #[repr(u8)] }
```
rslawson, line 181: Remove?

rslawson (commented):
Nothing stood out to me as incorrect, and if CI says it passes then it's probably good (:

clash-protocols-memmap/src/Protocols/MemoryMap/Check/Normalized.hs
```diff
@@ -0,0 +1,126 @@
...
+type Path = [PathComp]
+
+type Address = Integer
+type RelAddress = Integer
+type Prefix = Integer
+type DeviceName = String
+
+{- | A tree structure that describes the memory map of a device. Its definitions
+are using non-translatable constructs on purpose: Clash is currently pretty
+bad at propagating contants properly, so designers should only /produce/
```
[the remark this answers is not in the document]
hydrolarus: I think it's meant to be "constants", because a different approach than the one we use would be to pass a memory down into all the different circuits - if it's a non-synthesizable structure then we would have to rely on the compiler to optimise out all the uses of that value, only propagating the *constants* that are actually used and not the whole structure.

We use a different approach, this comment is just for clarification?

I just took it from Martijn's initial draft gist :sweat_smile:

firmware-support/Cargo.toml
```diff
@@ -4,6 +4,9 @@
 
 [workspace]
 members = [
-  "bittide-sys", "gdb-trace",
+  "bittide-sys",
+  "gdb-trace",
+  # "memmap-generate"
```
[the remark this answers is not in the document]
hydrolarus: At the moment  we are not using the Rust code for anything. There is outdated code in there and it's not fully in the standard we'd want it to be, so if it's part of the workspace it will produce tons of warnings etc.

This is mentioned in the PR description. I know the Rust code isn't great and nice yet, but that's the next step after this PR. So I'll leave your comments open so I can look at them later, but not address them for now. Is that okay?

hydrolarus:
I tried to address all your comments, thank you! :pray: I left the comments regarding the (currently unused) Rust code unresolved so I can look at them later, once this PR is merged.

bittide/src/Bittide/SwitchDemoProcessingElement.hs
martijnbastiaan: Understandable. I have to build some components from the ground up soon, we can try to add new primitives then and over time phase out the non-MM ones.

bittide-instances/src/Bittide/Instances/Pnr/Ethernet.hs
```diff
@@ -98,18 +99,37 @@ vexRiscGmii SNat sysClk sysRst rxClk rxRst txClk txRst fwd =
...
-          _dna <- dnaC -< dnaWb
-          macStatIf -< (macWb, macStatus)
+          mm <- idleSource -< ()
+          [ (preUart, (mmUart, uartBus))
+            , (preTime, (mmTime, timeBus))
+            , (preAxiRx, (mmAxiRx, wbAxiRx))
+            , (preAxiTx, (mmAxiTx, wbAxiTx))
+            , (preDna, (mmDna, dnaWb))
+            , (preGpio, (mmGpio, gpioWb))
+            , (preMac, (mmMac, macWb))
```
martijnbastiaan, line 109: I would _love_ some kind of ViewPattern style here such that we could write:
```suggestion
          [   (>- constBwd 0b0010, (mmUart, uartBus))
            , (>- constBwd 0b0011, (mmTime, timeBus))
            , (>- constBwd 0b0101, (mmAxiRx, wbAxiRx))
            , (>- constBwd 0b0110, (mmAxiTx, wbAxiTx))
            , (>- constBwd 0b0111, (mmDna, dnaWb))
            , (>- constBwd 0b0100, (mmGpio, gpioWb))
            , (>- constBwd 0b1001, (mmMac, macWb))
```

I wonder if this is syntactically valid. If it is, we _could_ patch the plugin..

martijnbastiaan: Also, IMO `pre` -> `prefix`.

martijnbastiaan: Am I correct in stating that we expect the right tuple deconstruction to disappear in the future? I.e., `(mmUart, uartBus)` -> `uartMmBus` because the UART component would produce a MM too. It could still be a tuple, of course.

martijnbastiaan (approved):
Relieved to see that Clash doesn't slow down. LGTM, excited to get this merged. (Let's wait for @hiddemoll, @rslawson or both though :-).)

hiddemoll (approved):

bittide-instances/src/Bittide/Instances/Pnr/Ethernet.hs
rslawson: > Also, IMO `pre` -> `prefix`.

Would you also be okay with something like `uartPfx`, `timePfx`, etc. or would you rather see things written out fully as `uartPrefix` and so on?

rslawson (approved):

bittide-instances/src/Bittide/Instances/Pnr/Ethernet.hs
hydrolarus: I can name it `prefix`, sure!

> Am I correct in stating that we expect the right tuple deconstruction to disappear in the future?

In some cases it is already I think, for the timer in some examples. This depends on how the circuit expects the memory map and wishbone bus to be passed. In the case of the UART it doesn't quite fit (`(memoryMap, (wishbone, rx))`, it would have to be `((memorymap, wishbone), rx)` to make it work), but it could be made to fit better.

Is that something I should do? Right now it doesn't work because the UART uses "normal form" of the memory map first and all the rest second, but it could be changed so that it works better with the interconnect.

hydrolarus: Ah, in many places there's already `prefixX` variables in scope, for example from configs, for example `ManagementConfig` has the prefixes stored in it, so we do
```hs
    constBwd prefixSCal -< preSCal
    constBwd prefixS -< preS
    constBwd prefixGCal -< preGCal
    constBwd prefixG -< preG
```

I can still change the ones where there is no clash (and keep it consistent within the circuit)

rslawson (approved):

PR #760, opened 2025-04-29
Use GHA matrix for Haskell tests

martijnbastiaan:


hiddemoll (approved):
Good idea, especially since not all tests were run (e.g. bittide-extra).

hiddemoll (approved):

Issue #481, opened 2024-03-14
Sovling the constant crisis


martijnbastiaan:
We currently hard code memory addresses all across our code base. As far as I can tell, they are scattered across three parts of our code base:

1. `clash-vexriscv`. It contains a (hard coded) reset vector (which we patch in this repo to be somewhere else, but that's another story).
2. Our hardware designs in `bittide-instances`
3. Our Rust code in `firmware-binaries`

This has always been undesirable, but lately it has been reaching a boiling point due our stack maturing and us wanting to do more and more in software. To solve this, we basically have two options:

1. Define a memory map, and integrate it in our hardware designs. This is what https://github.com/bittide/bittide-hardware/pull/315 has proposed, based on https://github.com/bittide/bittide-hardware/issues/47.

2. Use _the power of Clash_ to create a protocol agnostic wrapper for memory mapped protocols, see a proposal below.

Originally (2) was just a weekend project of mine and I wouldn't have proposed it as an alternative to (1). However, during its implementation I found that:

* This would allow us to generate Rust data types / getters / setters for the contents of the registers, straight from the Haskell types.
* This would allow us to generate documentation straight from Haskell types (this would need some work, so that's not really a MVP goal).
* It doesn't rely on Clash constant folding in presence of HO-functions. (Alternatively worded: we're not restricted in the way we design our circuits.)
* We can change ONE thing (the circuit) and have the rest follow suit. That it, there is no way the DTS would get out of sync with the design.

I'm not sure how much work there is left to be done, but it seems worthwhile to at least investigate to see how much time it would take to get to an MVP.

--------------------------------

```haskell
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE MagicHash #-}
{-# LANGUAGE NamedFieldPuns #-}

-- | A protocol agnostic wrapper for memory mapped protocols. See 'MemoryMapped'
-- for more information. Throughout this module we'll refer to devices initiating
-- a transaction on a bus as \"managers\" and devices receiving a transaction as
-- \"subordinates\".
--
-- TODO:
--
--   * Provide examples on how to use the data structures protocols
--   * Provide standard memory map aware registers and interconnects
--   * Provide a way to generate human readable documentation
--   * Provide a way to generate data structures for target languages
--   * Provide a way to check whether memory maps are valid
--
module Protocols.MemoryMapped where

import Control.Monad (void, forM)
import Data.Bits (popCount)
import Data.Foldable (toList, Foldable (foldl'))
[...]

PR #775, opened 2025-05-17
Solving the Constant Crisis 2: Electric Boogaloo

hydrolarus:
This is the second PR in the series of adding and integrating memory maps into the system #481 

This second PR most importantly adds Rust code generation for the devices and types used in a memory map. This is done in two parts

- the `memmap-generate` crate
  - code to read in a memory map in JSON format into a structure nicely readable and usable from Rust
  - deduplicator for devices and types when multiple memory maps are used (`MemoryMapSet`) 
  - generators from `MemoryMapSet` to Rust code
  - this is where future C output could live!

- the `bittide-hal` crate
  - if `memmap-generate` is the crate that generates all the code, this is the crate that actually contains the generated code!
  - this crate contains a module for each memory map found, with all the devices and types exposed for that hardware
  - because raw hardware access often isn't the most pleasant or enough, this also contains a `manual_additions` module where nicer wrappers and trait instances can be defined


martijnbastiaan (commented):
1. Can the generated Rust HALs include descriptions from the JSON? 
2. How hard is it to run a formatter on the generated files? I don't really care about the actual formatting, but making them a little bit human readable makes sense to me -- even if they're generated files.

martijnbastiaan:
From our Slack:

Screenshot from 2025-05-17 11-40-02

martijnbastiaan:
One more thing from my testing: passing in "weird" names makes the generator render broken names in the Rust code. The easiest thing is to just error if the name is unexpected. Can go onto the issue list though :).

martijnbastiaan:
A `cabal build bittide-instances` doesn't clean up the `_build/memory_maps`. It probably should, as it leaves obsolete (potentially broken) memory maps behind upon renaming.

bittide-instances/src/Bittide/Instances/MemoryMaps.hs
```diff
@@ -0,0 +1,126 @@
...
+              msg =
+                case err of
+                  SizeExceedsError
+                    { startAddr
+                    , availableSize
+                    , requestedSize
+                    , path
+                    , location
+                    } ->
+                      printf
```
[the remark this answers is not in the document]
hydrolarus: `printf` is polymorphic in the return type, here it creates a `String` and doesn't print anything on its own! The printing happens later with `reportError`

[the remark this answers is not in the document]
martijnbastiaan: Oh woops, sorry

bittide-instances/src/Bittide/Instances/MemoryMaps.hs
```diff
@@ -0,0 +1,109 @@
...
+              makeAbsolute mm.deviceDefs (0x0000_0000, 0xFFFF_FFFF) normalised
+
+    forM_ (memoryMaps `zip` absResults) $ \((mmName, mm), (absTree, errors)) -> do
+      if not $ null errors
+        then do
+          -- report errors
+          forM_ errors $ \err -> do
+            let
+              msg =
+                case err of
```
martijnbastiaan, line 72: Could you factor this out into a `getErrorMessage`?

bittide/src/Bittide/ProcessingElement.hs
```diff
@@ -100,9 +106,22 @@ processingElement dumpVcd PeConfig{prefixI, prefixD, initI, initD, iBusTimeout,
     (splitAtCI <| singleMasterInterconnectC) -< (mmDbus, dBus1)
   MM.constBwd prefixD -< dPre
   MM.constBwd prefixI -< iPre
-  wbStorage "DataMemory" initD -< (mmD, dMemBus)
+
+  -- Both the instance and the device types are marked with
+  -- "no-generate" so that a code generator doesn't create wrappers.
+  -- We don't want n different versions of the data and instruction memories
+  -- generated if we don't even want to use them!
```
martijnbastiaan, line 113:
```suggestion
  -- Instruction and data memory are never accessed explicitly by developers, only
  -- implicitly by the CPU itself. We therefore don't need to generate HAL code. We
  -- instruct the generates to skip them by adding a "no-generate" tag.
```

I think the most confusing thing for me was: _why_ don't we want to use them. I guess this is the reason?

bittide/src/Bittide/ProcessingElement/Util.hs
```diff
@@ -67,6 +68,31 @@ vecsFromElf byteOrder elfPath maybeDeviceTree = do
...
+    dListPadded = padToSize "Data memory" (Just (natToNum @nWords)) 0 dList
+
+  pure (unsafeFromList dListPadded)
+unsafeVecFromElfInstr byteOrder elfPath = unsafePerformIO $ do
+  (iMemIntMap, _dMemIntMap) <- getBytesMems elfPath Nothing
+  let
+    (_iStartAddr, _, iList) = extractIntMapData byteOrder iMemIntMap
+    iListPadded = padToSize "Instruction memory" (Just (natToNum @nWords)) 0 iList
+
+  pure (unsafeFromList iListPadded)
```
martijnbastiaan, line 94: The documentation is currently wrong. I think you should opt for copying the type signature over trying to write generic documentation:

```suggestion
-- | Read data memory from ELF file. Marked `unsafe` because it `unsafePerformIO`.
unsafeVecFromElfData ::
  forall nWords.
  (HasCallStack, KnownNat nWords) =>
  -- | How the words should be ordered
  ByteOrder ->
  -- | Source file, assumed to be Little Endian.
  FilePath ->
  -- | Data memory
  Vec nWords (BitVector 32)
unsafeVecFromElfData byteOrder elfPath = unsafePerformIO $ do
  (_iMemIntMap, dMemIntMap) <- getBytesMems elfPath Nothing
  let
    (_dStartAddr, _, dList) = extractIntMapData byteOrder dMemIntMap
    dListPadded = padToSize "Data memory" (Just (natToNum @nWords)) 0 dList

  pure (unsafeFromList dListPadded)
  
-- | Read instruction memory from ELF file. Marked `unsafe` because it `unsafePerformIO`.
unsafeVecFromElfInstr ::
  forall nWords.
  (HasCallStack, KnownNat nWords) =>
  -- | How the words should be ordered
  ByteOrder ->
  -- | Source file, assumed to be Little Endian.
  FilePath ->
  -- | Instruction memory
  Vec nWords (BitVector 32)
unsafeVecFromElfInstr byteOrder elfPath = unsafePerformIO $ do
  (iMemIntMap, _dMemIntMap) <- getBytesMems elfPath Nothing
  let
    (_iStartAddr, _, iList) = extractIntMapData byteOrder iMemIntMap
    iListPadded = padToSize "Instruction memory" (Just (natToNum @nWords)) 0 iList

  pure (unsafeFromList iListPadded)
```

Also, IMO `unsafePerformIO` should move as close to where you're actually _need_ the unsafeness as possible. That would mean making this function operate in `IO` and where you use it (and you need the unsafeness), do the `unsafePerformIO`.

firmware-binaries/examples/hello/src/main.rs
```diff
@@ -42,14 +36,14 @@ fn main() -> ! {
...
-        uart.send(c);
+        let c = uart.data();
+        uart.set_data(c);
     }
 }
 
 #[panic_handler]
 fn panic_handler(_info: &core::panic::PanicInfo) -> ! {
-    test_failure();
+    INSTANCES.status_register.set_status(hal::TestStatus::Fail);
```
martijnbastiaan, line 46: Very cool.

martijnbastiaan (approved):
Very nice, thank you. I've got two questions:

1. I don't see it rejecting "wrong" names anywhere. Do you intend to open an issue about that?
2. For the `DeviceInstances` it generates:

```rust
impl DeviceInstances {
    pub const unsafe fn new() -> Self {
        DeviceInstances {
            timer: unsafe { Timer::new(3221225472u64 as *mut u8) },
            uart: unsafe { Uart::new(2147483648u64 as *mut u8) },
        }
    }
}
```

What's that `u64` doing there? Shouldn't it be `u32`?

bittide/src/Bittide/ProcessingElement.hs
hydrolarus: Yes, that is the reason. We don't need N*2 storage HALs that we never use. It's like accessing `/proc/PID/mem` on Linux, yeah you can access memory that way but you don't usually have a good reason to. I'll update to use your comment! :+1:

hydrolarus:
> 1. I don't see it rejecting "wrong" names anywhere. Do you intend to open an issue about that?

We could reject "wrong" names, we'd have to formalise what it means for a name to be proper for that! We could talk about that in an issue. Technically, adding another check pass on the memory map isn't a big deal

> 2.  For the DeviceInstances it generates:
> 
> ```rust
>impl DeviceInstances {
>    pub const unsafe fn new() -> Self {
>        DeviceInstances {
>            timer: unsafe { Timer::new(3221225472u64 as *mut u8) },
>            uart: unsafe { Uart::new(2147483648u64 as *mut u8) },
>        }
>    }
>}
> ```
> What's that `u64` doing there? Shouldn't it be `u32`?

That's mostly a quirk from how the data flows, since this is all about 32 bit systems, `u32` should be fine, in the JSON deserialiser I'm using `u64` for all integers, and if those are turned into `TokenStream`s they automatically get the postfix assigned. I was considering making the literal output for this a bit nicer, showing hex instead of decimal, I could also make it have a `u32` postfix, but functionally this shouldn't really have an effect on anything

hydrolarus:
After the latest push, the generated code for the `DeviceInstances` looks like this btw 
```rust
impl DeviceInstances {
    pub const unsafe fn new() -> Self {
        DeviceInstances {
            dna: unsafe { Dna::new(0x70000000 as *mut u8) },
            mac_status: unsafe { MacStatus::new(0x90000000 as *mut u8) },
            timer: unsafe { Timer::new(0x30000000 as *mut u8) },
            uart: unsafe { Uart::new(0x20000000 as *mut u8) },
        }
    }
}
```

martijnbastiaan:
>  the generated code for the DeviceInstances looks like this btw

Stunning! :-) PR LGTM!

PR #776, opened 2025-05-20
Fix cache fallout after merging #775

martijnbastiaan:
After HAL generation we need to archive the following directories on CI:

```
f"{PWD}/firmware-support/bittide-hal/src/shared/",
f"{PWD}/firmware-support/bittide-hal/src/hals/",
```

To keep `cache.py` from crashing, we also need to _exclude_ a file that already exists in the repository:

```
f"{PWD}/firmware-support/bittide-hal/src/shared/mod.rs",
```

To do that I've added exclude pattern support to the script.


hiddemoll (approved):

martijnbastiaan:
Oh well, it's a nice feature I guess :-). I was just following what you were doing in the `.gitignore`.

Issue #807, opened 2025-06-04
`ReframeStateKind` doesn't get decoded correctly


martijnbastiaan:
In Rust:

```rust
pub enum ReframingState {
    Detect,
    Done,
    Wait [..]
}
```

in Haskell:

```haskell
      readState 0 = Just Detect
      readState 1 = Just Wait
      readState 2 = Just Done
```

I.e., `Done` and `Wait` are swapped.

Should be auto-fixed when we use the infrastructure introduced in https://github.com/bittide/bittide-hardware/pull/801.

hydrolarus:
Is this still relevant with #801 merged now?

martijnbastiaan:
Yes, still relevant, #808 will make it use the infra in #801.

PR #808, opened 2025-06-04
Use `registerWbC` and friends in `debugRegisterWb`

martijnbastiaan:
Fixes https://github.com/bittide/bittide-hardware/issues/807


firmware-binaries/test-cases/clock-control-wb/src/main.rs
```diff
@@ -7,11 +7,9 @@
...
 use core::panic::PanicInfo;
 
-use bittide_sys::{
-    clock_control::{ClockControl, SpeedChange},
-    debug_register::DebugRegister,
-    uart::Uart,
-};
+use bittide_hal::shared::devices::clock_control::{ClockControl, SpeedChange};
+use bittide_hal::shared::devices::debug_register::DebugRegister;
+use bittide_sys::uart::Uart;
```
hydrolarus, line 12: Why not use the `bittide_hal::shared::devices::uart::Uart`? Less work for when we see what to keep from `bittide_sys` and what not (I guess the UART wouldn't be part of it though, so it would be (re)moved anyway)

martijnbastiaan: Oh you're saying UART is already automatically generated and I can boot out another file?

martijnbastiaan: I'll create an issue out of it. I'd like to get some help from you/Ryan to see how we can best do this :)

PR #821, opened 2025-06-07
Various Rust codegen fixes

### LINKED ISSUE #800: Can't generate HAL for tuples
I've got the Haskell type:

```haskell
  initWbT0 :: (Either (BitVector 8) (BitVector 16), Either (BitVector 8) (BitVector 16))
  initWbT0 = (Left 8, Right 8)
```

I get:

```
error: failed to run custom build command for `bittide-hal v0.1.0 (/home/martijn/code/bittide-hardware/firmware-support/bittide-hal)`

Caused by:
  process didn't exit successfully: `/home/martijn/code/bittide-hardware/firmware-support/../_build/cargo/firmware-support/debug/build/bittide-hal-4486307124a414b9/build-script-build` (exit status: 101)
  --- stdout
  cargo::rerun-if-changed=/home/martijn/code/bittide-hardware/firmware-support/bittide-hal/../../_build/memory_maps

  --- stderr

  thread 'main' panicked at memmap-generate/src/generators/mod.rs:28:5:
  Ident is not allowed to be empty; use Option<Ident>
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### LINKED ISSUE #817: `firmware-support/bittide-hal/src/shared/devices` should be cleared before building
I've renamed a device from `clock_control_debug` to `debug_register`, but the old generated code remains available. This leads to false positives when compiling code.

Image

### LINKED ISSUE #819: `Vec`s should probably also genreate a method yielding the whole vector
In clock control, we do:

```rust
    let measured_sum = cc.data_counts().sum::<i32>();
```

For the auto-generated code, this will have to change, because:

```rust
pub fn data_counts(&self, idx: usize) -> Option<u32>
```

This makes a lot of sense if you're only interested in a single element, but if you want to collect the whole thing it is a bit cumbersome. Perhaps we should also offer:

```rust
pub fn data_counts_iter(&self) -> impl Iterator<Item = u32>
```

hydrolarus:
Various changes and fixes to the Rust code generator, including more streamlined name handling.

fixes #817
fixes #800
fixes #819


martijnbastiaan:
Could you add a test for a six-tuple? I see up to 6 has a special code path. Does the generator handle the weird names for anything above that well?

PR #835, opened 2025-06-10
add annotation based config of code gen

### LINKED ISSUE #795: Generated types with floats in them can't be compiled due to `Eq` trait derivation
E.g.,

```haskell
data F = F { f :: Float }
  deriving (Generic, NFDataX, ShowX, Show, ToFieldType, BitPackC, BitPack)
```

generates

```rust
#![allow(unused_imports)]
#![allow(clippy::missing_safety_doc)]
#![allow(clippy::empty_docs)]
#![allow(clippy::unused_unit)]

pub use crate::shared::types::*;
#[repr(C)]
#[derive(Copy, Clone, PartialEq, Eq)]
pub struct F {
    pub f: f32,
}
```

This does not compile, because:

```
8   | #[derive(Copy, Clone, PartialEq, Eq)]
    |                                  -- in this derive macro expansion
9   | pub struct F {
10  |     pub f: f32,
    |     ^^^^^^^^^^ the trait `core::cmp::Eq` is not implemented for `f32`
    |
```

For now, I can work around it by removing `Eq`:

```diff
diff --git a/firmware-support/memmap-generate/src/generators/types.rs b/firmware-support/memmap-generate/src/generators/types.rs
index 71340964..ba41f1fb 100644
--- a/firmware-support/memmap-generate/src/generators/types.rs
+++ b/firmware-support/memmap-generate/src/generators/types.rs
@@ -90,13 +90,13 @@ impl TypeGenerator {
         let repr = self.generate_repr(ty);
         let derives = match debug {
             DebugDerive::None => quote! {
-                #[derive(Copy, Clone, PartialEq, Eq)]
+                #[derive(Copy, Clone, PartialEq)]
             },
             DebugDerive::Std => quote! {
-                #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+                #[derive(Debug, Copy, Clone, PartialEq)]
             },
             DebugDerive::Ufmt => quote! {
-                #[derive(uDebug, Copy, Clone, PartialEq, Eq)]
+                #[derive(uDebug, Copy, Clone, PartialEq)]
             },
         };
```

### LINKED ISSUE #822: More flexible code generation based on present types etc
Issues like #795 show that generating code isn't super straightforward, sometimes things like derives might need to be changed based on whether certain types are present or not.

A system that lets us query all sub-types and calculate some kind of feature-set would help with this. We'd have to figure out where the right place is to add those things. Could the same be used for a C codegen?

hydrolarus:
fixes #822
fixes #795

This adds a way to add annotations to types and devices. One usecase is to check for usage of `f32` or `f64` values and NOT deriving some traits based on that.

### DIFF
--- a/firmware-binaries/test-cases/registerwbc_test/src/main.rs
+++ b/firmware-binaries/test-cases/registerwbc_test/src/main.rs
@@ -104,8 +104,8 @@ fn main() -> ! {
     expect("init.v2[1]", Some([0x24, 0x32]), many_types.v2(1));
 
     // XXX: No uDebug trait for enums, so we can't use expect() here.
-    expect("init.sum0", true, hal::Abc::C == many_types.sum0());
-    expect("init.sum1", true, hal::Xyz::S == many_types.sum1());
+    expect("init.sum0", hal::Abc::C, many_types.sum0());
+    expect("init.sum1", hal::Xyz::S, many_types.sum1());
 
     expect("init.sop0.f", true, many_types.sop0().f == 3.14);
     expect("init.sop0.u", true, many_types.sop0().u == 6.28);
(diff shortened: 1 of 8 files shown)


martijnbastiaan (approved):
Gets us what we want, but I'm not a great fan of the hardcodedness of it. IMO a cleaner function would be one that collects all the features you want. In pseudo code:

```haskell
derive :: Type -> Set Derive
derive = \case
  Float | Double -> [PartialEq, ...]
  Int -> [Eq, PartialEq, ...]
  Product ts -> intersections (map derive ts)
  ...
```

If we ever add a constructor to the list this would get us a warning/error instead of it silently breaking. I.e., consider what happens when adding `Half`.

PR #841, opened 2025-06-14
Support for Index types, cleanup of test

### LINKED ISSUE #823: Add support for more complex types, like `Index x` to codegenerator
Right now, every read and write accessor creates a pointer, casts to the target type and performs the read/write. In the case of `Index n` we'd like to do some checks, so a raw pointer and read isn't good enough. The code generator needs to be adjusted to handle both "simple" and more complex types like this.

Also on the Rust side, `Index<n, ty>` is already generated but not defined anywhere. This needs to be done too!

hydrolarus:
fixes #823 

Turns out we don't need more complex loading/storing for this, but this adds support for Index types!


martijnbastiaan:
@rslawson Could you take a look too?

firmware-support/bittide-hal/src/manual_additions/index.rs
```diff
@@ -0,0 +1,436 @@
...
+            }
+        }
+
+        impl<const N: u128> Add for IndexTy<N, $t> {
+            type Output = Self;
+
+            fn add(self, rhs: Self) -> Self::Output {
+                let res = self.0 + rhs.0;
+                debug_assert!(res < N as $t, "Addition overflowed");
+                Self(res)
```
rslawson, line 108: Doesn't this lead to some potentially weird behaviour on release? Should the behaviour there be wrapping by default, or are you happy with this implementation as-is?

firmware-support/bittide-hal/src/manual_additions/index.rs
```diff
@@ -0,0 +1,436 @@
...
+            }
+        }
+
+        impl<const N: u128> Sub for IndexTy<N, $t> {
+            type Output = Self;
+
+            fn sub(self, rhs: Self) -> Self::Output {
+                let res = self.0 - rhs.0;
+                debug_assert!(res < N as $t);
+                Self(res)
```
rslawson, line 177: Same behaviour question here for release mode.

firmware-support/bittide-hal/src/manual_additions/index.rs
```diff
@@ -0,0 +1,436 @@
...
+impl_add!(u128);
+
+macro_rules! impl_sub {
+    ($t:ty) => {
+        impl<const N: u128> IndexTy<N, $t> {
+            pub fn saturating_sub(self, rhs: Self) -> Self {
+                let res = self.0.saturating_sub(rhs.0);
+                Self(res)
+            }
+        }
```
rslawson, line 169: No checked subtraction?

firmware-support/bittide-hal/src/manual_additions/index.rs
```diff
@@ -0,0 +1,436 @@
...
+impl_sub!(u128);
+
+macro_rules! impl_mul {
+    ($t:ty) => {
+        impl<const N: u128> IndexTy<N, $t> {
+            pub fn saturating_mul(self, rhs: Self) -> Self {
+                let res = self.0.saturating_mul(rhs.0);
+                Self(res.min(N as $t))
+            }
+        }
```
rslawson, line 238: No checked mul?

firmware-support/bittide-hal/src/manual_additions/index.rs
```diff
@@ -0,0 +1,436 @@
...
+
+        impl<const N: u128> Mul<&IndexTy<N, $t>> for IndexTy<N, $t> {
+            type Output = Self;
+
+            fn mul(self, rhs: &Self) -> Self::Output {
+                let res = self.0 - rhs.0;
+                debug_assert!(res < N as $t, "Multiplication overflowed");
+                Self(res)
+            }
+        }
```
rslawson, line 258: Implementations here do subtraction, not multiplication.

hydrolarus: Good catch!

PR #850, opened 2025-06-23
Move switch demo to general purpose processing element

rslawson:
I am once again PRing a big diff. This time with fun additions!

- All processing elements now have a non-optional identifier component. This can be ignored, but it is intended to be used as a way to guarantee that you're looking at the correct CPU in a JTAG chain.
- Some helper functions for creating these identifiers and search strings for GDB output have been added
- A number of GADTs were updated to use record syntax. I didn't like not having it be explicit what things were.
- Significant rewrite of `Node.hs`. Should be a lot simpler to work with now.
- Added a utility function for prefixes. This is primarily intended to be used for the prefix of the identifier component, where it's expected to have a minimum address width that may need to get widened with 0s to fit other processing elements' prefix widths. Usage looks like
```hs
whoAmIPfx :: forall n. (KnownNat n) => Unsigned (3 + n)
whoAmIPfx = extendLsb0s @3 @n 0b111
-- This prefix can now be used in arbitrary locations expecting
-- a prefix with a width of 3 bits or more.
```
- Added a binary for the switch demo GPPE (WIP)

### DIFF
--- a/.github/synthesis/debug.json
+++ b/.github/synthesis/debug.json
@@ -1,0 +1,3 @@
+[
+  {"top": "switchDemoTest", "stage": "test"}
+]
--- a/bittide-extra/src/Protocols/Extra.hs
+++ b/bittide-extra/src/Protocols/Extra.hs
@@ -21,10 +21,18 @@ mapCircuit ::
   Circuit (Vec n a) (Vec n b)
 mapCircuit circ = Circuit go
  where
-  circuitFn :: (Fwd a, Bwd b) -> (Bwd a, Fwd b)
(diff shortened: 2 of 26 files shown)


martijnbastiaan:
As it currently stands, this PR can't be reviewed. From your PR cover letter, it seems like there are at least 5 logical chunks of work you did and this should be reflected in your commit history. For example, things like:

> Significant rewrite of Node.hs. Should be a lot simpler to work with now.

..should _never_ be mixed with other commits. For a reviewer it's really unclear what changes you made as a refactor and what changes you made because you wanted to move to a GPPE.

PR #852, opened 2025-07-01
split BitPackC and procols-memmap

### LINKED ISSUE #804: Split off `BitPackC` to own package (and rename)
After merging https://github.com/bittide/bittide-hardware/pull/801 I think we should split off `BitPackC` to its own package. We can keep it here in this repository for a while, but it should make its way into the wider ecosystem at some point. We should also rename the module `BitPackC` to `Clash.Class.BitPackC` to fit in with what `clash-prelude` does.

hydrolarus:
fixes #804


martijnbastiaan (approved):
Noice.

Could you add `clash-bitpackc` to:

https://github.com/bittide/bittide-hardware/blob/d6cc802a39de19d2bf19f5ad4bcfe6816903509e/.github/workflows/ci.yml#L244-L250

and:

https://github.com/bittide/bittide-hardware/blob/d6cc802a39de19d2bf19f5ad4bcfe6816903509e/.github/scripts/check_consistent_cabal_common.py#L136-L142

PR #860, opened 2025-07-02
Add GPPE calendar calculator

rslawson:
This also replaces the use of the previous switch demo processing element calculator. Instead, calendars are calculated as though the system were operating on metacycles with double-buffered memory for the scatter/gather engines. These calendars are the converted back into the SDPE calendars and use for testing until the actual GPPE implementation is ready.


bittide/src/Bittide/Calculator.hs
```diff
@@ -0,0 +1,571 @@
...
+  CalendarProps (CalendarCalculator n c p r)
+  where
+  type CalNumNodes (CalendarCalculator n c p r) = n
+  type CalCyclesPerWrite (CalendarCalculator n c p r) = c
+  type CalPadding (CalendarCalculator n c p r) = p
+  type CalRepetitions (CalendarCalculator n c p r) = r
+  type CalGroupCycles (CalendarCalculator n c p r) = GroupCycles n c
+  type CalWindowCycles (CalendarCalculator n c p r) = WindowCycles n c
+  type CalActiveCycles (CalendarCalculator n c p r) = ActiveCycles n c r
+  type CalMetacycleLength (CalendarCalculator n c p r) = MetacycleLength n c p r
```
martijnbastiaan, line 181: Why do we need this on the type level?

bittide/src/Bittide/Calculator.hs
```diff
@@ -0,0 +1,571 @@
...
+  findK :: Index nNodes -> Index nNodes -> Integer -> Integer
+  findK src dst srcCycle
+    | offsetInDstMetacycle <= linkSelectOffset0 = linkSelectOffset0 - offsetInDstMetacycle
+    | offsetInDstMetacycle <= linkSelectOffset1 = linkSelectOffset1 - offsetInDstMetacycle
+    | otherwise = linkSelectOffset0 + metacycleLength - offsetInDstMetacycle
+   where
+    -- For a graphical representation of what we're doing here, check out:
+    -- bittide-instances/imgs/ugn_calculator.drawio. For a general introduction
+    -- see slides "determining read/write cycles" in
+    -- https://docs.google.com/presentation/d/1AGbAJQ1zhTPtrekKnQcthd0TUPyQs-zowQpV1ux4k-Y
```
martijnbastiaan, line 464: Could you update the drawio diagram? Could you copy the presentation and create an updated one?

bittide/src/Bittide/Calculator.hs
```diff
@@ -0,0 +1,571 @@
...
+import Data.Maybe (catMaybes, fromJust)
+import Data.String.Interpolate (i)
+import GHC.Stack (HasCallStack)
+
+import qualified Data.List as L
+import qualified Data.Map as Map
+
+type FpgaId = String
+
+-- TODO: find a way to derive this 4 from the code instead of using a magic number.
```
martijnbastiaan, line 28:
```suggestion
-- TODO: find a way to derive this 4 from the code instead of using a magic number.
--
-- This is currently '4', because we traverse two switches and each switch has registered
-- inputs and outputs.
```

bittide/src/Bittide/Calculator.hs
rslawson: A lot of these are for convenience elsewhere. I also do it at the type level because in some places I need the metacycle length at type level (e.g. for the `Index (ActiveCycles nNodes cyclesPerWrite repetitions)` in `DefaultMetaPeConfig`). Plus I wanted to do these calculations only in one place so there was only one place I could make mistakes, so I just lower the types to values where needed.

martijnbastiaan (approved):
Could you also delete `bittide/imgs/ugn_calculator.drawio.license`?

Issue #845, opened 2025-06-19
Add UART multiplexing


martijnbastiaan:
We use UART for debugging information while running a system. Most bittide instances now consist of multiple CPUs, but we still only have one physical UART port. This leaves us with only one CPU being able to output logging over UART, which is undesirable.

I've discussed this with @lmbollen, our conclusion is that the mux should take care to only interleave whole "messages". A message could be either on of:

1. A bunch of characters followed by an `\n`
2. A bunch of characters followed by a `ETX` / `EOT`

While the second option gives makes sure you can write multiline strings without them being intermingled with other CPUs's output, it would need extra software support and care from developers to mark their transmissions. My proposal would therefore be KISS and pick (1).

The mux should also take care to prefix outputs with their source. E.g., output could look like:

```
[CPU0] hello
[CPU1] world
[CPU2] !
```

where `CPU0`, `CPU1`, and `CPU2` are user configurable strings when instantiating the mux.

Lastly, the mux should be able to time out and forcefully insert a `\n`.

-------

Schematic proposal for implementation:

![Image](https://github.com/user-attachments/assets/1f45fe25-b62d-4cb1-8fae-30c5e3c06e54)

martijnbastiaan:
I've assigned this to myself and dragged it to TODO on the Kanban board, because I'm stuck debugging an issue for https://github.com/bittide/bittide-hardware/issues/741.

PR #862, opened 2025-07-03
Add `asciiDebugMux`

martijnbastiaan:
Part of #845

### DIFF
--- a/bittide/src/Bittide/Df.hs
+++ b/bittide/src/Bittide/Df.hs
@@ -1,0 +1,129 @@
+-- SPDX-FileCopyrightText: 2025 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE OverloadedRecordDot #-}
+
+module Bittide.Df (asciiDebugMux) where
+
+import Clash.Prelude
+import Protocols
+
+import Bittide.PacketStream (timeout)
+import Bittide.SharedTypes (Byte)
+import Data.Char (ord)
+import Data.Coerce (coerce)
+import Data.Functor ((<&>))
+import Protocols.Df (CollectMode (Skip))
+import Protocols.PacketStream (
+  PacketStream,
+  PacketStreamM2S (..),
+  PacketStreamS2M (..),
+  packetizerC,
+ )
+import Protocols.PacketStream.Base (truncateAbortedPackets)
+import Protocols.PacketStream.Routing (packetArbiterC)
+import Protocols.Vec (vecCircuits)
+
+{- | Muxes multiple 'Df' streams of bytes into a single 'Df' stream of bytes,
+interleaving the streams from the different sources on newline characters. As
+an example, if two streams are supplied trying to write the messages @Hello\n@
+and @World\n@, respectively, the output will be:
+
+> [LBL0] Hello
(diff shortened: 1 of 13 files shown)


bittide/src/Bittide/Df.hs
```diff
@@ -0,0 +1,132 @@
...
+  ) =>
+  (meta -> Vec n Byte) ->
+  Circuit (PacketStream dom 1 meta) (Df dom Byte)
+unsafeBytesFromPacketStream f =
+  packetizerC (const ()) f |> unsafePacketStreamToDf
+ where
+  unsafePacketStreamToDf = Circuit (unbundle . fmap go . bundle)
+   where
+    go (m2s, ack)
+      | Just PacketStreamM2S{_abort = True} <- m2s = (coerce True, Nothing)
```
lmbollen, line 126: The `PacketStream` protocol does not require the `_abort` signal to stay asserted during the remainder of the packet, if you drop when `_abort` is set, you might just drop a single transfer. I don't think this is what you want.

Given that it is already unsafe, I think it would be better for this component to ignore the `_abort` signal and we should add a component that:
* When `_abort` is set, it truncates the packet by setting `_last` to `Just 0`.
* Consumes all following transactions up to the actual end of the packet without forwarding them.

lmbollen (approved):
Some small remarks and suggestions

bittide/src/Bittide/Df.hs
lmbollen: See https://github.com/clash-lang/clash-protocols/pull/159 :)

martijnbastiaan: Thanks, I'll add it here. I'll postpone upstreaming `timeout` a bit, because I would need to figure out how to write tests for it..

bittide/src/Bittide/Df.hs
```diff
@@ -0,0 +1,132 @@
...
+      | Just PacketStreamM2S{_last = Just 0} <- m2s = (coerce True, Nothing)
+      | otherwise = (coerce ack, fmap (pack . (._data)) m2s)
+
+-- | What it says on the tin. Maybe we should upstream this to @clash-protocols@?
+bundleC :: forall n a b. Vec n (Circuit a b) -> Circuit (Vec n a) (Vec n b)
+bundleC circuits = Circuit (unzip . zipWith ($) (map toSignals circuits) . uncurry zip)
```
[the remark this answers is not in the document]
martijnbastiaan: Yes, `bundleC` is the better name! But thanks for the hint, I'll remove it.

bittide/src/Bittide/Df.hs
```diff
@@ -0,0 +1,132 @@
...
+        , _abort = False
+        }
+
+  -- Characters needed to format the an output to "[${label}] ${message}\n"
+  (newline, openBracket, closeBracket, space) =
+    ( fromIntegral $ ord '\n'
+    , fromIntegral $ ord '['
+    , fromIntegral $ ord ']'
+    , fromIntegral $ ord ' '
+    )
```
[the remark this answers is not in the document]
martijnbastiaan: Clash won't translate that to HDL due to the operators on `[]` :(

PR #864, opened 2025-07-07
Rename constant in calculator to describe better what it actually does.

rslawson:
(:


martijnbastiaan:
Woohooo! :tada:

Could you amend your commit to look like:

```
Rename `dELAY_FROM_EB_TO_CROSSBAR` to `dELAY_CROSSBAR_TO_PE` in calculator

As discussed in #753, this turns out to be the correct name.

Fixes #753
```

? In general I think commit titles should be as specific as they can be (without becoming paragraphs) and the commit body should provide context. The footer last but not least can refer (and cause auto-closing) issues. ([Here is a blog post I really like on the topic.](https://mtlynch.io/no-longer-my-favorite-git-commit/))

martijnbastiaan (approved):

PR #865, opened 2025-07-08
add metaPeConfig device

### LINKED ISSUE #854: Add GPPE switch demo configuration device
From https://github.com/bittide/bittide-hardware/issues/755#issuecomment-3022897859: we need a device that is memory mapped that can store configuration data. I.e., it should store:

```haskell
data MetaPeConfig a b = MetaPeConfig
  { writeMetacycle :: a
  , writeOffset :: b
  , writeForN :: b

  , readMetacycle :: a
  , readOffset :: b
  , readForN :: b
  }
  deriving (Functor, Eq, Show)

-- The device can use the monomorphic `MetaPeConfig (Unsigned 32) (Unsigned 32)`. To keep
-- the GDB interaction easy enough, let's expose the fields as separate registers.
```

Plus, it should have a small buffer to store results like the current [SwitchDemoProcessingElement](https://github.com/bittide/bittide-hardware/blob/c7f25fedd149550d8b7f74b690ea73e213b1bb94/bittide/src/Bittide/SwitchDemoProcessingElement.hs#L54-L55). Inspiration on how to write a memory mapped component can be taken from [Freeze.hs](https://github.com/bittide/bittide-hardware/blob/b033cfce6da0b8e798b6027229e06c3f49338035/bittide/src/Bittide/ClockControl/Freeze.hs).

hydrolarus:
fixes #854

### DIFF
--- a/bittide/bittide.cabal
+++ b/bittide/bittide.cabal
@@ -209,6 +209,7 @@ library
     Bittide.Switch
     Bittide.SwitchDemoProcessingElement
     Bittide.SwitchDemoProcessingElement.Calculator
+    Bittide.SwitchDemoProcessingElement.MetaPeConfig
     Bittide.Sync
     Bittide.Transceiver
     Bittide.Transceiver.Cdc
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
I'm not convinced that it needs to be as generic as it is especially if that introduces `AllowAmbiguousTypes`, could you just slap `Unsigned 32` on everything?

PR #866, opened 2025-07-08
Move application of crossbar to PE to delay in calendar calculation

rslawson:
I believe that this better reflects the new understanding we have come to of where delays come from within the system.

### DIFF
--- a/bittide/src/Bittide/SwitchDemoProcessingElement/Calculator.hs
+++ b/bittide/src/Bittide/SwitchDemoProcessingElement/Calculator.hs
@@ -210,19 +210,19 @@ chainConfigurationWorker fpgaConfig ugnParts writeOffset cyclesPerWrite =
 
   findK :: Index nNodes -> Index nNodes -> Integer -> Integer
   findK src dst srcCycle
-    | dstCycle <= linkStart0 = linkStart0 - dstCycle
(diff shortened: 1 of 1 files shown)


bittide/src/Bittide/SwitchDemoProcessingElement/Calculator.hs
```diff
@@ -218,11 +218,11 @@ chainConfigurationWorker fpgaConfig ugnParts writeOffset cyclesPerWrite =
     -- bittide-instances/imgs/ugn_calculator.drawio. For a general introduction
     -- see slides "determining read/write cycles" in
     -- https://docs.google.com/presentation/d/1AGbAJQ1zhTPtrekKnQcthd0TUPyQs-zowQpV1ux4k-Y
-    dstCycle = mapCycle srcCycle src dst
+    dstCycle = mapCycle srcCycle src dst - dELAY_CROSSBAR_TO_PE
```
martijnbastiaan, line 221:
```suggestion
    crossbarDstCycle = mapCycle srcCycle src dst - dELAY_CROSSBAR_TO_PE
```

Let's make it more obvious what we're mapping to

martijnbastiaan (approved):
LGTM, but let's reflect what we're mapping to in the variable names too.

PR #874, opened 2025-07-13
Add UART debug logging in switch demo to both CC and MU processor

martijnbastiaan:
Example output in the picocom logs:

```
[..]
[CC] Starting clock control..
[MU] Hello from management unit..
```

This is quite a lot, but I've tried to make it as reviewable as possible by splitting it up in many self-contained commits.

The plan for clock control is to remove the separate `-swcctopologies` clock control binary in time (but not this PR).


bittide-instances/src/Bittide/Instances/Hitl/Driver/SwitchDemo.hs
```diff
@@ -475,82 +512,88 @@ driver testName targets = do
...
+            [i|MU GDB testing passed on #{gdbCount1} of #{L.length targets} targets|]
+        liftIO $ mapM_ ((errorToException =<<) . Gdb.loadBinary) muGdbs
+
+        let picocomStarts = liftIO <$> L.zipWith (initPicocom) targets [0 ..]
+        brackets picocomStarts (liftIO . snd) $ \_picocoms -> do
+          liftIO $ mapM_ Gdb.continue ccGdbs
+          -- XXX: If you also want to start the management units, uncomment this
+          --      next line.Beware that this breaks the demo, because we want to
+          --      take over control of the management unit again to read out
+          --      various values, but we don't have a way to pause the CPU again.
```
hiddemoll, line 549:
```suggestion
          -- XXX: If you also want to start the management units, uncomment this
          --      next line. Beware that this breaks the demo, because we want
          --      to take over control of the management unit again to read out
          --      various values, but we don't have a way to pause the CPU again.
```

hiddemoll (approved):
Great job in splitting up your work! I see no issues in this PR (except for a missing space in a comment). 
I think @rslawson should also review this PR.

PR #876, opened 2025-07-15
conditionally run clock control in simulation

### LINKED ISSUE #869: Add simulation level flag to skip clock control
We are working towards a cycle accurate simulator for Bittide systems. Real systems can take multiple seconds (~5-10) of wall clock time to negotiate links and settle their clock frequencies. Given that we will want to simulate multiple Bittide nodes, this would make simulation take unreasonably long. To make simulation easier, we want to abstract over a few systems, one of which is clock control.

To do so, there should be a way to override clock control to always claim it is done with frequency alignment.

hydrolarus:
fixes #869 

Is this good enough? Should this be a parameter to the whole circuit instead?


martijnbastiaan:
I think this makes sense. We can pull it out when we need to.

bittide-instances/src/Bittide/Instances/Hitl/SwitchDemo.hs
```diff
@@ -81,6 +81,7 @@ import Protocols.Wishbone
 import System.FilePath ((</>))
 import VexRiscv (DumpVcd (..), Jtag, JtagIn (..), JtagOut (..))
 
+import Bittide.ClockControl.StabilityChecker (StabilityIndication (..))
```
martijnbastiaan, line 84: Hopefully fourmolu will support it soon, but: `qualified` imports go last as a group. Could you move this import up?

martijnbastiaan (approved):

Issue #741, opened 2025-04-01
Refactor/simplify clock control to store results in non-ILA destination


martijnbastiaan:
Over time we've switched from a hardware based clock control implementation to a software defined one. In between, we've maintained both. Our debug infrastructure (i.e., the infra storing plot data) has grown to support both. We've now outgrown this infrastructure in multiple ways:

- We now have control over the hardware from Haskell, both through GDB and `vivado-hs`. This eliminates the need to define test cases in hardware.
- We can now easily query _any_ memory mapped address of _any_ CPU in the system, eliminating the need to use ILAs to store data (together with `callistoFreeze`).
- New insights have made us realize that software based clock control (even on our comparatively slow CPUs) is more than quick enough, ditched the need for a hardware implementation.

Hence, we're in a perfect position to move everything to software, only implementing hardware where we really need hardware acceleration. I propose the following:

![Image](https://github.com/user-attachments/assets/3c0763d7-e827-46da-bb61-313bc48a3fd8)

As a bonus, this strategy would eliminate the need to reset the CPU while running the test bench (like we do now), removing the last road block to upgrade to upstream OpenOCD.

PR #883, opened 2025-07-16
Store clock control data in SW accessible RAM

martijnbastiaan:
Closes https://github.com/bittide/bittide-hardware/issues/741

# TODO
- [ ] Dehardcode switch demo with memory map
- [ ] Extract CC data using GDB
- [ ] Feed extracted data to plot infra


firmware-binaries/clock-control/src/main.rs
```diff
@@ -43,14 +50,20 @@ fn main() -> ! {
         ReframingState::Detect,
     );
 
-    // Update clock control 10K updates per second
-    let interval = Duration::from_micros(100);
+    // Update clock control 50K updates per second
+    let interval = Duration::from_micros(50);
```
martijnbastiaan, line 54:
```suggestion
    let interval = Duration::from_micros(20);
```

bittide-instances/src/Bittide/Instances/Hitl/Utils/Gdb.hs
```diff
@@ -95,32 +118,55 @@ compareSections gdb = do
...
       , "set logging enabled on"
       ]
 
+dumpMemoryRegion ::
+  (HasCallStack) =>
+  ProcessHandles ->
+  FilePath ->
+  Integer ->
+  Integer ->
+  IO ()
```
martijnbastiaan, line 135: Add documentation

firmware-support/bittide-sys/src/sample_store.rs
```diff
@@ -0,0 +1,81 @@
...
+            store_samples_every,
+            counter: 0,
+        }
+    }
+
+    /// *Actually* store the contents of 'Freeze' to memory. Note that the public
+    /// function 'store' does 'store_samples_every' boundary checking.
+    fn do_store(&mut self, freeze: &bittide_hal::freeze::Freeze, bump_counter: bool) {
+        let n_samples_stored = self.memory.data(0).unwrap() as usize;
+        let start_index = n_samples_stored * WORDS_PER_SAMPLE + 1;
```
martijnbastiaan, line 32: In retrospect: could we have had a block ram that has a fancy record type instead?

martijnbastiaan (commented):
*

firmware-support/bittide-sys/src/sample_store.rs
martijnbastiaan: Not really.. I'll write up an issue.

PR #885, opened 2025-07-17
Wishbone to  AXI4 converter

lmbollen:
We want to be able to talk to our AXI4 DDR primitive using our Wishbone based RV32 CPU


martijnbastiaan (dismissed):
All public functions/types need documentation, but other than that it looks good to me :).

PR #894, opened 2025-07-21
Replace OpenOCD with `probe-rs`

rslawson:
Longer-term goal and not currently in active development, but this PR aims to achieve dynamic GDB port assignment for CPUs based on an identifier component within each CPU. Currently in the project, OpenOCD is used to open debugging targets on given CPUs at points in the JTAG chain, and then we check whether the identifier component reported by the CPU at each TAP is the one we expect. The rationale of this PR is that this should be done the other way - TAPs should be inspected for their identifiers, and this should be used to open a debugging target on the appropriate port number.

The current status of development in this is that `probe-rs` is properly able to connect to the CPUs when used through the CLI, but the tool that uses it as a library is not. As such, some investigation into how `probe-rs` opens up GDB sockets on targets should be done and the approach mirrored as is appropriate for this application. See also: [#3402](https://github.com/probe-rs/probe-rs/pull/3402).


martijnbastiaan:
It's unlikely that we'll pick this up (soon). Let's resurrect when we get there.

PR #901, opened 2025-07-22
Use `sane-records` in packages checked by Cabal header check

martijnbastiaan:


martijnbastiaan:
@lmbollen Done, also see https://github.com/bittide/bittide-hardware/issues/902.

PR #913, opened 2025-07-23
Use `deepErrorX` in `getConstBwdAny`.

lmbollen:
Previously we'd receive an XException when evaluating the spine of the undefined signals.


martijnbastiaan (approved):
Please drop the dot in the commit title. Otherwise LGTM :)

PR #914, opened 2025-07-24
Return `WaitResult` for absolute wait functions

martijnbastiaan:
Useful for detecting missed deadlines

### DIFF
--- a/firmware-support/bittide-hal/src/manual_additions/timer.rs
+++ b/firmware-support/bittide-hal/src/manual_additions/timer.rs
@@ -26,6 +26,15 @@ use ufmt::uwrite;
 
 pub mod self_test;
 
+#[derive(uDebug, Debug)]
+pub enum WaitResult {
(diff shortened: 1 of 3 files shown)


hydrolarus (approved):

firmware-support/bittide-hal/src/manual_additions/timer.rs
```diff
@@ -26,6 +26,14 @@ use ufmt::uwrite;
 
 pub mod self_test;
 
+pub enum WaitResult {
```
hydrolarus, line 29: This could have `uDebug`/`Debug` derives, might be nice for logging.
```suggestion
use ufmt::derive::uDebug;

#[derive(uDebug, Debug)]
pub enum WaitResult {
```

hydrolarus (approved):

hydrolarus (approved):

PR #919, opened 2025-07-24
Refactor calendar wishbone interface

lmbollen:
The older interface included a handwritten memory map, was never test and turned out to be wrong.

In this PR I refactor the wishbone interface for the calendar using the new clash-memmap infrastructure.
This includes a unit test that tests:
* Swapping the calendars
* Waiting for the end of the metacycle
* Obtaining the metacycle count
* Reading calendars
* Writing calendars


bittide/src/Bittide/Calendar.hs
```diff
@@ -58,7 +71,7 @@ data ValidEntry a repetitionBits = ValidEntry
   -- 0 = no repetition, entry is valid for one cycle.
   -- 1 = repeated once, entry is valid for two cycles.
   }
-  deriving (BitPack, Eq, Generic, NFDataX, Show, ShowX)
+  deriving (BitPack, Eq, Generic, NFDataX, Show, ShowX, BitPackC, ToFieldType)
```
hydrolarus, line 74: Types that have type arguments need a separate instance for `ToFieldType`, otherwise the arguments will be "inlined" and monomorphised but under the same name. With a separate instance for `ToFieldType` it can generate a Rust type that's also generic!

That said, `repetitionBits` is a `Nat`, that's currently not supported, but I could add support for that in `ToFieldType`. I'll open an issue.

clash-protocols-memmap/src/Protocols/MemoryMap/Registers/WishboneStandard.hs
```diff
@@ -78,6 +79,11 @@ import qualified Protocols.Vec as V
 data BusActivity a = BusIdle | BusRead a | BusWrite a
   deriving (Show, Eq, Functor, Generic, NFDataX)
 
+busActivityWrite :: BusActivity a -> Maybe a
```
hydrolarus, line 82: Small documentation comment would be nice

firmware-binaries/test-cases/switch_calendar_test/build.rs
```diff
@@ -0,0 +1,37 @@
...
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("../../../")
+        .join("_build")
+        .join("memory_maps")
+}
+
+/// Put the linker script somewhere the linker can find it.
+fn main() {
+    let memory_x = memory_x_from_memmap(
+        memmap_dir().join("RegisterWbC.json"),
```
hydrolarus, line 21: Is this right?

firmware-binaries/test-cases/switch_calendar_test/src/main.rs
```diff
@@ -0,0 +1,165 @@
...
+        // Uncomment below to print progress messages
+        // } else {
+        //     writeln!(INSTANCES.uart, "PASS: {} | {:?}", msg, expected).unwrap();
+    }
+}
+
+#[cfg_attr(not(test), entry)]
+fn main() -> ! {
+    let active_entry0: [Index![17]; 16] =
+        core::array::from_fn(|i| unsafe { IndexTy::new_unchecked(i as u8) });
```
hydrolarus, line 53: Is the performance that much worse if you used `index!(i, n = 17)` instead? It's a test case anyway, so doing extra checks shouldn't be a problem.

Same below when IndexTy is used (it shouldn't really be used, only `Index![]` and `index!()`)

firmware-binaries/test-cases/switch_calendar_test/src/main.rs
```diff
@@ -0,0 +1,165 @@
...
+use bittide_hal::hals::switch_c as hal;
+use core::fmt::Write;
+#[cfg(not(test))]
+use riscv_rt::entry;
+
+const INSTANCES: hal::DeviceInstances = unsafe { hal::DeviceInstances::new() };
+
+fn test_result(result: &str) -> ! {
+    let uart = &mut INSTANCES.uart;
+    writeln!(uart, "RESULT: {}", result).unwrap();
```
hydrolarus, line 24: `uwritln!` could help reduce binary size and should be preferred whenever possible

firmware-binaries/test-cases/switch_calendar_test/src/main.rs
```diff
@@ -0,0 +1,165 @@
...
+    loop {}
+}
+
+fn test_ok() -> ! {
+    test_result("OK")
+}
+
+fn test_fail(msg: &str) -> ! {
+    let mut full_msg = heapless::String::<1024>::new();
+    let _ = write!(full_msg, "FAIL: {}", msg);
```
hydrolarus, line 34: same here (`uwrite!`)

rslawson:
For the most part, yes. There's one or two more extra things that I'll do in another PR.

PR #899, opened 2025-07-22
Move stability checking to SW

martijnbastiaan:

PR #930, opened 2025-08-04
Remove ILA sample storage in favor of SW RAM storage

martijnbastiaan:
Commit message from main commit:

> In https://github.com/bittide/bittide-hardware/pull/899 we introduced a way of storing clock control results in (block) RAM. In that PR, we add it to the switch demo, which is a DUT that closely resembles what we think a "real" bittide system might look like. Due to difficulties with tying the "new" way together with the "old" way, it leaves the pre-existing method for what it is. This has left us with a split in how we store clock control data. This commit completely removes the "old" way of doing things -- in a regrettably large diff.
>
> To summarize:
> 
> * It unifies `SwCcTopologies` and `SwitchDemo`, the former now being an   alias of the latter. In the future we should update our test infrastructure such that a single DUT can have multiple test suites associated with it.
> 
> * It moves all parsing to `Bittide.Instances.Hitl.Driver.Clock.Control`   and plotting logic a directory below (`Plot`). A single script   `bittide-instances:samples-to` has been added to deal with the data   dumps generated by GDB. It can create CSVs, individual clock and   elastic buffer occupancy plots, and reports.
> 
> * It simplifies CI by storing what jobs generate data suitable for clock   control reports in the JSON.
> 
> Regrettably, it also removes a bunch of features that the previous topology tests used to have: topologies, clock offsets, and startup delays. The plan is to reintroduce these in later PRs.
> 
> Lastly, the plots make it obvious that the current design of the switch demo is lacking in some ways:
> 
>  * Domain difference counters start as soon as all links are stable,    which means clock control will never see them starting at zero. This    isn't a problem in practice as clock control is a simple    P-controller, but it makes the plots ugly / counter-intuitive.
> 
>  * Clock control and the associated debug infrastructure runs in the    _free_ domain. This means that the clock plots show measurements of    _uncontrolled_ clocks, which is obviously not what we want.


bittide-instances/src/Bittide/Instances/Hitl/Driver/ClockControl/Samples.hs
```diff
@@ -0,0 +1,148 @@
...
+
+instance Csv.DefaultOrdered Sample where
+  headerOrder = Csv.header . fmap fst . toNamedRecord#
+
+-- | Calculate the number of picoseconds from a sample
+sampleToPicoseconds :: Sample -> Picoseconds
+sampleToPicoseconds sample = seen1 * psPerSync + since1 * psPerCycle
+ where
+  psPerCycle = C.natToNum @(C.DomainPeriod Bittide)
+  psPerSync = C.natToNum @SyncOutGeneratorHalfPeriod
```
hiddemoll, line 139: Is `cyclesSinceSyncPulse` reset for every edge of the sync pulse? If not you probably need  `2*SyncOutGeneratorHalfPeriod`, right?

bittide-instances/bin/samples_to.hs
```diff
@@ -0,0 +1,154 @@
+-- SPDX-FileCopyrightText: 2025 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE QuasiQuotes #-}
+
+module Main where
+
```
hiddemoll, line 7: This module should have a module description, since its an executable.

hiddemoll (approved):
In general I think this work greatly improves the readablility and reusablitlity of the plotting infrastructure, so great work! 

I have very little comments for how large the diff is, most of them are questions rather than change requests. If you want we can also discuss these in the office.

bittide-instances/src/Bittide/Instances/Hitl/Driver/ClockControl/Samples.hs
martijnbastiaan: https://github.com/bittide/bittide-hardware/issues/939

PR #951, opened 2025-08-11
Various general CC cleanups to prepare to move ILA features to SW

martijnbastiaan:
This PR consists of a few general cleanups:

* Removing obsolete code
* Small refactor of `callisto.rs` to use Rust's `impl` instead of standalone functions
* Move clock control configuration to a single (synthesizable) hardware register

The end goal is that we'll have a single register (with sensible defaults) that configures everything clock control should know (link availability, callisto settings, ...).


hiddemoll (approved):
Nice work, I have no remarks at all.

PR #954, opened 2025-08-12
Move GDB utilities into own package `gdb-hs`

martijnbastiaan:
It also restructures the functions in `Gdb.hs` a bit such that there is only one place where we do stdin/stdout logic, instead of all over the place. This benefit can mostly be seen in the diff of `SwitchDemo`. It now also offers `withGdbs`: a safe way to deal with multiple GDB processes. Previous approach would leave it up to the user to write exception-free initialization code.

# TODO
- [x] Add basic test


gdb-hs/src/Gdb.hs
```diff
@@ -0,0 +1,288 @@
...
+-}
+readCommand :: (HasCallStack) => Gdb -> String -> IO [String]
+readCommand gdb cmd = do
+  output <- readCommandRaw gdb cmd
+  pure $
+    L.dropWhile (== "") $
+      L.dropWhileEnd (== "") $
+        L.map L.trimEnd $
+          L.lines $
+            output
```
rslawson, line 142:
```suggestion
  pure $
    L.dropWhile (== "") $
      L.dropWhileEnd (== "") $
        L.map L.trimEnd $
          L.lines output
```
or
```suggestion
  pure $ L.dropWhile (== "") $ L.dropWhileEnd (== "") $ L.map L.trimEnd $ L.lines output
```
maybe? Neither of these is particularly important, sort of nitpicky.

rslawson (changes requested):

gdb-hs/src/Gdb.hs
martijnbastiaan: Let me offer you a third option:

```haskell
  pure
    . L.dropWhile (== "")
    . L.dropWhileEnd (== "")
    . L.map L.trimEnd
    . L.lines
    $ output
```

Also the whole reason Fourmolu formats like this is because it differentiates between `Clash.Prelude`'s `$` (which it doesn't recognize) and `Prelude`'s `$` (which it does). Which is.. exhausting.

Which one do you prefer?

rslawson: Last one, that looks pretty nice to me.

martijnbastiaan:
Thanks @rslawson, applied the suggestions.

rslawson (approved):

PR #959, opened 2025-08-13
Add `Gdb.Commands.Read`

martijnbastiaan:
Contains utilities for reading memory wrapped in `Vec` / `Bytes`. It also adds functions to read arbitrary Haskell types from memory -- provided that they are laid out according to the C FFI.


hiddemoll (approved):
This is awesome!

PR #962, opened 2025-08-13
Simplify `SwitchDemo` driver by using `Gdb.Command.Read`

martijnbastiaan:


bittide-instances/src/Bittide/Instances/Hitl/Driver/SwitchDemo.hs
```diff
@@ -419,44 +350,22 @@ driver testName targets = do
...
-            expectedCounter = showHex myCfg.startWriteAt ""
-            counterEq = hexEq myCounter expectedCounter
-            dna0Eq = hexEq myDeviceDna0 headDeviceDna0
-            dna1Eq = hexEq myDeviceDna1 headDeviceDna1
-            dna2Eq = hexEq myDeviceDna2 headDeviceDna2
+            expectedCounter = myCfg.startWriteAt
+            counterEq = myCounter == expectedCounter
+            dna0Eq = myDeviceDna0 == headDeviceDna0
+            dna1Eq = myDeviceDna1 == headDeviceDna1
+            dna2Eq = myDeviceDna2 == headDeviceDna2
```
rslawson, line 368: I don't remember why I wrote it this way originally, but if you also think it would be better I'd like to just have each of the `when`s below be `when (a /= b) $ putStrLn ...`. I don't think putting these up here particularly helps with readability. It's also fine as it is.

martijnbastiaan: These booleans are reused a little bit down in a `return $ bool1 && bool2 && bool3 ...`. I'm also not super fond of it, but repeating it also felt meh.

PR #968, opened 2025-08-27
Add Initial mdbook

lmbollen:
I propose to create an mdbook for soft documentation. Initially this will be useful for others to read up on:
* The HITL setup
* The demo DUT
* The demo testbench
* The demo driver
* How we write code for the CPUs

So it eliminates the need for us to explain in person every time the questions arise.

For beyond the demo we are currently building it allows us store our knowledge of the system we build in a central location and document design decisions etc. For now I've filled it with _some_ text, but I didn't want to spend too much time on this as a proposal until I got some feedback.

I'd love to get some feedback on this.

### DIFF
--- a/nix/bin/book
+++ b/nix/bin/book
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+cd "$(git rev-parse --show-toplevel)"
+cd docs
+mdbook serve -o


.github/workflows/ci.yml
```diff
@@ -714,6 +715,44 @@ jobs:
...
+  mdbook:
+    name: Build and Publish mdBook
+    runs-on: [self-hosted, compute]
+    defaults:
+      run:
+        shell: git-nix-shell {0} --ignore-environment
+    container:
+      image: ghcr.io/clash-lang/nixos-bittide-hardware:2025-08-05
+      options: --memory=11g
+    if: github.ref == 'refs/heads/main' || github.event_name == 'pull_request'
```
hiddemoll: `github.event_name` will never be `pull_request`, since we only trigger CI on `schedule` and `push`. 
I guess you only want to publish a new book for changes which make it into main, is that correct? If so, you can just remove the second condition and only update the book on nightly CI runs.

.github/workflows/ci.yml
```diff
@@ -714,6 +715,44 @@ jobs:
...
+          mdbook build
+
+      - name: Upload book artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: mdbook
+          path: _build/book
+
+      - name: Deploy to GitHub Pages
+        if: github.ref == 'refs/heads/main'
```
hiddemoll: You already have this check for the whole job, you can remove it here.

hiddemoll (commented):
I think this is a nice framework. I went over your PR commit by commit and left 2 comments, but in a later commit you commented out the exact 2 lines where I added a remark. So those might not even show up in the review.

PR #980, opened 2025-09-15
Getting rid of routing interconnect prefixes

lmbollen:
This PR changes the way we deal with prefixes in the repository.

Before this PR we configure all prefixes of peripherals for the interconnect at top level, requiring us to manually maintain them and route them through our design to the interconnect.
The reason for this is that we can't just assign prefixes because the prefix for the instruction memory must have only the MSB set (address must be `0x80000000`).

After this PR we use `incrementWithBlacklist` to only assign the peripherals with some prefix that is not the instruction memory prefix or `0` since that address is usually used by compilers to represent a null pointer.
The consequence of this is that we can get rid of the routing and as long as you use the generated memory map for your linker script and rust HAL, you wont have to worry about addresses no more.


bittide-instances/src/Bittide/Instances/Pnr/Switch.hs
```diff
@@ -24,7 +24,7 @@ switchExample ::
   Circuit
     ( ConstBwd MM
     , ( Vec 16 (CSignal Basic200 (BitVector 64))
-      , Wishbone Basic200 'Standard 28 (Bytes 4) -- calendar interface
+      , Wishbone Basic200 'Standard 27 (Bytes 4) -- calendar interface
```
hiddemoll, line 27: Was this wrong before? If not, what changed?

bittide/src/Bittide/ProcessingElement.hs
```diff
@@ -128,6 +132,14 @@ processingElement dumpVcd PeConfig{prefixI, prefixD, initI, initD, iBusTimeout,
 
   idC -< extBusses
  where
+  -- We use `init` to generate at least 0 prefixes with `incrementWithBlacklist`
+  prefixes = init $ iMemPfx :> (incrementWithBlacklist prefixBlacklist)
```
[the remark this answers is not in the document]
lmbollen: No longer applies

bittide-instances/src/Bittide/Instances/Pnr/Switch.hs
lmbollen: We previously mapped UART to prefix 00, but that's now an excluded prefix.

PR #981, opened 2025-09-17
Software based UGN capture demo

lmbollen:
This PR adds the top entity for a demo that contains:
* Clock control
* A management unit for monitoring
* A processing element that has S/G units to each link


bittide-instances/src/Bittide/Instances/Hitl/Dut/SoftUgnDemo.hs
rslawson, line 1: The original intent of the `Dut/` directory in my branch was to provide a place put things depended on by the HITL infra and the driver function(s). As it stands, when I go to create a driver function for this HITL test that uses addresses in the memmaps for these devices I will create a circular dependency. In order to break this, you should create a `Hitl/SoftUgnDemo.hs` that has the parts of this file necessary to run the HITL tests and put the HITL things (`softUgnDemoTest` and `softUgnDemoDut` by my reading) in there. Then everything else can remain in this file, which would moving `circuitFnC` out of `softUgnDemoDut` and into its own separate function, or just leaving `softUgnDemoDut` in the current file if possible. That way `Hitl/SoftUgnDemo.hs` can depend on `Hitl/Dut/SoftUgnDemo.hs` for the DUT, and `Hitl/Driver/SoftUgnDemo.hs` can depend on `Hitl/Dut/SoftUgnDemo.hs` for the memmaps).

bittide-instances/src/Bittide/Instances/Hitl/Dut/SoftUgnDemo.hs
```diff
@@ -0,0 +1,857 @@
+-- SPDX-FileCopyrightText: 2025 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE RecordWildCards #-}
```
rslawson, line 5: I don't think this feature ends up being used.

bittide-instances/src/Bittide/Instances/Hitl/Dut/SoftUgnDemo.hs
```diff
@@ -0,0 +1,857 @@
...
+    }
+
+ccLabel :: Vec 2 Byte
+ccLabel = fromIntegral (ord 'C') :> fromIntegral (ord 'C') :> Nil
+
+muLabel :: Vec 2 Byte
+muLabel = fromIntegral (ord 'M') :> fromIntegral (ord 'U') :> Nil
+
+gppeLabel :: Vec 2 Byte
+gppeLabel = fromIntegral (ord 'P') :> fromIntegral (ord 'E') :> Nil
```
rslawson, line 162:
```suggestion
ccLabel, gppeLabel, muLabel :: Vec 2 Byte
ccLabel = fromIntegral (ord 'C') :> fromIntegral (ord 'C') :> Nil
gppeLabel = fromIntegral (ord 'P') :> fromIntegral (ord 'E') :> Nil
muLabel = fromIntegral (ord 'M') :> fromIntegral (ord 'U') :> Nil
```

rslawson (dismissed):

bittide/tests/Tests/DelayWishboneC.hs
```diff
@@ -0,0 +1,71 @@
...
+dutSimpleFn ::
+  (HiddenClockResetEnable dom, 1 <= DomainPeriod dom) =>
+  Signal dom (WishboneM2S 0 4 (Bytes 4)) ->
+  Signal dom (WishboneS2M (Bytes 4))
+dutSimpleFn m2s = let ((_, s2m), _) = circFn (((), m2s), ()) in s2m
+ where
+  Circuit circFn = dutSimple
+
+case_test_wishboneDelayC :: Assertion
+case_test_wishboneDelayC = assertEqual msg readResult.readData whoAmID
```
lmbollen: Is there a reason why you're not using `Protocols.Wishbone.Standard.Hedgehog.wishbonePropWithModel` it seems like this would be an ideal usecase. There are some usage examples in the repo.

martijnbastiaan (approved):

PR #988, opened 2025-09-25
Rewrite of `Node.hs` functions+types, add JTAG to MU+GPPE, add UART to GPPE

rslawson:
In specific, here's a rough list of the changes I made:
- All config types converted from plain GADT to a record GADT
- Nodes now handle capturing UGNs, so they now need the elastic buffer outputs as inputs. This adds another type parameter to the config type (`linkCount`).
- Added UART output to the GPPE function.
- Added JTAG I/O to both the MU and GPPE functions.
- Added function to call `withName` on the scatter/gather engine functions, since those are unsuited to using `withName` as is.
- Added type alias to processing element file for the number of internal Wishbone busses to avoid having an extra magic constant in a number of places.

### DIFF
--- a/bittide/src/Bittide/Node.hs
+++ b/bittide/src/Bittide/Node.hs
@@ -1,6 +1,7 @@
 -- SPDX-FileCopyrightText: 2022 Google LLC
 --
 -- SPDX-License-Identifier: Apache-2.0
+{-# OPTIONS_GHC -fconstraint-solver-iterations=6 #-}
 
 {-# OPTIONS -fplugin=Protocols.Plugin #-}
 -- {-# OPTIONS -fplugin-opt=Protocols.Plugin:debug #-}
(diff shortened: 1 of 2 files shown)


bittide/src/Bittide/Node.hs
```diff
@@ -1,6 +1,7 @@
 -- SPDX-FileCopyrightText: 2022 Google LLC
 --
 -- SPDX-License-Identifier: Apache-2.0
+{-# OPTIONS_GHC -fconstraint-solver-iterations=6 #-}
```
martijnbastiaan, line 4: This should be handled in `bittide.cabal` already. Is this necessary?

bittide/src/Bittide/Node.hs
```diff
@@ -1,6 +1,7 @@
...
 
 {-# OPTIONS -fplugin=Protocols.Plugin #-}
```
martijnbastiaan, line 6: Same for this, AFAIK we always run with the plugin enabled

bittide/src/Bittide/Node.hs
```diff
@@ -10,100 +11,156 @@ module Bittide.Node where
...
+         , Fwd peLinksOut
+         , peUartsOut
+         , cal
+         )
+ where
+  zipC5 ::
+    forall a b c d e n.
+    (KnownNat n) =>
+    Circuit (Vec n a, Vec n b, Vec n c, Vec n d, Vec n e) (Vec n (a, b, c, d, e))
+  zipC5 = applyC (\(a, b, c, d, e) -> zip5 a b c d e) unzip5
```
martijnbastiaan, line 128: Is this upstreamable? Does `clash-protocols` already have `zipC{2,3,4}` etc?

bittide/src/Bittide/Node.hs
```diff
@@ -148,44 +268,82 @@ Bittide Link. It takes a 'ManagementConfig', incoming link and a vector of incom
...
+         , localCounter
+         , switchMMWb
+         , externalMMWb
+         , captureUgnMMWbs
+         , scatterMMWbs
+         , gatherMMWbs
+         )
+
+withNamesSG ::
+  forall a b c.
```
martijnbastiaan, line 337: Could you add documentation? What does this do?

PR #989, opened 2025-09-25
Removed unused memory map outputs from `SwitchDemo` HITL test

rslawson:


martijnbastiaan:
Could you make the PR / commit title fit with what you're doing technically? 


```
git commit -am "Removed superfluous memory maps" -m "Fallout from #984"
```

or something (if this is actually what you're doing).

PR #996, opened 2025-10-02
Use `Signed` while calculating IGNs

martijnbastiaan:
Before:

```
  The IGN for this link is: 18446744073709550106 + 2810 = 1300
```

After:

```
  The IGN for this link is: 2712 + -1418 = 1294
```

Same result, but much prettier :-)


bittide/src/Bittide/ElasticBuffer.hs
rslawson, line 1: Good notes, but were they intended to be in this PR?

rslawson (approved):

Issue #965, opened 2025-08-25
Migration to `deviceWbC` and `registerWbC`


rslawson:
I'm attempting here to make an issue to track migrations of the remaining components that do not yet use these functions into ones that do. Here's a list of all the devices I've found that have a memmap but do not use `registerWbc` and/or `deviceWbC`, whether or not they are suitable for conversion. If they are not, then they should be marked with a note as such and as completed.

- [ ] `uartInterfaceWb` (`bittide/src/Bittide/Wishbone.hs:407`)
- [ ] `readDnaPortE2Wb` (`bittide/src/Bittide/Wishbone.hs:760`)
- [ ] `wbAlwaysAckWith` (`bittide/src/Bittide/Wishbone.hs:853`)
- [ ] `whoAmIC` (`bittide/src/Bittide/Wishbone.hs:870`)
- [ ] `switchDemoPeWb` (`bittide/src/Bittide/SwitchDemoProcessingElement.hs:158`)
- [ ] `scatterUnitWbC` (`bittide/src/Bittide/ScatterGather.hs:206`)
- [ ] `gatherUnitWbC` (`bittide/src/Bittide/ScatterGather.hs:341`)
- [ ] `wbStorageDPC` (`bittide/src/Bittide/DoubleBufferedRam.hs:116`)
- [ ] `wbStorage` (`bittide/src/Bittide/DoubleBufferedRam.hs:227`)
- [ ] `macStatusInterfaceWb` (`bittide/src/Bittide/Ethernet/Mac.hs:46`)

PR #1000, opened 2025-10-07
Move `whoAmIC` to `deviceWb`

martijnbastiaan:
See #965


hiddemoll (approved):
Congrats on PR 1000 :tada:. It's only fitting that this PR simplifies an existing device so much by switching to `deviceWb`.

PR #1012, opened 2025-10-15
Use packed data in `registerWb`

martijnbastiaan:
This makes sure the bus can write intermediate values to the register, without the register blowing up because the intermediate value does not have a representation in the target type. For example, say we have a color type:

```haskell
data RGB = R | G | B
```

The data type can represent three values, but its bit width is 2 bits. This necessarily means that one of the bit values is invalid, in this case 0b11. Now say we store this in an `Either RGB (BitVector 2)`. If the current value of the register is `Right 0b11`, just writing the constructor tag (a perfectly valid thing to do from the CPUs point of view) will result in `Left (unpack 0b11)`, an invalid value!

Note that this strategy might still expose the circuit to intermediate (invalid) values. I.e., these situations still call for custom locking logic.

--------------

This is a stepping stone towards having atomic operations, for which I'll need to do this anyway.

# TODO
- [x] Fix https://github.com/clash-lang/clash-compiler/issues/3036

### DIFF
--- a/clash-protocols-memmap/src/Protocols/MemoryMap/Registers/WishboneStandard.hs
+++ b/clash-protocols-memmap/src/Protocols/MemoryMap/Registers/WishboneStandard.hs
@@ -82,6 +82,7 @@ import Protocols.Wishbone (
 
 import qualified Data.List as L
 import qualified Data.Map as Map
+import qualified Data.String.Interpolate as I
 import qualified Protocols.Df as Df
 import qualified Protocols.Vec as V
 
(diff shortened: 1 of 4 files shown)


clash-protocols-memmap/src/Protocols/MemoryMap/Registers/WishboneStandard.hs
```diff
@@ -469,7 +470,20 @@ registerWbDf clk rst regConfig resetValue =
               , (pure resetValue, busActivity)
               )
  where
-  needReverse = ?busByteOrder /= ?regByteOrder
+  regByteOrder = ?regByteOrder
```
hydrolarus: Is there a reason that this introduces a new binder? AFAIK I only see it used in `needReverse` right below, the rest is using the implicit?

hydrolarus: Ahh right, I missed that in the quasiquote message. It's fine then, might be nice to move it closer to where the "issue" is, but that makes sense

firmware-binaries/test-cases/registerwb_test/src/main.rs
```diff
@@ -29,14 +29,14 @@ fn test_ok() -> ! {
...
-    let mut full_msg = heapless::String::<64>::new();
+    let mut full_msg = heapless::String::<512>::new();
     let _ = write!(full_msg, "FAIL: {}", msg);
     test_result(&full_msg)
 }
 
 fn expect<T: ufmt::uDebug + PartialEq>(msg: &str, expected: T, actual: T) {
     if expected != actual {
-        let mut err = heapless::String::<64>::new();
+        let mut err = heapless::String::<512>::new();
```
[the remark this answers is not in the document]
martijnbastiaan: Okay, could you take another look?

PR #1020, opened 2025-10-27
Replace `resettableXilinxElasticBuffer` with `xilinxElasticBufferWb` in `SwitchDemo`

lmbollen:
The current demo implementation uses `resettableXilinxElasticBuffer`. This component is essentially a wrapper around `xilinxElasticBuffer` that contains a statamachine to initialize the buffer occupancy to ~50% out of reset.

Since in bittide we want to be able to "groom" ugns, i.e control the latency introduced by the elastic buffers, we added low level controls to `xilinxElasticBuffer` to increase or decrease buffer occupancy by 1.

#1019 introduces `xilinxElasticBufferWb`, a wrapper around `xilinxElasticBuffer` that has a wishbone interface that enables us to adjust the buffer occupancy from software.

This pull request replaces the `resettableXilinxElasticBuffer` with `xilinxElasticBufferWb` such that we now center the buffers from software instead of hardware, opening the door to UGN grooming even futher!

TODO:
- [x] Merge #1019
- [x] Merge #981

### DIFF
--- a/.github/synthesis/all.json
+++ b/.github/synthesis/all.json
@@ -1,6 +1,6 @@
 [
   {"top": "counterReducedPins",            "stage": "pnr",   "cc_report": false},
-  {"top": "elasticBuffer5",                "stage": "pnr",   "cc_report": false},
+  {"top": "elasticBufferWb",               "stage": "pnr",   "cc_report": false},
   {"top": "freeze",                        "stage": "pnr",   "cc_report": false},
   {"top": "gatherUnit1KReducedPins",       "stage": "pnr",   "cc_report": false},
   {"top": "safeDffSynchronizer",           "stage": "pnr",   "cc_report": false},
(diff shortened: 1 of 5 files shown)


bittide-instances/src/Bittide/Instances/Hitl/Dut/SwitchDemo.hs
```diff
@@ -232,9 +236,10 @@ circuitFnC ::
     , "CAL_ENTRY" ::: CSignal Bittide (Vec (LinkCount + 1) (Index (LinkCount + 2)))
     , "UART_TX" ::: CSignal Basic125 Bit
     , "SYNC_OUT" ::: CSignal Basic125 Bit
+    , "EB_STABLES" ::: Vec LinkCount (CSignal Bittide Bool)
```
hiddemoll: `CallistoResult` also has a `allStable` field which is set by the CPU. How does `ebStables` differ from `CallistoResult.allStable`?

lmbollen: Callisto's all stable indicates that the clock deviations are within the margins set in `clock_control`'s `StabilityDetector`.

The `EB_STABLES` indicate that the elastic buffers have been centered _after_ the clocks are stable

PR #1037, opened 2025-11-05
Add docs + diagram for the software UGN demo

rslawson:
# TODO
- [x] Update docker image


martijnbastiaan:
@rslawson I've reverted the git dependency (even the new one). It takes 25% of our cache and CI is starting to slow down due to the cache sizes..

PR #1038, opened 2025-11-05
Switch demo gppe dut

lmbollen:
This pull request adds a demo top entity that is similar to the Switch demo with as important difference uses an actual RISC-V cpu instead of the `SwitchDemoProcessingElement`


martijnbastiaan (approved):
LGTM, we can merge this IMO. 

After this PR: when we've got all the hardware for the demos in the `Hello world` examples should be replaced by C programs. In case of the MU (that does actual work using the Rust HALs @ `firmware-binaries/demos/switch-demo2-mu/src/main.rs`) it should probably call a C-hello world. See https://github.com/bittide/bittide-hardware/blob/95d563ca9dbcc3052b4cbdfe2c2602ed6a51f77b/firmware-binaries/examples/c_hello/README.md for more info.

PR #1039, opened 2025-11-05
Reorganize rust binaries

lmbollen:
There was no consistency in the `firmware-binaries` structure and we are going to have to add a couple of more programs.

To keep the code maintainable I propose to differentiate between these 4 categories:
- `examples`: Simple examples of how to use certain pieces of code
- `sim-tests`: Programs that are used by simulation based tests
- `hitl-tests`: Programs that are used to test things on the board
- `demos`: Programs that are used for the hardware demos


hiddemoll:
There's also no difference between `hitl-tests` and `demos`, right?

PR #1058, opened 2025-11-11
Remove link negotation dependency

martijnbastiaan:
To obtain a common clock domain for the transmit side of things (the "Bittide" domain), we configure the transceivers as follows (UG576, figure 3-3):

Screenshot from 2025-11-11 13-39-53

That is, one `TXOUTCLK` serves as the source for _all_ other links. Before this PR, this would impose a dependency on link 0 to all other links, because the reset mechanism would reset the PLL('s datapath) as part of its boot process. The current state machine takes special care to only reset the PLL once. 

------------------

I've done a bunch of playing around with the transceivers and everything seems to behave as it should. We will later add software control of the transceivers which will make it MUCH easier to write tests for this, so I'm postponing tests for a later PR.

-----------------

Survives a bunch of runs:

- [x] https://github.com/bittide/bittide-hardware/actions/runs/19270680985
- [x] https://github.com/bittide/bittide-hardware/actions/runs/19269444544
- [x] https://github.com/bittide/bittide-hardware/actions/runs/19269532183
- [x] https://github.com/bittide/bittide-hardware/actions/runs/19269535458
- [x] https://github.com/bittide/bittide-hardware/actions/runs/19269537278
- [x] https://github.com/bittide/bittide-hardware/actions/runs/19269539426
- [x] https://github.com/bittide/bittide-hardware/actions/runs/19269541205
- [x] https://github.com/bittide/bittide-hardware/actions/runs/19270680985

### DIFF
--- a/bittide/src/Bittide/Transceiver/ResetManager.hs
+++ b/bittide/src/Bittide/Transceiver/ResetManager.hs
@@ -141,6 +141,9 @@ data Resets dom = Resets
   -- ^ Should be routed to @gtwiz_reset_tx_pll_and_datapath_in@.
   , txDatapath :: Reset dom
   -- ^ Should be routed to @gtwiz_reset_tx_datapath_in@.
+  , txDomain :: Reset dom
+  -- ^ Reset that can be paired with the TX clock. It is deasserted once the TX
+  -- clock is stable.
   , txUser :: Reset dom
   -- ^ Should be used to reset user-side logic, e.g., comma generation,
   -- PRBS generation, etc.
@@ -155,6 +158,7 @@ data FsmResets = FsmResets
   { all :: Bool
   , txPllAndDatapath :: Bool
   , txDatapath :: Bool
+  , txDomain :: Bool
   , txUser :: Bool
   , rxPllAndDatapath :: Bool
   , rxDatapath :: Bool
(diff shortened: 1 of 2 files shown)


bittide/src/Bittide/Transceiver/ResetManager.hs
```diff
@@ -144,38 +223,68 @@ resetManager config clk rst tx_init_done rx_init_done rx_data_good =
...
-            (StartRx, stats{rxRetries = satSucc SatBound rxRetries})
+            (ResetRx, stats{rxRetries = satSucc SatBound rxRetries})
         | otherwise ->
-            (RxWait (countSucc cntr), stats)
+            (WaitRx (countSucc cntr), stats)
+      --
       -- Monitor link. Move all the way back to 'StartTx' if the link goes down
       -- for some reason.
+      --
+      -- TODO: Make this more fine-grained, add error state, wait for user channel reset.
```
hiddemoll: The comment still mentions `StartTx`, which no longer exists.

bittide/tests/Tests/Transceiver.hs
```diff
@@ -148,12 +151,21 @@ gthCoreMock
...
+      unsafeFromActiveHigh (unsafeSynchronizer freeClk rxClk (unsafeToActiveHigh rstRxDatapath))
+    rxRst = rxRstAll `unsafeOrReset` rxRstRxPllAndDatapath `unsafeOrReset` rxRstRxDatapath
+
+    txRstAll = unsafeFromActiveHigh (unsafeSynchronizer freeClk txClk (unsafeToActiveHigh rstAll))
+    txRstTxPllAndDatapath =
+      unsafeFromActiveHigh
+        (unsafeSynchronizer freeClk txClk (unsafeToActiveHigh rstTxPllAndDatapath))
+    txRstTxDatapath =
+      unsafeFromActiveHigh (unsafeSynchronizer freeClk txClk (unsafeToActiveHigh rstTxDatapath))
+    txRst = txRstAll `unsafeOrReset` txRstTxPllAndDatapath `unsafeOrReset` txRstTxDatapath
```
hiddemoll: Why don't you use `xpmResetSynchronizer` here?

PR #1059, opened 2025-11-11
Add abstraction trait over calendar types

rslawson:
Previously abstractions were only made over a single switch calendar, however those abstractions should be useable for all other calendar types. As such, it was necessary to make those abstractions agnostic over what calendar they were used on. Now to use the HAL for calendars, one only needs to provide an implementation of the `CalendarInterface` trait. In order to facilitate this, a helper macro is provided (`impl_calendar_interface`). I also then use this helper macro to implement the `CalendarInterface` trait for all the calendar types that currently exist.

### DIFF
--- a/firmware-support/bittide-hal/src/manual_additions/calendar.rs
+++ b/firmware-support/bittide-hal/src/manual_additions/calendar.rs
@@ -2,92 +2,363 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 /*! Calendar Interface
- *
+
 The `calendar` module provides a hardware abstraction layer over based on the generated
 peripheral access code for the `Calendar` device.
  */
 
-use crate::{self as bittide_hal, hals};
+use crate::manual_additions::{FromAs, IntoAs};
 
-use bittide_hal::{
-    hals::switch_c::devices::calendar::Calendar,
-    types::{ValidEntry_12, ValidEntry_16},
-};
+pub trait ValidEntryType {
+    type Inner;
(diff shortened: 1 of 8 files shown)


hiddemoll (approved):
I don't see any problems in this PR. I am definitely not a macro expert, so it might be a good idea to have Lara take a look when she's back. But let's leave that for a possible clean-up PR later so we can get this merged.

firmware-support/bittide-hal/src/manual_additions/calendar.rs
```diff
@@ -2,66 +2,263 @@
...
-    /// Reads entry n from the shadow calendar.
-    pub fn read_shadow_entry(&self, n: usize) -> EntryType {
-        self.set_read_addr(n as _);
+            fn shadow_depth_index(&self) -> Self::ShadowIndex {
+                self.shadow_depth_index()
+            }
+
+            fn shadow_entry(&self) -> Self::EntryType {
+                self.shadow_entry()
+            }
```
hydrolarus: What about changing all these so they use the fully qualified version of the underlying type, so that there can't be a confused that this might be recursive somehow?

```suggestion
            fn shadow_depth_index(&self) -> Self::ShadowIndex {
                $cty::shadow_depth_index(self)
            }

            fn shadow_entry(&self) -> Self::EntryType {
                $ctx::shadow_entry(self)
            }
```

hydrolarus (approved):
Looks good overall

PR #1061, opened 2025-11-12
Refactor `gthCore` to use `inst` and records

martijnbastiaan:


bittide/src/Clash/Cores/Xilinx/Gth/Internal.hs
```diff
@@ -0,0 +1,343 @@
...
+      , Port "gtwiz_userdata_rx_out" rxUser2 (BitVector 64)
+      , Port "gtwiz_reset_tx_done_out" txUser2 Bit
+      , Port "gtwiz_reset_rx_done_out" rxUser2 Bit
+      , Port "txpmaresetdone_out" txUser Bit
+      , Port "rxpmaresetdone_out" rxUser Bit
+      , Port "rxctrl0_out" rxUser2 (BitVector 16)
+      , Port "rxctrl1_out" rxUser2 (BitVector 16)
+      , Port "rxctrl2_out" rxUser2 (BitVector 8)
+      , Port "rxctrl3_out" rxUser2 (BitVector 8)
+      )
```
martijnbastiaan: This should be `@`d where `unPort` is called, otherwise we're suffering from positional hell again.

PR #1060, opened 2025-11-11
C code generator

hydrolarus:
based on #1048

PR #1063, opened 2025-11-12
withName deduplication

lmbollen:
Components that obtain their name through `withName` are not being deduplicated as can be observed by changing `mkCalendarC` to use `withName` instead of altering the device name.

This results in multiple "scatter" and "gather" devices ending up in the `soft-ugn-mu` PAC.

This _clauded_ fix makes the problem go away but might not be the desirable solution. Since we don't rely on `withName` right now this is low priority and should be reviewed after merging #1060


lmbollen:
Indeed fixed by #1107, `withName` use is covered in #1125

PR #1064, opened 2025-11-13
Use transceiver clock for tests which used `USER_SMA_CLOCK`

hiddemoll:
Previously `fincFdecTests` and `si539xConfigTest` used `USER_SMA_CLOCK` to get a clock from the external clock board into logic, which was only connected on a single FPGA in the demo rig. PR #1046 exposed an additional clock from the transceivers which can be used in logic. Now, both the `fincFdecTests` and `si539xConfigTest` can be run on any FPGA in the demo rig, although this PR keeps it on a single FPGA. 

Additionally this PR fixes a reset issue in `si539xConfigTest` which caused random failures on nightly CI.


martijnbastiaan (approved):
Great! How much effort is it to introduce a "boot" CPU in all our DUTs?

PR #1072, opened 2025-11-20
Add diagrams and mdbook entries fo s/g units and calendar

lmbollen:


martijnbastiaan (approved):
I don't think the register descriptions should be in here. They're bound to go out of date. Maybe we can vibe code an HTML backend for the code generation?

PR #1082, opened 2025-11-26
Use latest HLS commit

jaschutte:
Modifies the version of HLS included in the flake to use the latest HLS available.

As for why this is needed; ask @martijnbastiaan

I have compiled HLS and ran it without crashing, but I have not opened an editor to test it within there.


martijnbastiaan:
Looks like the bug didn't actually get fixed. Thanks @jaschutte I'll investigate further.

martijnbastiaan (approved):

PR #1085, opened 2025-11-27
Add `Protocols.Spi` and use it

martijnbastiaan:


bittide/src/Bittide/ClockControl/Si539xSpi.hs
```diff
@@ -382,29 +374,27 @@ si539xSpiDriver ::
...
-    , "MOSI" ::: Signal dom Bit
-    , "SS" ::: Signal dom Bool
-    )
+  , Signal dom Spi.M2S
   )
-si539xSpiDriver SNat incomingOpS miso = (fromSlave, decoderBusy, spiOut)
+si539xSpiDriver SNat incomingOpS s2m = (fromSlave, decoderBusy, spiM2S)
  where
-  spiOut = (sck, mosi, ss)
+  miso = (.miso) <$> s2m
```
lmbollen: Simply use `s2m.miso` instead where you use `miso`?

PR #1100, opened 2025-12-01
Move boot logic to software

martijnbastiaan:


bittide-instances/src/Bittide/Instances/Hitl/SoftUgnDemo/BringUp.hs
```diff
@@ -0,0 +1,209 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP #-}
+
+{- | Switch demo for a Bittide system. In concert with its driver file, this device under
+test should demonstrate the predictability of a Bittide system once it has achieved logical
+synchronicity.
+
+For more details, see [QBayLogic's presentation](https://docs.google.com/presentation/d/1AGbAJQ1zhTPtrekKnQcthd0TUPyQs-zowQpV1ux4k-Y)
+on the topic.
+-}
```
hiddemoll: This is a copy-pasted module description, which is not accurate for this exact module. Shall we replace it with a reference to the mdbook?

hiddemoll: This goes for all 3 demo BringUp modules

PR #1107, opened 2025-12-04
better instance-name generation

hydrolarus:
supersedes #1063


firmware-support/memmap-generate/src/backends/rust/mod.rs
```diff
@@ -124,25 +126,19 @@ pub fn generate_device_instances(
...
-                    }
-                };
-                ident(IdentType::Instance, instance_name)
+            let instance_name_ident = {
+                let raw_name = name_mapping[&handle];
+                if name_counts[raw_name] > 1 {
+                    let count = names_used
+                        .entry(raw_name)
+                        .and_modify(|n| *n += 1)
+                        .or_insert(0);
```
martijnbastiaan: Same comment. Also, why is this duplicated?

PR #1119, opened 2025-12-08
Start MU in parallel with CC

hiddemoll:
For now the MU still needs to know whether clock control is finished. To achieve this the MU has access to `callistoSwClockControlC` with a wishbone arbiter. The MU must not flood this bus, as clock control is more important.

### DIFF
--- a/firmware-binaries/demos/switch-demo1-mu/src/main.rs
+++ b/firmware-binaries/demos/switch-demo1-mu/src/main.rs
@@ -5,7 +5,9 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use bittide_hal::{hals::switch_demo_mu::DeviceInstances, shared_devices::Transceivers};
+use bittide_hal::hals::switch_demo_mu::DeviceInstances;
+use bittide_hal::shared_devices::Transceivers;
+use bittide_sys::stability_detector::Stability;
 use core::panic::PanicInfo;
 use ufmt::uwriteln;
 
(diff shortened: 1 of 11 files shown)


firmware-binaries/demos/switch-demo1-mu/src/main.rs
```diff
@@ -27,32 +39,40 @@ fn main() -> ! {
...
-        &INSTANCES.elastic_buffer_3,
-        &INSTANCES.elastic_buffer_4,
-        &INSTANCES.elastic_buffer_5,
-        &INSTANCES.elastic_buffer_6,
-    ]
-    .iter()
-    .enumerate()
-    {
+    // Keep centering elastic buffers to avoid over/under-flows. Keep track of
+    // number of frames changed while centering.
```
martijnbastiaan:
```suggestion
    // Keep centering elastic buffers to avoid over/under-flows. Keep track of
    // number of frames changed while centering.
    //
    // TODO: Implement asynchronous communication. The idea is that we center
    //       the elastic buffers and capture the UGNs. From that point on, the
    //       links can be used for arbitrary communication. We'll keep on
    //       moving the buffers to their midpoints to prevent overflows (given
    //       that clock control still might be correcting clocks), so we'll
    //       use `eb_changes` to calculate the final UGN once we determine
    //       the whole network is up and therefore stop modifying the buffer
    //       occupancy.
```

martijnbastiaan (approved):

PR #1120, opened 2025-12-09
Lucas/port c code 2

lmbollen:


lmbollen:
We'll stick only to #1057

PR #1123, opened 2025-12-12
Add `c_registerwb_test` to catch C HAL bugs

lmbollen:
We have a massive test for our code generation that is written in rust, `registerwb_test`.

To catch problems with our C backend this pull request adds the C counter part of that test, which caught me two bugs.

~~In this PR I don't intend to fix the bugs themselves, I only added commits that contain fixes to make the tests pass, but I don't know enough about the generation framework to create a proper solution or gauge whether this suffices.~~

### DIFF
--- a/firmware-binaries/sim-tests/c_registerwb_test/build.rs
+++ b/firmware-binaries/sim-tests/c_registerwb_test/build.rs
@@ -1,0 +1,41 @@
+// SPDX-FileCopyrightText: 2025 Google LLC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use memmap_generate::build_utils::standard_memmap_build;
+
+fn main() {
+    // Generate memory.x linker script
+    standard_memmap_build("RegisterWb.json", "DataMemory", "InstructionMemory");
+
+    // Use C headers from auto-generated bittide-hal-c
+    let hal_c_headers = bittide_hal_c::generated_dir();
+
+    // Include directory for the Bittide HAL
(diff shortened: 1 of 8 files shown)


firmware-binaries/sim-tests/c_registerwb_test/build.rs
```diff
@@ -0,0 +1,41 @@
...
+        .include(&hal_c_headers)
+        .include(&hal_c_include)
+        .file("src/main.c")
+        .files(
+            std::fs::read_dir(&hal_c_src)
+                .expect("Failed to read HAL C source directory")
+                .filter_map(Result::ok)
+                .map(|entry| entry.path())
+                .filter(|path| path.extension().is_some_and(|ext| ext == "c")),
+        )
```
hydrolarus: there's `bittide_hal_c::manual_source_files()` which does that same thing and is already used in other build scripts for C crates.

firmware-support/memmap-generate/src/backends/c/mod.rs
```diff
@@ -217,7 +217,7 @@ fn po2_type(n: u64) -> u64 {
     }
 }
 
-fn tuple_name(
+pub fn tuple_name(
```
hydrolarus: this should be reverted too since the proper fix doesn't require this to be public

hydrolarus (changes requested):
The actual new test seems nice, but one of the fixes should not be merged as it is

PR #1124, opened 2025-12-12
Add `NestedInterconnect` test

lmbollen:
A test case that checks whether we can nest interconnects, which should now be possible!

### DIFF
--- a/clash-protocols-memmap/src/Protocols/MemoryMap/Check/AbsAddress.hs
+++ b/clash-protocols-memmap/src/Protocols/MemoryMap/Check/AbsAddress.hs
@@ -13,7 +13,9 @@ import Prelude
 
 import Protocols.MemoryMap
 
-import Data.List (sortBy)
+import Control.Monad (forM, when)
+import Control.Monad.Writer
+import Data.List (sortOn)
 import GHC.Stack (SrcLoc)
 
 import qualified Data.Map.Strict as Map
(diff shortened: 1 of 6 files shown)


clash-protocols-memmap/src/Protocols/MemoryMap/Check/AbsAddress.hs
```diff
@@ -44,44 +44,42 @@ makeAbsolute ::
...
                in (newDevInstance, [err])
             else (newDevInstance, [])
-makeAbsolute ctx (start, end) (AnnInterconnect (tags, path, assertedAddr) srcLoc comps) =
+makeAbsolute ctx (start, size) (AnnInterconnect (tags, path, assertedAddr) srcLoc comps0) =
   checkAssertedAddr start assertedAddr path $
-    let (unzip -> (comps1, concat -> errs)) = flip map (zip ranges comps) $ \((start', end'), (relStart, comp)) -> do
-          let (comp1, errs1) = makeAbsolute ctx (start', end') comp
-          ((relStart, comp1), errs1)
-     in (AnnInterconnect (tags, path, start) srcLoc comps1, errs)
+    let (unzip -> (comps2, concat -> errs)) = flip map componentList $ \((start', size'), comp) -> do
```
martijnbastiaan: It looks like this function should be `Writer [AddressError] MemoryMapTreeAbsNorm`. That would get rid of a bunch of constructor shuffling.

clash-protocols-memmap/src/Protocols/MemoryMap/Check/AbsAddress.hs
```diff
@@ -44,44 +44,42 @@ makeAbsolute ::
...
-    | [] <- comps = []
-    | [(relAddr, _comp)] <- comps = [(start + relAddr, end)]
-    | (first, _) : rest <- comps =
-        let
-          relStarts = zip (first : (fst <$> rest)) $ (fst <$> rest) <> [end - start]
-          startEnds = map (Data.Bifunctor.bimap (start +) (start +)) relStarts
-         in
-          map (Data.Bifunctor.bimap (start +) (end +)) startEnds
+  comps1 = sortBy (\(a, _) (b, _) -> a `compare` b) comps0
+  compRelStart = relAddrs `zip` (drop 1 $ (relAddrs <> [size]))
```
martijnbastiaan:
```suggestion
  compRelStart = relAddrs `zip` (drop 1 (relAddrs <> [size]))
```

or

```suggestion
  compRelStart = relAddrs `zip` (drop 1 $ relAddrs <> [size])
```

martijnbastiaan (changes requested):
For readability's sake I think memory maps should use records much more than they do. As in, `sortOn fst` doesn't mean anything to me, but `sortOn (.relativeAddress)` does. Likewise I think `AnnDeviceInstance{ann, srcLoc, deviceName}` is better than `AnnDeviceInstance (tags, path, assertedAddr) srcLoc deviceName` (with `ann.tags`, etc later on). Usually I wouldn't comment on this, but this code is quite `unzip`/`zip`/`fst`/`flip map` heavy.

martijnbastiaan (approved):

PR #1125, opened 2025-12-12
Use `withName` in `mkCalendarC` instead of altering the device name

lmbollen:
This way we don't make new Calendar devices for each named device. Now we can reuse the same `Calendar` device for switches, gather units and scatter units

### DIFF
--- a/bittide/src/Bittide/Calendar.hs
+++ b/bittide/src/Bittide/Calendar.hs
@@ -347,7 +347,7 @@ mkCalendarC
       (SNat :: SNat repetitionBits)
       bsActive
       bsShadow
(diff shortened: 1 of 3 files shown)


bittide/src/Bittide/Calendar.hs
```diff
@@ -348,9 +347,9 @@ mkCalendarC
       (SNat :: SNat repetitionBits)
       bsActive
       bsShadow
-    ) = circuit $ \(mm, wb) -> do
+    ) = withName compName $ circuit $ \(mm, wb) -> do
```
martijnbastiaan:
```suggestion
    ) = withName (compName <> "_calendar") $ circuit $ \(mm, wb) -> do
```

PR #1139, opened 2026-01-08
Ringbuffer smoltcp device implementation

lmbollen:
This pull request adds the implementation of `smoltcp`'s `Device` trait for a pair of `TransmitRingbuffer` and `ReceiveRingbuffer` alongside with a unit test that verifies whether we can connect, send and receive data.

I have to further sanity check if the current implementation for consuming the `RxToken` and `TxToken` is alright for now or if we can do better (thanks Claude), but at least we have a unit test that shows that the concept works.

The current device implementation uses the first 32 bits of the packet as identifier to differentiate between new and old packets. Incoming packets can be malformed on a physical level, so the implementation processes each packet twice.

TODOs:
- [ ] Investigate whether we can use Ip as medium instead of Ethernet, this would probably require us to calculate and verify the checksum manually. (this would be an optimization)
- [ ] Reconsider if we need to process each packet multiple times or if it would be better to just do it once
- [ ] Take another look at the logic for consuming `RxToken`'s
- [ ] Take another look at the logic for consuming `TxToken`'s
- [ ] Reconsider if the current API for the ringbuffer / scatter / gather components has the desired form `[[u8; 8]; SIZE]`


rslawson (changes requested):
Mostly just did a linting pass since this is a draft PR. I'll have a proper look at this when it's moved out of draft (:

PR #1140, opened 2026-01-14
Move utility functions to separate file, fix lookup for canonicalized devices

rslawson:
Previously if a device name was changed in canonicalization, this updated name was used to do lookup in the devices list when searching for registers. Now if the name is changed the old name is included in the annotation for that device instance, which is then used for register lookup.

### DIFF
--- a/bittide-instances/bittide-instances.cabal
+++ b/bittide-instances/bittide-instances.cabal
@@ -310,6 +310,7 @@ test-suite unittests
   other-modules:
     Df.ElasticBufferWb
     Df.WbToDf
+    Tests.Bittide.Instances.Hitl.Utils.MemoryMap
     Tests.Bittide.Instances.Hitl.Utils.OpenOcd
     Tests.ClockControlWb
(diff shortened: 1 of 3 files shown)


martijnbastiaan (changes requested):
Could you add regression tests?

bittide-instances/src/Bittide/Instances/Hitl/Utils/MemoryMap.hs
```diff
@@ -0,0 +1,176 @@
...
+This is so that any path the function attempts to follow can only refer to one
+location in memory.
+-}
+getPathAddress ::
+  (HasCallStack, Num a) =>
+  -- | Memory map to find the path in
+  MemoryMap ->
+  -- | The path to search for in the memory map
+  [String] ->
+  a
```
hydrolarus: Could this function not work on `MemoryMapTreeAbsNorm`? That already has all the absolute addresses calculated right next to fully canonicalised paths

PR #1143, opened 2026-01-14
Update flake to include latest HLS

jaschutte:
Updates the Nix flake to include the latest unpublished versions of HLS. In specific, this updates HLS to version 2.13.0.0.

Due to how HLS is set up, this required a few changes to accomplish:
1. Changed nixpkgs from nixos-25.05 to 25.11. HLS required newer packages than those available in 25.05.
2. Due to the newer version of nixpkgs, some packages have been renamed/removed. Notably the python3Full has been removed. See the comment in the code.
3. Included the HLS source as a flake input, rather than a hardcoded source inside of the flake

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,15 +3,15 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1761588595,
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
We had to wait long enough for HLS to be updated in `nixpkgs` anyway, making this PR obsolete. Sorry James :sweat:.

PR #1144, opened 2026-01-15
Addressable bytes wb

lmbollen:
_What (what did you do)_
This PR introduces a component `addressableBytesWb` that supports adding memory mapped memories such as `blockRam`s

_Why (context, issues, etc.)_
To be able to more easily integrate blockRam based components in our `clash-protocols-memmap` infrastructure.

Regarding the rust side I added a HAL that contains a reference of functionality that I would like to have. What was especially problematic was the alignment requirement for `[u8]`, or any `[[u8; N]]` that we use to represent bitvectors. Their alignment requirement is 1 byte, where I require the slice to be 4 byte aligned for efficient copying. I tried to offer an interface to create word aligned byte slices with `AlignedArray` that would allow the CPU to perform word sized reads and writes, but am not sure if this is the way to go.

We faced a related challenge earlier in the `axi` module, but that was trivially fixed with `align_to` because it has a streaming interface (we always write to the same destination address).

The unit test was mostly written with AI.

# TODO

### DIFF
--- a/bittide-extra/src/Protocols/ReqResp.hs
+++ b/bittide-extra/src/Protocols/ReqResp.hs
@@ -25,7 +25,7 @@ module Protocols.ReqResp (
   fromBlockRamWithMask,
 
   -- * Other utilities
-  partition,
+  partitionEithers,
   forceResetSanity,
 ) where
 
@@ -38,7 +38,6 @@ import Protocols.BiDf (BiDf)
 import Protocols.Idle
 
 import qualified Clash.Prelude as C
-import qualified Data.List as L
 import qualified Protocols as Df
 import qualified Protocols.BiDf as BiDf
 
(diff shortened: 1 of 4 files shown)


bittide-extra/src/Protocols/ReqResp.hs
```diff
@@ -66,6 +67,30 @@ instance IdleCircuit (ReqResp dom req resp) where
...
+  Simulate (ReqResp dom req resp)
+  where
+  type SimulateFwdType (ReqResp dom req resp) = [Maybe req]
+  type SimulateBwdType (ReqResp dom req resp) = [Maybe resp]
+  type SimulateChannels (ReqResp dom req resp) = 1
+  simToSigFwd _ = C.fromList_lazy
+  simToSigBwd _ = C.fromList_lazy
+  sigToSimFwd _ = C.sample_lazy
+  sigToSimBwd _ = C.sample_lazy
+  stallC conf stalls = C.withClockResetEnable clockGen resetGen enableGen $ circuit $ \rr0 -> do
```
martijnbastiaan:
```suggestion
  stallC conf stalls = C.withClockResetEnable clockGen noReset enableGen $ circuit $ \rr0 -> do
```

It seems like this component should do the "Nothing" thing _or_ listen to the reset settings of `conf`. What do you think?

PR #1175, opened 2026-02-02
Fix small problems in `Wishbone.CaptureUgn` and `Wishbone.Watchdog`

lmbollen:
1: We previously tried to read the counters before knowing they are stored, which lead to a failing test
Discovered in #1172 

2: We could get a wrong reading of the duration of a watchdog timer because we did not align our actions to the start of a microsecond.

3: Expected and actual were the wrong way around

4: When memory regions would change, the `memory.x` would become out of sync


bittide-instances/tests/Wishbone/Watchdog.hs
```diff
@@ -52,7 +52,7 @@ The test returns names of tests and a boolean indicating if the test passed.
 case_time_rust_self_test :: Assertion
 case_time_rust_self_test = do
   let result = L.head $ lines simResult
-  assertEqual "Measured timeout wrong " result "Timeout took 50 microseconds"
+  assertEqual "Measured timeout wrong " "Timeout took 50 microseconds" result
```
hiddemoll: Why does it expect 50 microseconds, while the rust code does a timeout of 10 microseconds?

PR #1176, opened 2026-02-02
Add Clash numeric types plus basic `impl`s

### LINKED ISSUE #1170: Add types for `Index`, `Unsigned`, `Signed`, and `BitVector` in Rust
I'm not a big fan of monomorphizing these types in codegen into Rust primitive integer types (or to array types), and I would much rather that it looks a lot more like how it does in Clash. My previous work on this was maybe _too_ experimental, but I did have an idea for making it work (and making the programmer make it work) in stable Rust.

This library/feature should have:
- Types for each of the ones mentioned in the issue title
- Should work in stable (even though we're on nightly) - should avoid any nightly features and especially incomplete ones
- Should provide macros for codegen and library users to invoke in order to resolve to correctly-formed types

rslawson:
This implements basic functionality for `Index<N, T>`, `Unsigned<N, T>`, `Signed<N, T>`, and `BitVec<N, M>`. This PR is currently marked as draft since I still need to:
- [x] Add documentation to the types themselves and to the size check traits, not just their methods
- [x] Create macros as aliases for the types
- [x] Create instantiation macros
- [x] Add tests?

Anywho, at least an initial review pass and/or suggestions on things I should add in this PR that I didn't (and that I don't have WIP stuff for already from before I did a lot of downscoping, like operator `impl`s).

### DIFF
--- a/firmware-binaries/sim-tests/addressable_bytes_wb_test/src/main.rs
+++ b/firmware-binaries/sim-tests/addressable_bytes_wb_test/src/main.rs
@@ -7,6 +7,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+use bittide_macros::bitvector;
 use ufmt::{uWrite, uwrite, uwriteln};
 
 use bittide_hal::hals::addressable_bytes_wb as hal;
@@ -63,7 +64,7 @@ fn main() -> ! {
     uwriteln!(uart, "Starting addressable_bytes_wb_test").unwrap();
 
     // Test data: 8 words with distinct patterns
-    let test_data: [[u8; 4]; 8] = [
+    let test_data = [
         [0x00, 0x10, 0x20, 0x30],
         [0x01, 0x11, 0x21, 0x31],
         [0x02, 0x12, 0x22, 0x32],
@@ -72,7 +73,8 @@ fn main() -> ! {
         [0x05, 0x15, 0x25, 0x35],
         [0x06, 0x16, 0x26, 0x36],
         [0x07, 0x17, 0x27, 0x37],
-    ];
+    ]
+    .map(|bytes| bitvector!(bytes, n = 32));
 
     // Test 1: Word-based access for whole memory
     uwriteln!(uart, "Test word-based access").unwrap();
@@ -86,7 +88,11 @@ fn main() -> ! {
 
     buffer.clear();
     for i in 0..8 {
-        expect("Clearing failed", Some([0u8; 4]), buffer.data(i));
+        expect(
+            "Clearing failed",
+            Some(bitvector!(0x0, n = 32)),
+            buffer.data(i),
+        );
     }
     uwriteln!(uart, "  PASS").unwrap();
 
@@ -115,7 +121,11 @@ fn main() -> ! {
 
         buffer.clear();
         for i in 0..8 {
-            expect("Clearing failed", Some([0u8; 4]), buffer.data(i));
+            expect(
+                "Clearing failed",
+                Some(bitvector!(0x0, n = 32)),
+                buffer.data(i),
+            );
         }
     }
 
@@ -139,7 +149,11 @@ fn main() -> ! {
 
                 buffer.clear();
                 for i in 0..8 {
-                    expect("Test 3 clear", Some([0u8; 4]), buffer.data(i));
+                    expect(
+                        "Test 3 clear",
+                        Some(bitvector!(0x0, n = 32)),
+                        buffer.data(i),
+                    );
                 }
             }
         }
--- a/firmware-binaries/sim-tests/nested_interconnect_test/src/main.rs
+++ b/firmware-binaries/sim-tests/nested_interconnect_test/src/main.rs
@@ -6,9 +6,9 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use ufmt::{uwrite, uwriteln};
-
 use bittide_hal::hals::nested_interconnect as hal;
+use bittide_macros::unsigned;
+use ufmt::{uwrite, uwriteln};
 
 use core::fmt::Write;
 #[cfg(not(test))]
@@ -72,23 +72,23 @@ fn main() -> ! {
 
         // Read initial values (should be 0)
         write!(name_buf, "peripheral{i}.status.init").unwrap();
-        expect(&name_buf, 0u32, peripheral.status());
+        expect(&name_buf, 0u32, peripheral.status().into_inner());
         name_buf.clear();
 
         write!(name_buf, "peripheral{i}.control.init").unwrap();
-        expect(&name_buf, 0u32, peripheral.control());
+        expect(&name_buf, 0u32, peripheral.control().into_inner());
         name_buf.clear();
 
         // Write and read back status
-        peripheral.set_status(*status_val);
+        peripheral.set_status(unsigned!(*status_val, n = 32));
         write!(name_buf, "peripheral{i}.status.readback").unwrap();
-        expect(&name_buf, *status_val, peripheral.status());
+        expect(&name_buf, *status_val, peripheral.status().into_inner());
         name_buf.clear();
 
         // Write and read back control
-        peripheral.set_control(*control_val);
+        peripheral.set_control(unsigned!(*control_val, n = 32));
         write!(name_buf, "peripheral{i}.control.readback").unwrap();
-        expect(&name_buf, *control_val, peripheral.control());
+        expect(&name_buf, *control_val, peripheral.control().into_inner());
     }
 
     // Verify all peripherals are still accessible
@@ -97,7 +97,7 @@ fn main() -> ! {
     {
         let mut name_buf = heapless::String::<50>::new();
         write!(name_buf, "peripheral{i}.status.final").unwrap();
-        expect(&name_buf, *status_val, peripheral.status());
+        expect(&name_buf, *status_val, peripheral.status().into_inner());
     }
 
     test_ok()
--- a/firmware-binaries/sim-tests/switch_calendar_test/src/main.rs
+++ b/firmware-binaries/sim-tests/switch_calendar_test/src/main.rs
@@ -12,6 +12,7 @@ use bittide_hal::{
     manual_additions::calendar::{CalendarEntryType, CalendarType},
     types::ValidEntry_12,
 };
+use bittide_macros::{index, unsigned};
 use ufmt::{uDebug, uwrite, uwriteln};
 
 #[cfg(not(test))]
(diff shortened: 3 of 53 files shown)


firmware-support/bittide-macros/src/lib.rs
```diff
@@ -151,6 +147,365 @@ impl Parse for IndexValArgs {
...
+                    input.span(),
+                    concat!(
+                        "Expected a hexadecimal literal! Must start with `0x` or `0X` and contain ",
+                        "only hexdigits, spaces, and underscores",
+                    )
+                    .to_string(),
+                )
+                .into_compile_error()
+                .into();
+            }
```
hydrolarus: Wouldn't it make sense to also support binary literals and maybe even base 10 literals?

firmware-support/bittide-hal/src/manual_additions/bitvector.rs
```diff
@@ -0,0 +1,455 @@
...
+                for Signed<A, $backer>
+            where
+                Self: SignedInterface<Inner = $backer>,
+                BitVector<B, $size>: BitVectorSizeCheck<Inner = [u8; $size]>,
+            {
+                const _SIZE_CHECK: () = {
+                    if A as usize > B {
+                        const_panic::concat_panic!(
+                            const_panic::fmt::FmtArg::DISPLAY;
+                            "`Unsigned<",
```
hydrolarus: This should be `Signed`

firmware-support/bittide-hal/src/manual_additions/index.rs
```diff
@@ -0,0 +1,451 @@
...
+                    if N == 0 {
+                        panic!("Cannot represent Index<0, T>!");
+                    }
+                    // `u128::BITS - N.leading_zeros()` is equivalent to `N.clog2()`
+                    let correct_size = match
+                        (u128::BITS - (N - 1).leading_zeros()).next_power_of_two()
+                    {
+                        size if size >= 8 => size,
+                        _ => 8,
+                    };
```
hydrolarus: This could use `.max(8)` instead of this match

hydrolarus: `const` 🫠

firmware-support/bittide-hal/src/manual_additions/signed.rs
```diff
@@ -0,0 +1,399 @@
...
+    ///
+    /// # Safety
+    ///
+    /// Due to the intended use-case of interfacing with Clash hardware, and there being no
+    /// guarantee of behaviour in the case that an out-of-bounds value is written to a register,
+    /// this function has been marked as `unsafe`. To make calling this function safe, you must
+    /// guarantee that `val` is in the range `-2.pow(N - 1)..2.pow(N - 1)`.
+    unsafe fn uns_new_unchecked(val: Self::Inner) -> Self;
+    /// Unwrap the inner value contained by this `Signed<N, T>`
+    fn uns_into_inner(self) -> Self::Inner;
```
hydrolarus: What does the `uns_` prefix mean here?

rslawson: Missed that this was in the `signed` module. Fixing.

firmware-support/bittide-macros/src/lib.rs
```diff
@@ -39,12 +36,12 @@ pub fn Index(toks: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
 
     let ty = {
-        let type_name = format!("u{}", next_pow2(&clog2(&n_val)));
+        let type_name = format!("u{}", get_correct_size(n_val));
         syn::Ident::new(&type_name, n_lit.span())
     };
 
     quote! {
-        bittide_hal::manual_additions::index::IndexTy<#n_lit, #ty>
+        Index<#n_lit, #ty>
```
hydrolarus: Sadly proc macros don't have the same hygiene as declarative macros... I guess it's because of using the macros in the `bittide_hal`? Otherwise I feel like the fully qualified version would be more robust?
Maaaybe it's possible to check if the macro is being invoked in `bittide_hal` and then change the prefix?

firmware-binaries/sim-tests/nested_interconnect_test/src/main.rs
```diff
@@ -71,23 +70,23 @@ fn main() -> ! {
...
-        peripheral.set_status(*status_val);
+        peripheral.set_status(Unsigned::new(*status_val).unwrap());
         write!(name_buf, "peripheral{}.status.readback", i).unwrap();
-        expect(&name_buf, *status_val, peripheral.status());
+        expect(&name_buf, *status_val, peripheral.status().into_inner());
         name_buf.clear();
 
         // Write and read back control
-        peripheral.set_control(*control_val);
+        peripheral.set_control(Unsigned::new(*control_val).unwrap());
```
hydrolarus: `unsigned!(<expr>, n = N)`

firmware-binaries/sim-tests/switch_calendar_test/src/main.rs
```diff
@@ -53,32 +57,32 @@ fn expect<T: uDebug + PartialEq>(msg: &str, expected: T, actual: T) {
...
-            ve_repeat: 16,
+            ve_repeat: Unsigned::new(16).unwrap(),
         },
     ];
 
     let cal_shadow: [CalendarEntryType<Calendar>; 16] = core::array::from_fn(|i| ValidEntry_12 {
-        ve_entry: core::array::from_fn(|_j| i as u8),
-        ve_repeat: i as u16,
+        ve_entry: core::array::from_fn(|_j| Index::new(i as u8).unwrap()),
+        ve_repeat: Unsigned::new(i as u16).unwrap(),
```
hydrolarus: Same here, contruction macros should probably be preferred unless there's a good reason to not use them

hydrolarus (commented):
Overall looking good, but some questions regarding macro hygiene, style of type macros and usage of expr-macros

firmware-support/bittide-macros/src/lib.rs
rslawson: I did find the `proc-macro-crate` crate that does this, and was able to put something together that works for this.

firmware-support/bittide-macros/src/lib.rs
rslawson: Added binary literals, at least. Still not sure what to do for decimal literals.

firmware-support/bittide-sys/src/sample_store.rs
```diff
@@ -37,49 +39,69 @@ impl SampleStore {
...
         );
 
         // Store net speed change
-        self.memory
-            .set_data(start_index + 5, (net_speed_change as u32).to_le_bytes());
+        self.memory.set_data(
+            start_index + 5,
+            // BitVector::new((net_speed_change as u32).to_le_bytes()).unwrap(),
+            unsafe { BitVector::new_unchecked((net_speed_change as u32).to_le_bytes()) },
+        );
```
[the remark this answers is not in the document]
rslawson: After the debug profile changes, this seems to be less of an issue. Reverted code back to the safe interfaces.

firmware-support/bittide-hal/src/manual_additions/calendar.rs
```diff
@@ -211,11 +175,12 @@ pub trait CalendarType: CalendarInterface {
...
     /// Panics if `n` is an invalid value for a `Self::ReadIndex` type instance.
     fn read_shadow_entry(&self, n: usize) -> Self::EntryType {
+        let n_inner = n.into_as();
         assert!(
-            n as u128 <= Self::READ_ADDR_MAX.into_as(),
+            n_inner <= Self::Index::MAX,
             "Shadow entry read index is out of bounds."
         );
-        self.calint_set_read_addr(Self::ReadIndex::from_as(n));
+        self.calint_set_read_addr(unsafe { Self::Index::idx_new_unchecked(n_inner) });
```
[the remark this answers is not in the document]
rslawson: Switched to the safe stuff, seems to work well after the debug profile changes.

firmware-support/bittide-hal/src/manual_additions/calendar.rs
```diff
@@ -50,36 +47,28 @@ macro_rules! impl_valid_entry_type {
 
 impl_valid_entry_type! {
     type: [crate::types::ValidEntry_12],
-    repeat: u16,
+    repeat: Unsigned!(16),
```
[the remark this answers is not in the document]
hydrolarus: I think I probably prefer `Unsigned![16]` personally, looks more "typey" to me

firmware-binaries/sim-tests/addressable_bytes_wb_test/src/main.rs
```diff
@@ -72,7 +73,8 @@ fn main() -> ! {
         [0x05, 0x15, 0x25, 0x35],
         [0x06, 0x16, 0x26, 0x36],
         [0x07, 0x17, 0x27, 0x37],
-    ];
+    ]
+    .map(|bytes| bitvector!(bytes, n = 32));
```
hydrolarus, line 77: Hmmm this does add quite a lot of overhead, right? Because if we did `[[bitvector!(0x00, n = 32), ...]]` instead of the map, then all the checks would be elided, right?

Does the compiler know that because it's byte arrays that there doesn't need to be a size check?

firmware-support/bittide-hal-c/Cargo.toml
```diff
@@ -13,6 +13,7 @@ workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [build-dependencies]
+bittide-macros = { path = "../bittide-macros" }
```
hydrolarus: This shouldn't be needed, right? This crate doesn't make use of any generated Rust code _or_ the macros themselves.

firmware-support/bittide-hal/src/manual_additions/index.rs
```diff
@@ -0,0 +1,661 @@
...
+    m
+}
+
+const U128_MAXLEN: usize = clog2_cgu128::<{ u128::MAX }>() + 1;
+type CGu128FmtData = [u8; U128_MAXLEN];
+
+struct CGU128Formatter<const N: u128>;
+
+impl<const N: u128> CGU128Formatter<N> {
+    const DATA: CGu128FmtData = {
```
hydrolarus, line 384: I can see what it's doing, but I'm not sure what the "CG" prefix is for?

firmware-support/bittide-macros/src/lib.rs
```diff
@@ -151,6 +170,499 @@ impl Parse for IndexValArgs {
...
+            (Some(byte), shift) => {
+                *byte |= val << shift;
+                byte_idx += 1;
+            }
+        }
+    }
+}
+
+#[proc_macro]
+#[allow(unstable_name_collisions)]
```
hydrolarus: Huh, never seen that one before 😅 What does it do and why do we allow it here?

hydrolarus (commented):
Github not showing diffs between commit updates/force pushes means I didn't look at everything completely fresh again, but overall this looks okay!

Just small documentation requests and nitpicks :)

firmware-binaries/sim-tests/addressable_bytes_wb_test/src/main.rs
rslawson: It shouldn't add much in the way of overhead - the emitted code _should_ use the `new_unchecked` function in the case that `n` is exactly the same size as the backing type (and it should be in this case).

firmware-support/bittide-hal/src/manual_additions/index.rs
rslawson: It's for Const Generic.

firmware-support/bittide-macros/src/lib.rs
rslawson: Ah. I forgot to remove that when I removed the `itertools` dependency.

firmware-support/bittide-hal-c/Cargo.toml
rslawson: Huh, wonder how that slipped in. Weird.

firmware-binaries/sim-tests/addressable_bytes_wb_test/src/main.rs
rslawson: Just checked and apparently I had written the condition wrong (oops!), but now I have it fixed and it does in fact expand to `unsafe { BitVector::<_, _>::new_unchecked(...) }`, which should be a no-op.

hydrolarus: Ok I'll trust you on that! I guess we'll see if it would cause issues, just thought the "inline" versions would be very easily safe haha

hydrolarus (approved):
Github says I already saw all other files and they are unchanged, soooo I guess I'm happy :)

firmware-support/memmap-generate/src/backends/rust/mod.rs
```diff
@@ -675,33 +687,45 @@ fn generate_type_ref(
...
-            if let TypeRef::Nat(n) = size {
-                // TODO maybe special case on n == 1 like the C backend?
-                let n = n.div_ceil(8) as usize;
-                let n_lit = Literal::usize_unsuffixed(n);
-                quote! { [u8; #n_lit] }
+            if let &TypeRef::Nat(n) = size {
+                refs.use_bitvec = true;
+                let n = n as usize;
+                let len = n.div_ceil(8);
+                quote! { BitVector<#n, #len> }
```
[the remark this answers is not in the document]
rslawson: Missed responding to this, but I believe I have fixed this issue with use of `Literal::*_unsuffixed`.

PR #1177, opened 2026-02-03
`SoftUgnDemo` catch missed deadlines

lmbollen:
_What (what did you do)_
I increased `STARTING_DELAY_READ` in the C code of `SoftUgnDemo` and added a detection mechanism to allow the C code to fail the hitl test, which I then used to fail the test on missed deadlines.

_Why (context, issues, etc.)_
In the logs of a failing HITL run I noticed that it contained a significant number (25) of missed read deadlines. I also noticed that most runs had 7 missed read deadlines. 

My suspicion is that `STARTING_DELAY_READ` was too low, causing the first deadlines for each port to be missed.
For this run however, it had not only 7, but 25 missed read deadlines. I suspect we missed the deadline for an event that captured a UGN, leading to an off by 4k on that value.  

Sanity check if the infra to make the C code fail

I ~tried to~ use _actual intelligence_, instead of _artificial intelligence_
# TODO

- [x] Write (regression) test

### DIFF
--- a/bittide-instances/src/Bittide/Instances/Hitl/SoftUgnDemo/Driver.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/SoftUgnDemo/Driver.hs
@@ -224,6 +224,17 @@ driver testName targets = do
                       <> " out)"
                 error "Roundtrip latencies did not match between hardware and software"
               when (not $ L.null mismatchedUgns) $ error "Some UGN edges did not match!"
+
+              liftIO
+                $ T.tryWithTimeoutOn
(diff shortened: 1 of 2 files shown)


hiddemoll (changes requested):
It now properly fails when deadlines are missed. But increasing `STARTING_DELAY_READ` alone does not seem to make the test pass consistently as it just failed. Could you either increase the `STARTING_DELAY_READ` even further or add another fix for the missed deadlines? Either way, please run the HITL test a few times so we don't just get a lucky pass.

Issue #1168, opened 2026-01-28
Intermittent elastic buffer overflows during parallel UGN capture and clock control


hiddemoll:
In PR #1165 we show we can do UGN capture while clocks are not yet stable (e.g. [this](https://github.com/bittide/bittide-hardware/actions/runs/21436423198/job/61728412892) CI run). However, those  tests could sometimes fail due to elastic buffer over/underflows ([this](https://github.com/bittide/bittide-hardware/actions/runs/21402489899/job/61621087524) CI run). The failure rate very low so it is hard to find the cause. The solution depends on when the elastic buffers over/underflowed.

### Situation 1:
The elastic buffers over/underflow after we start sending bittide data and before we capture the UGN. During this time we cannot center the elastic buffers since we might discard the frame with UGN data.
Possible solutions:
- Delay sending bittide data, so clock frequencies are closer together
- Speed up MU statemachine so the elastic buffer centering happens as close as possibe to the start of sending bittide data and UGN capture.

### Situation 2:
The stability detector is not strict enough and so we stop centering the elastic buffers too soon.
Possible solutions:
- Also use `clock_control.links_settled()` in MU stability detector. However, MU must not flood the clock_control bus since the clock control CPU needs prio.
- Make the stability detector more strict
- Continue centering the elastic buffers for x time after stability is detected

PR #1190, opened 2026-02-12
Fix elastic buffer overflows in switch demo

hiddemoll:
_What (what did you do)_
Track more debug information from the elastic buffer:
- Store local counter upon under/overflow
- Min and max data count seen

I've reduced the time between switching to user mode and starting with auto-centering the elastic buffers after the UGN frame has been captured. Before this PR this was >600,000 cycles, now it is at most 1,600 cycles. 

_Why (context, issues, etc.)_
Fixes: #1168

I'm not particulary happy with the commit "Clear under/overflow registers together with their timestamps", so take a good look at that.

AI was useful while debugging, but didn't make it in the PR.

# TODO

- [ ] ~~Update documentation, including `docs/`~~ Does not apply imo
- [x] Link to existing issue

### DIFF
--- a/firmware-binaries/demos/switch-demo1-mu/src/main.rs
+++ b/firmware-binaries/demos/switch-demo1-mu/src/main.rs
@@ -45,7 +45,10 @@ impl LinkStartup {
     ) {
         self.state = match self.state {
             UgnCaptureState::WaitForChannelNegotiation => {
-                if transceivers.handshakes_done(channel).unwrap_or(false) {
+                if transceivers
+                    .handshakes_done(channel)
+                    .expect("Channel out of range")
+                {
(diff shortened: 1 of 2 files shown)


firmware-binaries/demos/switch-demo1-mu/src/main.rs
```diff
@@ -68,6 +68,7 @@ impl LinkStartup {
                 // Center the elastic buffer once before switching to user mode
                 elastic_buffer.set_occupancy(0);
                 elastic_buffer.clear_flags();
```
martijnbastiaan: Maybe this should have a better name now that we have multiple things we can clear @ `clear_flags`.

PR #1196, opened 2026-02-17
Fix nested interconnect canonicalization and path traversal

### LINKED ISSUE #1195: `canonicalizeMemoryMapTree` does not work with nested interconnects
https://github.com/bittide/bittide-hardware/pull/1187 introduces a nested interconnect in the memorymap which then causes the `SwitchDemoTest` to fail due to `getPathAddress` not being able to resolve the address to `CaptureUgn0`. After investigation we found that `canonicalizeMemoryMapTree` was not able to canonicalize the `CaptureUgn` components in its memory map. All `CaptureUgn` components still have the `CaptureUgn` name instead of `CaptureUgnN`

lmbollen:
Fixes #1195
_What (what did you do)_
Change the implementation of `getPathAddress` to allow nested interconnects.  We add a utility to list the children of a given path (like the `ls` Shell command). This was implemented in order to allow for the user to determine the correct path to use in the case of finding a device in a memory map with nested interconnects.

_Why (context, issues, etc.)_
`getPathAddress` was broken for memorymaps with nested interconnects.

# TODO

- [x] Write (regression) test
- [x] Link to existing issue

### DIFF
--- a/bittide-instances/src/Bittide/Instances/Hitl/Utils/MemoryMap.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/Utils/MemoryMap.hs
@@ -6,7 +6,6 @@ module Bittide.Instances.Hitl.Utils.MemoryMap where
 import Prelude
 
 import Control.Monad (unless, when)
-import Data.Bifunctor (bimap)
 import Data.Either.Extra (maybeToEither)
 import Data.List (unsnoc)
 import Data.Maybe (fromMaybe, mapMaybe)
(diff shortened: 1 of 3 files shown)


bittide-instances/src/Bittide/Instances/Hitl/Utils/MemoryMap.hs
```diff
@@ -75,48 +82,72 @@ getPathAddress ::
...
   a
-getPathAddress mm = traverseTree canonicalTree
+getPathAddress mm pathWithName =
+  let
+    errorMessage = error "Empty path given!"
+    (pfx, name) = fromMaybe errorMessage (unsnoc pathWithName)
+    children = listChildren mm pfx
+   in
+    case filter (\(_, childName) -> childName == name) children of
+      [] -> error [i|Prefix #{pfx} has no child #{name}|]
```
martijnbastiaan: If you want to write this properly, things that are expected to go wrong (wrong user input) should return a `Left`. Things that are an internal error (i.e., the logic within this function would be wrong) can use `error`.

martijnbastiaan:
@lmbollen I don't understand the test changes. Do they actually test nested interconnects now?

PR #1213, opened 2026-03-03
Use new ringbuffers in `SoftUgnDemo`

lmbollen:
_What (what did you do)_
Replace the scatter and gather unit in the `SoftUgnDemo` with their dedicated ringbuffer counterparts.

This required a few functional changes to the softUgnDemo, namely:
* We have to extract the applied alignment correction to correctly calculate the arrival time for ugn messages
* We have  to adjust the receive event timing based on the extracted alignment offset

Both changes stem from the fact that there was a fixed relation between when a message would arrive and where it would be written. This relation does not exist anymore because we temporarily pause the counter inside the receive buffer to align it to its neighboring transmit buffer.

_Why (context, issues, etc.)_
We aim to simplify the demos and the current scatter and gather units are used as ringbuffers.

Are there any other places we want them? Is the rust API feature complete?

Used AI for the initial Rust HAL and to find and fix outdated comments.

# TODO

- [x] Write (regression) test
- [ ] Link to existing issue


bittide-instances/src/Bittide/Instances/Hitl/SoftUgnDemo/Core.hs
```diff
@@ -96,8 +95,8 @@ muConfig ::
 muConfig =
   PeConfig
     { cpu = Riscv32imc.vexRiscv1
-    , depthI = SNat @(Div (16 * 1024) 4)
-    , depthD = SNat @(Div (16 * 1024) 4)
+    , depthI = SNat @(Div (10 * 16 * 1024) 4) -- One RAMB18E2 is 16KB, this uses 10 of them.
+    , depthD = SNat @(Div (10 * 16 * 1024) 4) -- One RAMB18E2 is 16KB, this uses 10 of them.
```
hiddemoll: Why do we suddenly need 10x the instruction and data memory sizes? The scatter/gather units were just as big.

docs/sections/demos/switch-demo.md
```diff
@@ -52,15 +52,14 @@ Components:
 - Crossbar calendar
 - Null link
 - ASIC processing element
-- 1x scatter/gather units
+- 1x transmit/receive ring buffers
```
hiddemoll: I left in this chapter about the SwitchDemo, partly because we refer to it in `wire-demo.md`. But now that I see you make changes to it, even though the `SwitchDemo` no longer existst, it feels kind of wrong. Do you think we should delete this chapter?

firmware-binaries/demos/soft-ugn-mu/src/ring_buffer_align.h
```diff
@@ -14,78 +14,37 @@
...
- *
- * Phase 2: Acknowledgement
- * - Each node changes TX message to RINGBUFFER_ALIGN_ACKNOWLEDGE
- * - Each node waits for all partners to send ACKNOWLEDGE at known positions
- *
- * @param ugn_ctx The UGN context containing scatter and gather units
- * @param incoming_offsets Array to store discovered offsets for each port
+ * @brief Align ring_buffers across all ports using a per-port state machine.
+ *
+ * This function implements a distributed alignment protocol that physically
```
hiddemoll: I don't think the buffers are 'physically' alligned

hiddemoll (approved):
Nice work Lucas! 

> Are there any other places we want them?
I think both demos is enough. I'll probably remove the scatter/gather units in a different PR soon(tm).

> Is the rust API feature complete?
I think so

docs/sections/demos/switch-demo.md
lmbollen: I think we should remove, I made a PR!

bittide-instances/src/Bittide/Instances/Hitl/SoftUgnDemo/Core.hs
lmbollen: I think this may have been leftover from development

PR #1208, opened 2026-02-24
Add ringbuffer components

lmbollen:
_What (what did you do)_
Add dedicated components that implement the ringbuffer memories. These components use renewed Protocols infrastructure that aims to make it easier to create components using high level composition of inherently correct `Circuit`s rather than creating the low level implementations by hand.

_Why (context, issues, etc.)_
To simplify the hardware architecture and the way we describe our `Circuit`s.

_Dear reviewer (anything you'd like the reviewer to pay close attention to?)_
Whether the new components and their types improves the readability / robustness of the code.

_AI disclaimer (heads-up for more than inline autocomplete)_
Used AI to generate the ringbuffer tests.

# TODO
_~~Cross-out~~ any that do not apply_

- [x] Update documentation, including `docs/`
- [ ] Link to existing issue

Issue #1173, opened 2026-01-29
Add option to auto-align ring buffers


martijnbastiaan:
We imagine a general bittide system v1 looks like the following:

![Image](https://github.com/user-attachments/assets/96b57539-c888-40ad-9aaa-53520fad9c72)

An RX ringbuffer takes the output of an elastic buffer and writes to consecutive indices -- with wraparound. That is, on clock cycle 0 it writes to index 0, on clock cycle 1 it writes to index 1, etc. A TX ring has the same index logic, but instead of taking an output from an elastic buffer, it writes to a transceiver/link. From the following assumptions:

* The elastic buffers keep a consistent mapping from a clock counter on node _a_ to a clock counter on node _b_
* The ring buffers are the same size on node _a_ and _b_

We can derive that a mapping exist from a ring buffer index _n_ on node _a_ to a ring buffer _n + offset_ on node _b_. To find this offset, we write a 1 to the first index of the TX buffer on node _a_. Node _b_ scans its RX buffer until it finds that and remembers that offset. 

We can simplify this process by:

- Attaching the reset of an RX buffer to whether the transceiver is receiving (user) words
- Attaching the reset of a TX buffer to whether the transceiver is sampling (user) words

This will auto-align buffers, meaning index 0 of a TX buffer always refers to index 0 of a corresponding RX buffer.

PR #1214, opened 2026-03-03
Add example application of TCP-over-ringbuffers

lmbollen:
_What (what did you do)_
This pull request adds the necessary library components plus a demo application that showcases asynchronous commmunication between nodes over a bittide network.

With these library components we can add a networking layer between nodes that allow them to arbitrarily exchange data in a reliable way by leveraging the following techniques:

**Aligned ringbuffers**
The hardware architecture allows CPUs to communicate through a ringbuffer extraction that is shared with its neighbour. Through the `transmitRingbuffer` and `receiveRingbuffer` implemented in #1208 we now have a way to share unidirectional memory regions with our link partners.

**Packetless transmission header**
Since we simply share a memory region with our neighbor to which we can write and they can read, there is no notion of packet transmission and arrival, the data will just end up in memory accessible to our neighbor.
To still allow us to transfer packets, we made a minimal header that allows our neighbor to reliably determine if there is valid data in its memory.

The packet header consists of a checksum, sequence counter, length and payload. We use the checksum to see if there are new packets, the length to determine how long the packet is and for which bytes to calculate the checksum and the sequence counter ensures that when the same packet is send twice, the checksum changes so we can distinguish it from a stale packet.

**TCP based communication statemachine**
To still deal with data corruption, we leverage the TCP statemachine as offered by the smoltcp package to take care of reliable data transmission.

*LinkInterface*
Each node will have a LinkInterface abstraction that contains everything necessary to facilitate this asynchronous communication channel. It allows the CPU to send and receive arbitrary data to/from its direct neighbor.

_Why (context, issues, etc.)_
To be able to perform asynchronous communication in-band over the bittide network. Having an arbitrary asynchronous communication layer allows us to  distribute programs, schedules and relevant parameters over the network.

I use MaybeUnInit to deal with the self-referential nature of the smoltcp elements in LinkInterface, is this the way?
Is the LinkInterface API (over)complete?
I'm not sure if this `WireEncode` / `WireDecode` approach is the way to got. I saw the other way required you to use certain types `u32<LE>` for example.

AI has been used throughout the engineering of different approaches to implement this layer and used to combine all aspects into the LinkInterface. 

# TODO
- [x] Write (regression) test
- [ ] Link to existing issue


firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
+#![no_std]
+#![cfg_attr(not(test), no_main)]
+#![feature(sync_unsafe_cell)]
```
hydrolarus, line 3: Does this crate actually need this feature? It doesn't seem to be used here anywhere?

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+
+#[cfg(not(test))]
+use riscv_rt::entry;
+
+#[cfg_attr(not(test), entry)]
+fn main() -> ! {
+    let mut uart = INSTANCES.uart;
+    unsafe {
+        use bittide_sys::uart::log::LOGGER;
+        let logger = &mut (*LOGGER.get());
```
hydrolarus, line 41: Ah is it maybe about this `.get()`?

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+    // 6) Run manager state machines to connect to neighbors and request UGNs.
+    // 7) Collect UGN edges over TCP and aggregate locally.
+
+    info!("Bringing up links...");
+    let mut link_startups = [LinkStartup::new(); LINK_COUNT];
+    while !link_startups.iter().all(|ls| ls.is_done()) {
+        for (i, link_startup) in link_startups.iter_mut().enumerate() {
+            link_startup.next(transceivers, handshakes, i, elastic_buffers[i]);
+        }
+    }
```
hydrolarus: Would it make sense to bound this loop?

firmware-binaries/sim-tests/tcp_simultaneous_open_test/build.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2025 Google LLC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use bittide_build_utils::standard_memmap_build;
+
+fn main() {
+    standard_memmap_build("RingBufferTest.json", "DataMemory", "InstructionMemory");
```
hydrolarus, line 8: Is this the right json?

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
```diff
@@ -0,0 +1,352 @@
...
+    tcp_tx_buffer: [u8; BUF_SIZE],
+    socket_storage: MaybeUninit<[SocketStorage<'static>; 1]>,
+
+    // Always initialized
+    device: RingBufferDevice<RxRb, TxRb>,
+    timer: bittide_hal::shared_devices::Timer,
+
+    // Initialized by connect()
+    iface: MaybeUninit<Interface>,
+    sockets: MaybeUninit<SocketSet<'static>>,
```
hydrolarus: I feel like using `Option` and filling it in in `connect()` is the safer and better approach. Was a reason to use `MaybeUninit`?

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
```diff
@@ -0,0 +1,352 @@
...
+            socket_handle: smoltcp::iface::SocketHandle::default(),
+            last_state: tcp::State::Closed,
+        }
+    }
+
+    /// Initialize smoltcp and start TCP simultaneous open.
+    ///
+    /// # Safety
+    /// After this call the struct contains self-referential borrows.
+    /// It must not be moved (e.g. via `Vec::push` or assignment).
```
hydrolarus: This should then use the "pin" machinery

https://doc.rust-lang.org/core/pin/index.html

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
```diff
@@ -0,0 +1,315 @@
...
+use smoltcp::phy::{self, Device, DeviceCapabilities, Medium};
+use smoltcp::time::Instant;
+
+/// Size of packet header: CRC32 (4 bytes) + sequence (2 bytes) + length (2 bytes)
+const PACKET_HEADER_SIZE: usize = 8;
+
+/// Minimum IP packet size (20 byte IPv4 header minimum)
+const MIN_IP_PACKET_SIZE: usize = 20;
+
+/// CRC-32 (Castagnoli) instance for packet integrity checking.lgienalized at compile-time for zero-cost runtime usage.
```
hydrolarus: > lgienalized

is that a word? 😅

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    fn receive(&mut self, _timestamp: Instant) -> Option<(Self::RxToken<'_>, Self::TxToken<'_>)> {
+        // Disable RX buffer to prevent it from advancing while we process the buffer contents
+        if !(self.rx_buffer.buffer.get_enable()) {
+            error!("RX buffer is enabled at start of receive - this should not happen. Enabling and returning None.");
+            self.rx_buffer.buffer.set_enable(true);
+            return None;
+        }
+        self.rx_buffer.buffer.set_enable(false);
+        core::sync::atomic::fence(core::sync::atomic::Ordering::AcqRel);
```
hydrolarus, line 143: Our core doesn't support any atomic operations (it lacks the A extension), are fences supported and do what they should?

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
```diff
@@ -0,0 +1,315 @@
...
+
+        // Get direct pointer to hardware buffer (base pointer points to array of [u8; 8])
+        let hw_buffer_ptr = self.rx_buffer.buffer.base_ptr() as *const u8;
+
+        unsafe {
+            // Read header fields directly from hardware buffer using volatile reads
+            // Header format: CRC32 (4 bytes) + sequence (2 bytes) + length (2 bytes)
+            let crc = *(hw_buffer_ptr as *const u32);
+            let seq_num = *(hw_buffer_ptr.add(4) as *const u16);
+            let packet_len = *(hw_buffer_ptr.add(6) as *const u16) as usize;
```
hydrolarus: in the `TxToken` it explicitly uses `.to_le()` to write those numbers, should this use `uXX::from_le(...)` for symmetry?

firmware-support/bittide-sys/src/net_state.rs
```diff
@@ -0,0 +1,127 @@
...
+//     Some(UgnEdge::from(*wire))
+// }
+
+// pub fn ip_for_link(role: NodeRole, port: usize) -> [u8; 4] {
+//     let host = match role {
+//         NodeRole::Manager => 1,
+//         NodeRole::Subordinate => 2,
+//     };
+//     [10, 0, port as u8, host]
+// }
```
hydrolarus: Is this commented out code needed?

hydrolarus (commented):
Took a first look at the Rust code changes

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
lmbollen: Yes! :)

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
lmbollen: Yup!

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
lmbollen: It's for the `*LOGGER.get()`

firmware-binaries/sim-tests/tcp_simultaneous_open_test/build.rs
lmbollen: Ah the test did not make it in the PR, I re-added it. This is the intended memory map!

bittide-instances/src/Bittide/Instances/Tests/RingBuffer.hs
```diff
@@ -33,13 +34,17 @@ import Bittide.SharedTypes (withLittleEndian)
...
+import qualified Protocols.Vec as Vec
 
 createDomain vSystem{vName = "Slow", vPeriod = hzToPeriod 1_000_000}
 
 -- | Memory depth for the ringbuffers (16 entries of 8 bytes each)
 memDepth :: SNat 16
 memDepth = SNat
 
+replicateC :: forall n a b. SNat n -> Circuit a b -> Circuit (Vec n a) (Vec n b)
+replicateC SNat = fmapC
```
martijnbastiaan, line 49: 🤯 _what_

docs/sections/demos/soft-ugn-procedure.md
```diff
@@ -49,7 +49,7 @@ Events are scheduled according to the send and receive periods:
 ## Procedure Operation
 
 1.  **Initialization**:
-    *   The MU aligns the ring buffers leveraging the [Ring Buffer Alignment Procedure](../ringbuffer-alignment.md).
+    *   The MU aligns the ring buffers  leveraging the [Ring Buffer Alignment Procedure](../ring-buffer-alignment.md).
```
martijnbastiaan:
```suggestion
    *   The MU aligns the ring buffers leveraging the [Ring Buffer Alignment Procedure](../ring-buffer-alignment.md).
```

PR rejected

firmware-binaries/demos/async-comms-demo-boot/src/main.rs
```diff
@@ -0,0 +1,39 @@
+#![no_std]
+#![cfg_attr(not(test), no_main)]
+
+// SPDX-FileCopyrightText: 2025 Google LLC
```
martijnbastiaan: Could you please check all the new files and make it "202**6** Google LLC"?

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+use bittide_sys::smoltcp::link_interface::LinkInterface;
+use bittide_sys::smoltcp::link_protocol::{Command, CommandWire, UgnEdgeWire};
+use core::fmt::Write;
+use log::{debug, error, info, warn, LevelFilter};
+use riscv::register::{mcause, mepc, mtval};
+use ufmt::uwriteln;
+
+const INSTANCES: DeviceInstances = unsafe { DeviceInstances::new() };
+const LINK_COUNT: usize = 7;
+const MANAGER_DNA: [u8; 12] = [133, 129, 48, 4, 64, 192, 105, 1, 1, 0, 2, 64];
```
martijnbastiaan: Where does this magic number come from? If it _is_ magic, could you mark it as `MANAGER_DNA_MAGIC`? That makes it clear that it is just an arbitrary number. If it related to the FPGA DNA, please add a comment.

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+    ];
+    let capture_ugns = &INSTANCES.capture_ugns;
+    // Pseudocode setup:
+    // 1) Initialize MU peripherals and scatter/gather calendars for ring_buffers.
+    // 2) Align ring_buffers on all ports (two-phase protocol).
+    // 3) Run LinkStartup per port to bring up physical links and capture UGNs.
+    // 4) Wait for clock stability; stop auto-centering and record EB deltas.
+    // 5) For each port, create RingBufferDevice + smoltcp Interface with static IP.
+    // 6) Run manager state machines to connect to neighbors and request UGNs.
+    // 7) Collect UGN edges over TCP and aggregate locally.
```
martijnbastiaan: This doesn't match the code. The code seems to align ring buffers _after_ calling `LinkStartup`.

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+        (stop_time - start_time)
+    )
+    .unwrap();
+
+    // Create all LinkInterfaces
+    info!("Creating LinkInterfaces");
+    let timer = INSTANCES.timer;
+    let make_timer = || unsafe { bittide_hal::shared_devices::Timer::new(INSTANCES.timer.0) };
+
+    let mut links: heapless::Vec<LinkInterface<_, _, 256>, LINK_COUNT> = rx_aligned
```
martijnbastiaan: Factor out the magic `256`, or comment on why it is what it is (if it is a local property).

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
lmbollen: I tried this approach along with pinning, but just got rid of the self referencing instead since it made the code a lot cleaner.

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
lmbollen: I could not get this in a state I wanted, so I got rid of the self referencing.

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    info!("Waiting for stability...");
+    loop {
+        let stable = cc.links_stable()[0];
+        if stable != 0 {
+            break;
+        }
+    }
+
+    info!("Stopping auto-centering...");
```
martijnbastiaan, line 111: Two things:

1. You should link to https://github.com/bittide/bittide-hardware/issues/1173
2. You never enable auto-centering again, making this test a real "happy" path. In a real system we'd like to do async comms while the system is still synchronizing in terms of frequency. We'd either have to have hardware accelerated alignment options, or continuously poll and account for auto-center activity. It's fine to push this to another PR, but it should be obvious from the code (i.e. a comment) that this is the plan.

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
lmbollen: Nope nope nope

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+    let timer = INSTANCES.timer;
+    let make_timer = || unsafe { bittide_hal::shared_devices::Timer::new(INSTANCES.timer.0) };
+
+    let mut links: heapless::Vec<LinkInterface<_, _, 256>, LINK_COUNT> = rx_aligned
+        .into_iter()
+        .zip(tx_buffers)
+        .map(|(rx, tx)| LinkInterface::new(rx, tx, make_timer()))
+        .collect();
+    for link in links.iter_mut() {
+        link.connect();
```
martijnbastiaan: `connect()` sounds like the connection is setup after this call. Shouldn't it be called `start_connect` or something?

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
lmbollen: So this was a point of discussion @rslawson and I had and I even made an issue about it (which turned out to not be necessary as there was no real problem) https://github.com/smoltcp-rs/smoltcp/issues/1134

I was not sure how I'd have to verify it, but as far as I'm aware this fence prevents the compiler from moving stores / loads across the fence boundary, preserving the order of accesses. I'll add a comment :)

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+        );
+        for (i, dna) in partner_dnas.iter().enumerate() {
+            if dna.is_none() {
+                error!("    Link {}: no DNA received", i);
+            }
+        }
+    }
+
+    // Step 3: Build complete UGN report from all neighbor information
+    info!("Step 3: Building UGN report...");
```
martijnbastiaan, line 300: This application skirts the line between a demo and a test case. I'd rather have it be more like the latter to increase our confidence. Could you add code that checks whether these values (DNAs) are as we expect? I'd imagine this means embedded the info from `Setup.hs` and doing a lookup based on an application's own DNA.

martijnbastiaan: Alternatively, you could test it in `Driver.hs`, but that would mean a bunch of UART parsing which is meh.

firmware-binaries/sim-tests/tcp_simultaneous_open_test/src/main.rs
```diff
@@ -0,0 +1,303 @@
+// SPDX-FileCopyrightText: 2025 Google LLC
+//
+// SPDX-License-Identifier: Apache-2.0
+#![no_std]
```
martijnbastiaan: This module / `main` should have a description. I can't review code I don't know the intention for.

firmware-support/bittide-sys/src/net_state.rs
lmbollen: Nope, thanks :)

firmware-support/bittide-cpus/src/clock_control.rs
```diff
@@ -49,6 +49,18 @@ impl DomainDiffCountersInterface
     }
 }
 
+impl DomainDiffCountersInterface
```
martijnbastiaan, line 52: @rslawson @hydrolarus Is "Interface" a rusty name for this kind of construct? Without knowing too much about the ecosystem, I'd prefer "GenericX" or "XGeneric". Or maybe "PolyX"?

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
```diff
@@ -0,0 +1,352 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SendError {
+    WouldBlock,
+    NotEstablished,
+    Closed,
+}
+
+const LINK_IP: IpAddress = IpAddress::v4(100, 100, 100, 100);
+const LINK_PORT: u16 = 8080;
```
martijnbastiaan, line 37: Do IPs really matter? As I understand it we hijack smoltcp's TCP stack, and don't really care about routing. If this is true, that should be made clear by a comment. If it isn't, you should add a comment on how we make different IPs for different FPGAs.

firmware-binaries/sim-tests/tcp_simultaneous_open_test/src/main.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    let mut rx_aligned0 = AlignedReceiveBuffer::new(rx_buffer0);
+    let mut rx_aligned1 = AlignedReceiveBuffer::new(rx_buffer1);
+    while !(rx_aligned0.is_aligned() & rx_aligned1.is_aligned()) {
+        rx_aligned0.align_step(&tx_buffer0);
+        rx_aligned1.align_step(&tx_buffer1);
+    }
+
+    // Create two links with TCP simultaneous open
+    let mut link0 = LinkInterface::<_, _, 512>::new(rx_aligned0, tx_buffer1, timer);
```
martijnbastiaan: Please move the constant to an intelligible constant.

firmware-binaries/sim-tests/tcp_simultaneous_open_test/src/main.rs
```diff
@@ -0,0 +1,303 @@
...
+    link1.connect();
+
+    info!("Links created, waiting for connection...");
+    info!("Link0 initial state: {:?}", link0.state());
+    info!("Link1 initial state: {:?}", link1.state());
+
+    // Poll until both links are established
+    for _ in 0..10000 {
+        link0.poll();
+        link1.poll();
```
martijnbastiaan: I think this highlights the need to switch to `async`. Could you add a comment? Maybe in the source, maybe on the issue talking about `embassy`.

firmware-binaries/sim-tests/tcp_simultaneous_open_test/src/main.rs
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }
+
+    // Report results
+    if node0_received && node1_received && struct0_received && struct1_received {
+        info!("=== All tests passed! ===");
+        uwriteln!(uart, "=== Simultaneous Open Success ===").unwrap();
+    } else {
+        info!("=== Test failed ===");
+        uwriteln!(uart, "=== Test Failed ===").unwrap();
```
martijnbastiaan: Prepend this with `[ERROR]`. That way the GitHub UI will highlight it.

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
```diff
@@ -0,0 +1,352 @@
...
+
+/// TCP communication over a pair of ring_buffers.
+///
+/// Construction is two-phase to allow the struct to own its TCP buffers:
+/// 1. `new()` — allocates buffers, stores the device and timer
+/// 2. `connect()` — initializes smoltcp and starts TCP simultaneous open
+///
+/// `connect()` creates internal borrows into the struct's own buffer arrays,
+/// so the struct must not be moved after `connect()` is called.
```
martijnbastiaan: Hmm. This seems very suspicious. What do you meant with this? Shouldn't the `move` keyword be used on this? If so, isn't Rust supposed to enforce memory issue like that?

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
```diff
@@ -0,0 +1,352 @@
...
+    ///
+    /// # Safety
+    /// After this call the struct contains self-referential borrows.
+    /// It must not be moved (e.g. via `Vec::push` or assignment).
+    pub fn connect(&mut self) {
+        // SAFETY: We create 'static references into our own buffer fields.
+        // Sound because:
+        //  - The struct is not moved after this call (caller obligation).
+        //  - In embedded main() -> !, the stack frame lives forever.
+        unsafe {
```
martijnbastiaan: This `unsafe` block is HUGE. Is it truly needed for all statements in here?

firmware-support/bittide-sys/src/smoltcp/link_protocol.rs
```diff
@@ -0,0 +1,87 @@
...
+//! Wire Format Types for Link Protocol
+//!
+//! This module provides wire format types for communication between nodes.
+//! Use these with LinkInterface's zerocopy send/recv methods.
+
+use crate::net_state::UgnEdge;
+use zerocopy::byteorder::{I64, LE, U32};
+use zerocopy::{AsBytes, FromBytes, FromZeroes, Unaligned};
+
+pub const DNA_BYTES: usize = 12;
```
martijnbastiaan:
```suggestion
pub const N_DNA_BYTES: usize = 12;
```

martijnbastiaan: Or `_COUNT`.

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
```diff
@@ -0,0 +1,315 @@
...
+//! - Header (8 bytes): CRC32 (4 bytes LE) + sequence number (2 bytes LE) + length (2 bytes LE)
+//! - Payload (variable, up to MTU)
+//!
+//! The CRC32 is calculated over sequence + length + payload (i.e., everything except the CRC itself).
+//!
+//! # Volatile Buffer Handling
+//!
+//! The receive ring_buffer is volatile - its contents can change at any time due to
+//! incoming data. To safely handle this, we:
+//! 1. Disable the buffer to prevent it from advancing during processing
```
martijnbastiaan: Hmm, the wording is quite ambiguous here. I think you should say "disable writes" and highlight that the ring buffer itself keeps on spinning.

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
```diff
@@ -0,0 +1,315 @@
...
+        let mut cap = DeviceCapabilities::default();
+        cap.max_transmission_unit = self.mtu;
+        cap.medium = Medium::Ip;
+        cap
+    }
+
+    fn receive(&mut self, _timestamp: Instant) -> Option<(Self::RxToken<'_>, Self::TxToken<'_>)> {
+        // Disable RX buffer to prevent it from advancing while we process the buffer contents
+        if !(self.rx_buffer.buffer.get_enable()) {
+            error!("RX buffer is enabled at start of receive - this should not happen. Enabling and returning None.");
```
martijnbastiaan: Either return a `Result` or panic. Don't warn and continue. If you do return a `Result` it should be clear from the documentation what state the system is in after.

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
```diff
@@ -0,0 +1,315 @@
...
+    Tx: TransmitRingBufferInterface + 'static,
+{
+    /// Create a new ring_buffer device.
+    ///
+    /// The ring_buffers must already be aligned using the alignment protocol.
+    /// The MTU is calculated as the minimum of the RX and TX buffer sizes in bytes.
+    pub fn new(rx_buffer: AlignedReceiveBuffer<Rx, Tx>, tx_buffer: Tx) -> Self {
+        // Calculate MTU from buffer sizes (each word is 8 bytes)
+        // Reserve space for packet header (CRC is part of header)
+        let mtu = (Rx::DATA_LEN * 8).min(1500).min(Tx::DATA_LEN * 8) - PACKET_HEADER_SIZE;
```
martijnbastiaan: Inline constant (1500) again.

firmware-binaries/sim-tests/ring_buffer_smoltcp_test/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+use log::{info, LevelFilter};
+use ufmt::uwriteln;
+
+#[cfg(not(test))]
+use riscv_rt::entry;
+
+const INSTANCES: DeviceInstances = unsafe { DeviceInstances::new() };
+
+#[cfg_attr(not(test), entry)]
+fn main() -> ! {
```
martijnbastiaan: Is this ever executed?

firmware-binaries/sim-tests/ring_buffer_smoltcp_test/src/main.rs
```diff
@@ -0,0 +1,429 @@
+// SPDX-FileCopyrightText: 2025 Google LLC
+//
+// SPDX-License-Identifier: Apache-2.0
```
martijnbastiaan: This module (or `main`) should have documentation. I don't know how to review if I don't know the intention of the code.

firmware-binaries/sim-tests/tcp_simultaneous_open_test/src/main.rs
```diff
@@ -0,0 +1,303 @@
...
+        Self {
+            id: msg.id.get(),
+            value: msg.value.get(),
+            flags: msg.flags.get(),
+        }
+    }
+}
+
+#[cfg_attr(not(test), entry)]
+fn main() -> ! {
```
martijnbastiaan, line 82: Is this ever executed?

firmware-binaries/sim-tests/ring_buffer_smoltcp_test/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+            break;
+        }
+    }
+
+    if !manager_received || !subordinate_received {
+        info!(
+            "  FAILURE: DNA exchange failed! Manager={}, Subordinate={}",
+            manager_received, subordinate_received
+        );
+        uwriteln!(uart, "=== Test Failed ===").unwrap();
```
martijnbastiaan: Could you please open an issue stating we should formalize this into a framework? We keep copying strings like these everywhere.

firmware-binaries/sim-tests/ring_buffer_smoltcp_test/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+                info!("    Subordinate: sent count={}", subordinate_report.count);
+            }
+        }
+
+        // Manager: try to receive count
+        if !count_received {
+            let mut count_buf = [0u8; 1];
+            if let Ok(1) = link_manager.try_recv_bytes(&mut count_buf) {
+                manager_count = count_buf[0];
+                info!("    Manager: received count={}", manager_count);
```
martijnbastiaan: Should it error when manager_count ~ 0?

firmware-binaries/demos/async-comms-demo-management-unit/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+            link_startup.next(transceivers, handshakes, i, elastic_buffers[i]);
+        }
+    }
+
+    info!("Waiting for stability...");
+    loop {
+        let stable = cc.links_stable()[0];
+        if stable != 0 {
+            break;
+        }
```
martijnbastiaan: This exists if _any_ link is stable, not all links, which I think is your intention.

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
```diff
@@ -0,0 +1,315 @@
...
+            return None;
+        }
+        self.rx_buffer.buffer.set_enable(false);
+        core::sync::atomic::fence(core::sync::atomic::Ordering::AcqRel);
+
+        // Get direct pointer to hardware buffer (base pointer points to array of [u8; 8])
+        let hw_buffer_ptr = self.rx_buffer.buffer.base_ptr() as *const u8;
+
+        unsafe {
+            // Read header fields directly from hardware buffer using volatile reads
```
martijnbastiaan: Where does it do volatile reads?

firmware-binaries/sim-tests/switch_demo_pe_test/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+bittide-hal = { path = "../../../firmware-support/bittide-hal" }
+riscv-rt = "0.11.0"
+ufmt = "0.2.0"
+
+[dependencies.log]
+version = "0.4.21"
+features = ["max_level_trace", "release_max_level_info"]
+
+[build-dependencies]
+bittide-build-utils = { path = "../../../firmware-support/bittide-build-utils" }
```
martijnbastiaan: Rebase artifact? Delete this file.

martijnbastiaan (changes requested):
Round 1! I've got more review comments, but I've already accumulated 30. Paired with the ones from @hydrolarus that should already be quite a lot to handle.

firmware-binaries/sim-tests/ring_buffer_smoltcp_test/src/main.rs
lmbollen: https://github.com/bittide/bittide-hardware/issues/1331

bittide-instances/src/Bittide/Instances/Tests/RingBuffer.hs
```diff
@@ -103,6 +125,20 @@ simResultRingBuffer latency = do
           $ (dutWithPeConfig @System latency peConfig)
           -< mm
       idC -< uartTx
-    uartStream = sampleC def{timeoutAfter = 1_000_000} dutNoMm
+    uartStream = sampleC def{timeoutAfter = timeout} dutNoMm
```
martijnbastiaan:
```suggestion
    uartStream = sampleC def{timeoutAfter} dutNoMm
```

Tiny thing, but if you rename the argument then this becomes simpler.

bittide-instances/tests/Wishbone/RingBuffer.hs
```diff
@@ -39,5 +40,22 @@ prop_ring_buffer_test =
     H.annotate [i|Result of ring_buffer_test with latency #{latency} cycles: \n#{result}|]
     H.assert ("TEST PASSED" `isInfixOf` result)
 
+prop_tcp_open_test :: H.Property
+prop_tcp_open_test =
+  -- This test is _very_ slow, so we only run it once.
+  H.withTests 1 $ H.property $ do
```
martijnbastiaan, line 46: Can you comment that this picks size ~ 100.

firmware-binaries/sim-tests/tcp_simultaneous_open_test/src/main.rs
```diff
@@ -0,0 +1,344 @@
...
+        if link0.is_established() && link1.is_established() {
+            info!("Both links established!");
+            info!("Link0 state: {:?}", link0.state());
+            info!("Link1 state: {:?}", link1.state());
+            break;
+        }
+    }
+
+    if !link0.is_established() || !link1.is_established() {
+        info!("Connection timeout!");
```
martijnbastiaan:
```suggestion
        error!("Connection timeout!");
```

martijnbastiaan (approved):
Thanks for applying the suggestions! I thought I had more, but I don't remember now..

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
```diff
@@ -0,0 +1,329 @@
...
+        }
+    }
+
+    fn socket(&self) -> &tcp::Socket<'a> {
+        self.sockets.get::<tcp::Socket>(self.socket_handle)
+    }
+
+    fn socket_mut(&mut self) -> &mut tcp::Socket<'a> {
+        self.sockets.get_mut::<tcp::Socket>(self.socket_handle)
+    }
```
hydrolarus, line 109: Would it make sense to use `Deref<Target = tcp::Socket<'a>>` and `DerefMut` to get rid of a lot of the accessor functions below? (like `is_open()` or `is_active()`)

firmware-support/bittide-sys/src/smoltcp/link_interface.rs
```diff
@@ -0,0 +1,329 @@
...
+            // Only consume bytes once a complete T has arrived; a partial read
+            // would permanently corrupt the TCP stream on retry.
+            if socket.recv_queue() < size {
+                return Err(RecvError::WouldBlock);
+            }
+        }
+        let mut buffer = T::new_zeroed();
+        self.socket_mut()
+            .recv_slice(buffer.as_bytes_mut())
+            .unwrap_or(0);
```
hydrolarus, line 301: this last `.unwrap_or(0)` doesn't seem to have any effect, it's just ignored right after. If you want to ignore the result then `_ = self.socket_mut()...` is the more intentional way to express this.

firmware-support/bittide-sys/src/smoltcp/ring_buffer.rs
```diff
@@ -0,0 +1,336 @@
...
+        self.rx_buffer.buffer.set_enable(false);
+
+        // Make sure loads and stores do not cross this boundary to ensure correct order of
+        // hardware accesses.
+        core::sync::atomic::fence(core::sync::atomic::Ordering::AcqRel);
+
+        // Get direct pointer to hardware buffer (base pointer points to array of [u8; 8])
+        let hw_buffer_ptr = self.rx_buffer.buffer.base_ptr() as *const u8;
+
+        unsafe {
```
hydrolarus, line 148: This is a really large unsafe block, I'm sure not all of this needs to be in there, but I can't see from a glance what exactly needs this. So I think this should be split up into smaller blocks where they are actually needed.

hydrolarus (commented):
Seems fine! Some form suggestions

Issue #1199, opened 2026-02-18
Decouple transceiver / handshake logic


martijnbastiaan:
We currently have a mono-block that:

* Gets the transceivers up and running
* Does word alignment
* Implements a user-handshake mechanism

The handshake information travels through the elastic buffer. In the future we might not have the luxury of doing this and we might be restricted to run only after the elastic buffers. To completely de-couple, the infrastructure needs to look like:

![Image](https://github.com/user-attachments/assets/d7ea62f3-e87d-4b39-b2b3-2a5adee8bebf)

All of "handshake", "ugn capture", "tx ring buffer", and "rx ring buffer" be transparent by default (excluding pipeline registers).

For now it seems like relying on word alignment is fine.

PR #1229, opened 2026-03-25
Continue separating out handshake from transceiver

enathang:
_What (what did you do)_
This PR continues the work from https://github.com/bittide/bittide-hardware/pull/1226 to separate handshake and transceiver.

This PR:
 - Moves handshake into its own module `Transceiver.Handshake`
 - Completely separates transceiverPrbsN and (newly created) userDataHandshakeN, including creating new data types for both functions
 - Modifies the unit and HITL tests to still compile/pass by instantiating handshake in addition to transceiver. In the future, these tests can be changed to test either `transceiver` or `handshake` in isolation, but for now they're basically tested together.

Work that still needs to be done:
 - The transceiver and handshake are now completely separated in non-circuit functions. However, the bittide HITL tests use the circuit versions (`transceiverPrbsNC` and `transceiverPrbsNWb`), which still have the transceiver and handshake connected inside. These two functions need to be refactored (and create new `handshakeNC` and `handshakeNWb` functions), and then the HITL tests need to be modified to use the refactored functions.
 - Clean up dead code (specifically `Input`, `Inputs`, `Output`, `Outputs` in `Transceiver.hs` and do renaming
 - Then, one needs to move the handshake in front of the ElasticBuffer to behind it. I've already created a new handshake (`src/Bittide/Handshake.hs`), which does not do any CDCs and SHOULD (famous last words) work behind the ElasticBuffer. But, if there are bugs introduced, the other option is taking the existing `Transceiver.Handshake.hs` and simply eliminating the CDCs.

_Why (context, issues, etc.)_
https://github.com/bittide/bittide-hardware/issues/1199

- The functions are now cleanly separated, but the namespaces are still a bit mangled. Ie `HandshakeInput` inside `Transceiver/Handshake.hs` could probably be renamed `Input` and imported qualified. Similarly, there are both `TransceiverInput`, `TransceiverInputs`, `TransceiverOutput`, `TransceiverOutputs` in addition to `Input`, `Inputs`, `Output`, `Outputs` in `Transceiver.hs`. Once the circuit notation stuff is refactored (see top), the vestigial `Input`, `Inputs`,... can be removed.

- I know I'm going to catch flak for the nesting of `fromHandshake` inside `TransceiverInput` and `fromTransceiver` inside `HandshakeInput`, etc. These were mostly to help me keep track of which wires came from/went where. I'll PROBABLY remove them, but first I'm trying to sneak it through without doing the extra work :)

- This pr should be updated to run the Transceiver HITL tests in the `debug.json`, since this PR modifies those tests but doesn't automatically run them.

None

# TODO

- [x] Write (regression) test
- [x] Link to existing issue


martijnbastiaan:
I'm going to take a few shortcuts, which will make this PR obsolete. Thanks for the work @enathang and sorry for the unfortunate end.

PR #1244, opened 2026-04-13
Add Df based wrappers for `tdpBram`

lmbollen:
_What (what did you do)_
Adds Df based wrappers for `tdpBram` compatible true dual ported blockRams.

_Why (context, issues, etc.)_
The CPU needs both read and write access to the transmit buffer to calculate the checksum. (and we need to be able to continuously read content from the same buffer.

Am I missing relevant test coverage of the wrapper?

Rewrite as mealy mostly done by AI

# TODO

- [x] ~Write (regression) test~
- [ ] ~Update documentation, including `docs/`~
- [ ] ~Link to existing issue~

### DIFF
--- a/bittide-extra/src/Protocols/Df/Extra.hs
+++ b/bittide-extra/src/Protocols/Df/Extra.hs
@@ -160,30 +160,38 @@ fromDualPortedBramWithMask prim clkA clkB = Circuit goS
     (rightOpAck, rightOp1, rightDat1) = goChannel clkB (rightOp0, rightDat0, rightDatAck)
     (leftDat0, rightDat0) = prim leftOp1 rightOp1
 
-  goChannel clk = E.mealyB clk E.noReset enableGen goT RamNoOp
+  goChannel clk = E.mealyB clk E.noReset enableGen goT Nothing
 
-  goT lastOp (maybeOp, ramData, Ack ack) = (op1, (Ack opAck, op1, dat1))
+  goT lastRead (maybeOp, ramData, Ack ack) = (nextState, (Ack opAck, outgoingOp, dat1))
    where
     -- Construct rhs data from previous cycle's operation
     dat1
-      | isRead lastOp = Just ramData
+      | isJust lastRead = Just ramData
       | otherwise = Nothing
 
     -- Determine if we receive backpressure
     stall = isJust dat1 && not ack
 
     -- If we receive backpressure on our rhs, we have to reissue our ramop.
-    op1
-      | stall = lastOp
+    outgoingOp
+      | stall = addrToOp lastRead
       | otherwise = fromMaybe RamNoOp maybeOp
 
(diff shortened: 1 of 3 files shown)


martijnbastiaan:
What do you mean writing a test doesn't apply? :P You even wrote one!

bittide-extra/src/Protocols/Df/Extra.hs
```diff
@@ -75,6 +76,115 @@ skid = Circuit go
 ackWhen :: Signal dom Bool -> Circuit (Df dom a) ()
 ackWhen canDrop = Circuit $ \_ -> (Ack <$> canDrop, ())
 
+-- | Creates a wrapper around `tdpbram` to make it work on `RamOp
```
martijnbastiaan:
```suggestion
-- | Creates a wrapper around 'tdpbram' to make it work on 'RamOp'
```

Backticks don't produce Haddock

lmbollen:
But it was not a regression test  :eyes:

bittide-extra/tests/unittests/Tests/Protocols/Df/Extra.hs
```diff
@@ -308,5 +312,67 @@ splitWriteInBytes Nothing _ = repeat Nothing
 
 -- End of shamelessly copied code from bittide
 
+deriving instance (NFData a) => NFData (RamOp addr a)
+deriving instance (ShowX a) => ShowX (RamOp addr a)
+deriving instance (Eq a) => Eq (RamOp addr a)
```
martijnbastiaan, line 317: Please open a PR on `clash-compiler` :)

bittide-extra/src/Protocols/Df/Extra.hs
```diff
@@ -75,6 +76,115 @@ skid = Circuit go
...
+    -- Determine if we receive backpressure
+    stall = isJust dat1 && not ack
+
+    -- If we receive backpressure on our rhs, we have to reissue our ramop.
+    op1
+      | stall = lastOp
+      | otherwise = fromMaybe RamNoOp maybeOp
+
+    -- Determine if we ack our lhs
+    opAck = isWrite op1 || isNothing dat1 || ack
```
martijnbastiaan: I'm confused: you can't ACK a write if you're still trying to get data to the RHS and you're getting backpressure right? Because that would mean you're still emitting reads to the RAM.

My expectation was that the logic would be roughly as follows:

```haskell
opAck = lastOp == RamNoOp || ack
newOp
  | isRead op1 = op1
  | otherwise = RamNoOp
```

I.e., if your "command buffer" is empty, you can always acknowledge even if the RHS hasn't received anything yet.

martijnbastiaan (changes requested):

martijnbastiaan (approved):
LGTM, modulo minor comment we talked about on Slack!

PR #1246, opened 2026-04-13
Introduce typeclasses for vendored CDC primitives

rslawson:
_What (what did you do)_
This should implement (most of?) the desired changes as per #1222.

_Why (context, issues, etc.)_
See issue for context.

I think there's a GHC bug to be found in here. If in `Clash.Cores.Xilinx` I write
```hs
withXilinxI :: forall r. (forall vendor. (VendorCdcImplicit vendor) => r) -> r
withXilinxI = withVendorI @Xilinx
```
and then replace all `withVendorI @Xilinx` with `withXilinxI` calls, I suddenly get a bunch of compilation errors. If it's somehow possible to get this to work though, I'd appreciate a review comment (:

None.

# TODO

- [x] Link to existing issue

### DIFF
--- a/bittide-extra/src/Clash/Class/Cdc.hs
+++ b/bittide-extra/src/Clash/Class/Cdc.hs
@@ -1,0 +1,755 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
+
+{- | This module introduces typeclasses, instances, and some helper functions + type aliases for
+writing components that require CDC crossings, but it may not be desirable to use a specific CDC
+crossing. This enables the designer to write functions that are general over the many parameters
+of CDC crossings as well, including what type the input signal contains, the particular clock
+domains, and the number of stages in the CDC primitive. The goal of this is then that the
+designer is able to write another file with orphaned instances containing the primitives for a
+given vendor, which may then be used to specialize more general components.
+
+To be more concrete, a designer may want to be able to write a component that requires a gray CDC
+primitive and synthesize for both Xilinx and Intel FPGAs. In this case, the designer can write that
+component with the constraint @Gray vendor InnerType SrcDomain DstDomain@ and provide the vendor
+later through the 'withVendor' or 'withVendorI' functions.
+
+N.B.: this module is written with qualified use in mind. Generally it should only be imported as
+@import qualified Clash.Class.Cdc as Cdc@ or similar.
+-}
+module Clash.Class.Cdc where
+
+import Clash.Prelude hiding (Bit, isRising, regEn)
+
+import Bittide.Extra.Maybe (orNothing)
+import Clash.Explicit.Prelude (isRising, noReset, regEn)
+import Data.Data (Proxy (Proxy))
+import Data.Maybe (fromMaybe, isJust, isNothing)
+import GHC.Stack (HasCallStack)
+
+{- | Guarantees the existence of a CDC primitive for a fixed-width array of independent bits for a
+given vendor
+-}
+class IndependentBits (vendor :: Symbol) where
+  -- | Additional constraints on the independent bits CDC primitive with a custom configuration
+  type
+    IndependentBitsWithConstraints vendor (stages :: Nat) (a :: Type) (src :: Domain) (dst :: Domain) ::
+      Constraint
+
+  -- | Configuration type for an independent bits CDC primitive
+  type IndependentBitsConfig vendor (stages :: Nat) (a :: Type) (src :: Domain) (dst :: Domain) :: Type
+
+  -- | Independent bits CDC primitive with a custom configuration
+  independentBitsWith ::
+    forall stages a src dst.
+    ( KnownDomain src
+    , KnownDomain dst
+    , HasCallStack
+    , NFDataX a
+    , BitPack a
+    , IndependentBitsWithConstraints vendor stages a src dst
+    ) =>
+    IndependentBitsConfig vendor stages a src dst ->
+    Clock src ->
+    Clock dst ->
+    Signal src a ->
+    Signal dst a
+
+  -- | Additional constraints on the independent bits CDC primitive with the default configuration
+  type IndependentBitsConstraints vendor (a :: Type) (src :: Domain) (dst :: Domain) :: Constraint
+
+  -- | Default number of stages in an independent bits CDC primitive
+  type IndependentBitsDefaultStages vendor (a :: Type) (src :: Domain) (dst :: Domain) :: Nat
+
+  -- | Default configuration for an independent bits CDC primitive
+  defaultIndependentBitsConfig ::
+    forall a src dst.
+    (KnownDomain src, KnownDomain dst) =>
+    Proxy src ->
+    Proxy dst ->
+    Proxy a ->
+    IndependentBitsConfig vendor (IndependentBitsDefaultStages vendor a src dst) a src dst
+
+  -- | Independent bits CDC primitive with the default configuration
+  independentBits ::
+    forall a src dst.
+    ( KnownDomain src
+    , KnownDomain dst
+    , HasCallStack
+    , NFDataX a
+    , BitPack a
+    , IndependentBitsConstraints vendor a src dst
+    ) =>
+    Clock src ->
+    Clock dst ->
+    Signal src a ->
+    Signal dst a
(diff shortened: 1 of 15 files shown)


bittide/src/Bittide/Sync.hs
```diff
@@ -42,9 +42,14 @@ import Protocols
...
+import Clash.Class.Cdc (
+  HasVendorCdc,
+  VendorCdc (Single, SyncRst, cdcSingle, cdcSyncRst, defaultAssert),
+ )
 import Clash.Class.Counter (Counter (countSuccOverflow))
-import Clash.Cores.Xilinx.Xpm (xpmCdcSingle, xpmCdcSyncRst)
-import Clash.Cores.Xilinx.Xpm.Cdc.SyncRst (Asserted (..))
+
+-- import Clash.Cores.Xilinx.Xpm (xpmCdcSingle, xpmCdcSyncRst)
+-- import Clash.Cores.Xilinx.Xpm.Cdc.SyncRst (Asserted (..))
```
martijnbastiaan: leftover

martijnbastiaan (changes requested):
Overall this LGTM, the GHC bug is too bad.

I do have three comments:

1. Instead of making it a big God class, could you make it into singular classes? This would help us if vendor can't supply all the primitives.
2. I get the appeal of adopting Xilinx's naming scheme, but at the same time I think that scheme doesn't really fit Clash. What do you think of:
  * `Single`: `Bit`
  * `ArraySingle`: `IndependentBits`
3. Could you please remove the `cdc` prefix from the names and design for `qualified` use? I.e., call sites would look like `Cdc.gray`.

I'm fairly certain (1) is the way to go, less so of (2), and unswayable on (3) :-)

bittide/src/Bittide/Sync.hs
lmbollen: How about:
`ArraySingle`: `BitVector` instead of `IndependentBits` 

To bikeshed a bit I don't like the `Vendor` name in the API, I know that the usecase will most likely be implementing vendor instances, but I'd rather have something descriptive :)

martijnbastiaan: Right, the class can be named `Cdc`. 

For `BitVector`: The problem is that it's really easy to assume that the bitvector will get CDCed in its entirety. However, for these functions each bit gets CDCed individually, meaning that the words you put into the functions are not necessarily observed on the other side. With `IndependentBits` you highlight that property.

martijnbastiaan: But yeah it's verbose.

rslawson:
Made changes (1) and (3), plus introduced a couple other convenience typeclasses so that it's easier to get the right constraints into scope so that you can call the CDC primitives' functions. I'm a little less enthusiastic about change (2), but I do see the reasoning for it so I did make that change.

bittide/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake/Extra.hs
```diff
@@ -22,6 +21,7 @@ xpmCdcMaybeLossy ::
   , NFDataX a
   , 1 <= BitSize a
   , BitSize a <= 1024
+  , Cdc.ValidHandshake Xilinx a src dst
```
martijnbastiaan, line 18: Hmm, why doesn't that resolve without this constraint? Shouldn't the `1 <= BitSize a` already take care of that?

bittide-extra/src/Clash/Class/Cdc.hs
```diff
@@ -0,0 +1,697 @@
...
+class
+  (Pulse vendor, PulseWithCst vendor stages a src dst) =>
+  ValidPulseWith (vendor :: Symbol) (stages :: Nat) (a :: Type) (src :: Domain) (dst :: Domain)
+
+instance
+  (Pulse vendor, PulseWithCst vendor stages a src dst) =>
+  ValidPulseWith (vendor :: Symbol) (stages :: Nat) (a :: Type) (src :: Domain) (dst :: Domain)
+
+class Bit (vendor :: Symbol) where
+  -- | Additional constraints on the single bit CDC primitive without the default configuration
```
martijnbastiaan:
```suggestion
  -- | Additional constraints on the single bit CDC primitive with custom configuration
```

Makes the name "With" and documentation match.

bittide-extra/src/Clash/Class/Cdc.hs
```diff
@@ -0,0 +1,697 @@
...
+  type BitWithCst vendor (stages :: Nat) (a :: Type) (src :: Domain) (dst :: Domain) :: Constraint
```
martijnbastiaan:
```suggestion
  type BitWithConstraints vendor (stages :: Nat) (a :: Type) (src :: Domain) (dst :: Domain) :: Constraint
```

It's more verbose, but: these constraints are very likely to go on their own line anyway and `Cst` isn't a well known abbreviation of `Constraint`(s).

bittide-extra/src/Clash/Class/Cdc.hs
```diff
@@ -0,0 +1,697 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
+
+module Clash.Class.Cdc where
```
martijnbastiaan, line 24: Could you add module documentation? It should cover what it is and why we want it.

martijnbastiaan: Also explain the general methodology: that each class has a "with" and no-"with" function, etc.

bittide-extra/src/Clash/Class/Cdc.hs
```diff
@@ -0,0 +1,697 @@
...
+
+import Clash.Prelude hiding (Bit, isRising, regEn)
+
+import Bittide.Extra.Maybe (orNothing)
+import Clash.Explicit.Prelude (isRising, noReset, regEn)
+import Data.Data (Proxy (Proxy))
+import Data.Maybe (fromMaybe, isJust, isNothing)
+import GHC.Stack (HasCallStack)
+
+class IndependentBits (vendor :: Symbol) where
```
martijnbastiaan, line 37: Could you add documentation to each of the the functions? They should cover what they synchronize and under what circumstances they can be used. If it helps, this is Xilinx's decision tree:

Image

This one is also very cool, but more out-of-scope:

Image

bittide-extra/src/Clash/Class/Cdc.hs
```diff
@@ -0,0 +1,697 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
martijnbastiaan, line 4:
```suggestion
```

Given that I see a lot of `Proxy`s, is this necessary?

bittide/src/Bittide/Counter.hs
```diff
@@ -1,6 +1,7 @@
 -- SPDX-FileCopyrightText: 2022 Google LLC
 --
 -- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE ImplicitParams #-}
```
martijnbastiaan, line 4: Let's enable this by default across all packages. Could you open a separate PR?

martijnbastiaan (changes requested):
Almost there.. :)

bittide/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake/Extra.hs
rslawson: Ah, it might be because the `1 <= BitSize a, BitSize a <= 1024` is equivalent to the `HandshakeCst` constraint for `Xilinx`. I should swap those out.

bittide-extra/src/Clash/Class/Cdc.hs
rslawson: Of course! Didn't mean to have basically no documentation, but remembering things is not my forte.

bittide/src/Bittide/Counter.hs
rslawson: Sure thing, will do.

bittide-extra/src/Clash/Class/Cdc.hs
rslawson: Unfortunately yes ):

martijnbastiaan (approved):

PR #1249, opened 2026-04-14
Add wire demo

### LINKED ISSUE #1109: Simplify switch demo
# Status quo

Back in April we added a [switch demo](https://github.com/bittide/bittide-hardware/pull/723) to showcase that we could use the bittide mechanism to thread a path through an 8-node network, all statically scheduled. The hardware [used to look like this](https://github.com/user-attachments/assets/f26dbb8c-bfe0-4e74-bd2d-3d9722b98401). To simplify things, we programmed a static calendar and used a [calculator](https://github.com/bittide/bittide-hardware/pull/728) to produce the exact clock cycles at which the PE should send and receive data. For more details, see [this presentation](https://docs.google.com/presentation/d/1AGbAJQ1zhTPtrekKnQcthd0TUPyQs-zowQpV1ux4k-Y/edit?usp=sharing).

# Proposal
While talking to potential users of bittide we've found that they're not interested in this switching behavior (yet). Instead, they're more interested in the "wire" behavior of bittide and would simply like to be kept away from anything else. To showcase this, I propose the following hardware design:

Image

The idea is still the same: thread a path through a bunch of nodes/FPGAs (in our case 8). The implementation is much simpler than the previous iteration though: instead of using calendars to do time multiplexing on every link, the links are connected to the PE through a single mux. This mux can either be set to serve the MU or the PE.

### Management unit
Still responsible for buffer centering, getting links up and running, etc. It will program the mux and PE config.

### Programmable mux
2 bus accessible registers:
 * `cycle_to_switch :: Unsigned 64`: cycle to switch from MU to PE
 * `arm :: Bool`: whether to actually look at the first register yet. Used to prevent atomicity issues.

Inputs:
* Links from the PE
* Links from the MU

Outputs:
* Reset to PE
* Links to elastic buffers

### PE
Takes two configuration values (from "PE config"):

- What link to read from (`:: Maybe (Index 7)`)
- What link to write to (`:: Maybe (Index 7)`)

The PE is active for two cycles after reset:

* First cycle: read from link indicated by config (or source static "0").
* Second cycle: write `value_read_in_first_cycle XOR fpga_dna_lsbs` to link indicated by config (or do nothing)

### PE Config
2 bus accessible registers:
* `read_link :: Maybe (Index 7)`
* `write_link :: Maybe (Index 7)`

## Why this shows bittide
Because the PE only reads a single cycle and writes a single cycle, we need to be able to predict exactly when data packets will arrive on any given node from any given node.

## Why this satisfies "seeing bittide as a wire"
The MU will take care of all reset logic and set everything up for the PE / business logic. The business logic does not have to care about this at all and just does a simple read, trusting that the data from its neighbor is presented on the relevant link. Vice versa for writing.

## Things we will _not_ do for this demo
We won't do in-band communication of UGNs. They are captured and read out by the host PC over a debug connection. The host PC will also calculate the schedule for each FPGA (programmable mux config, PE config). Doing this all in-band is planned, but we need to take one bite at a time.

## Things left out of the diagram
* Clock control
* Transceiver IP
* Debug connections
* ʰᵃʳᵈʷᵃʳᵉ ᵘᵍⁿ ᶜᵒᵐᵖᵒⁿᵉⁿᵗˢ
* Peripherals (timers, uart, etc.)

hiddemoll:
_What (what did you do)_
Added the wire demo as a fourth demo. See the linked issue for an overview, or check out the documentation in the mdbook.

_Why (context, issues, etc.)_
Fixes #1109

I've copy-pasted the Soft UGN Demo for the Wire Demo in a separate commit to make reviewing it easier. I have a few specific points I'd like you (the reviewer) to pay extra attention to:
1. I've moved some functions in the Switch Demo driver so I could reuse them for the Wire Demo. We already do this a lot in that driver, but it is not good practice. Should I move them, and if so, where to?
2. I fear my understanding of `internalDelay` in `toCounterMap` might be wrong. For the wire demo I had to set it to `1` ([here](https://github.com/bittide/bittide-hardware/blob/079ce6e1df20aa338eb4068decd53fe6fd484827/bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs#L65)), but I cannot find the additional register on the path between PEs which is not also on the `captureUgn` path.
3. Are [`generateSchedule`](https://github.com/bittide/bittide-hardware/blob/079ce6e1df20aa338eb4068decd53fe6fd484827/bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs#L94) and its documentation clear?
4. What should the PE write to the links while 'inactive'? Issue and internal documentation say 'write 0', but while debugging it was easier to use local counter.
5. I added `wireDemoTest` to `main.json` and moved both `switchDemoTest` and `switchDemoGppeTest` to nighly. Do you agree?

Nothing worth a heads-up

# TODO

- [x] Write (regression) test 
-> Added a test for the `wireDemoPe` and the HITL test as a demo
-> Added docs and diagram for wire demo in mdbook
- [x] Link to existing issue

### DIFF
--- a/bittide-instances/src/Bittide/Instances/Hitl/WireDemo/BringUp.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/WireDemo/BringUp.hs
@@ -1,0 +1,182 @@
+-- SPDX-FileCopyrightText: 2025 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP #-}
+
+module Bittide.Instances.Hitl.WireDemo.BringUp (bringUp) where
+
+import Clash.Explicit.Prelude
+import Clash.Prelude (HiddenClockResetEnable, withClockResetEnable)
+import Protocols
+
+import Bittide.BootPe (BootPeBusses, bootPe)
+import Bittide.CaptureUgn (sendUgnC)
+import Bittide.ClockControl
+import Bittide.Df (asciiDebugMux)
+import Bittide.Instances.Domains (
+  Basic125,
+  Bittide,
+  Ext200,
+  GthRx,
+  GthRxS,
+  GthTxS,
+ )
+import Bittide.Instances.Hitl.Setup (LinkCount)
+import Bittide.Instances.Hitl.WireDemo.Core (InternalCpuCount, core)
+import Bittide.Jtag (jtagChain, unsafeJtagSynchronizer)
+import Bittide.ProcessingElement (PeConfig (..))
+import Bittide.SharedTypes (Byte, withLittleEndian)
+import Bittide.Sync (Sync)
+import Bittide.Wishbone (arbiterMm, extendAddressWidthWb, uartDf)
+import Clash.Cores.Xilinx.DcFifo (dcFifoDf)
+import Data.Char (ord)
+import Protocols.Extra
+import Protocols.MemoryMap (Mm)
+import Protocols.Spi (Spi)
+import Protocols.Wishbone.Extra (xpmCdcHandshakeWb)
+import VexRiscv (Jtag)
+
+import qualified Bittide.Cpus.Riscv32imc as Riscv32imc
+import qualified Bittide.Transceiver as Transceiver
+import qualified Bittide.Transceiver.Wishbone as Transceiver
+import qualified Clash.Cores.Xilinx.Gth as Gth
+import qualified Protocols.Vec as Vec
+
+#ifdef SIM_BAUD_RATE
+type Baud = MaxBaudRate Basic125
+import Clash.Cores.UART.Extra
+#else
+type Baud = 921_600
+#endif
+
+baud :: SNat Baud
+baud = SNat
+
+uartLabels :: Vec 3 (Vec 2 Byte)
+uartLabels =
+  fmap (fromIntegral . ord)
+    <$> ( $(listToVecTH "BT")
+            :> $(listToVecTH "MU")
+            :> $(listToVecTH "CC")
+            :> Nil
+        )
+
+bootPeConfig :: PeConfig BootPeBusses
+bootPeConfig =
+  PeConfig
+    { cpu = Riscv32imc.vexRiscv0
+    , depthI = SNat @(Div (4 * 1024) 4)
+    , depthD = SNat @(Div (16 * 1024) 4)
+    , initI = Nothing
+    , initD = Nothing
+    , iBusTimeout = d0
+    , dBusTimeout = d0
+    , includeIlaWb = False
+    }
+
+{- | See 'Bittide.Instances.Hitl.SwitchDemo.BringUp.bringUp'
+
+TODO: Deduplicate
+-}
+bringUp ::
+  "REFCLK" ::: Clock Basic125 ->
+  "TEST_RST" ::: Reset Basic125 ->
+  Circuit
+    ( Vec (InternalCpuCount + 1) (ToConstBwd Mm)
+    , Jtag Basic125
+    , Gth.Gths GthRx GthRxS Bittide GthTxS Ext200 LinkCount
+    )
+    ( Spi Basic125
+    , Sync Bittide Basic125
+    , "UART_TX" ::: CSignal Basic125 Bit
+    , "FINC_FDEC" ::: CSignal Bittide (FINC, FDEC)
+    )
+bringUp refClk refRst = withLittleEndian $ circuit $ \(memoryMaps, jtag, gths) -> do
+  ([bootMm], coreMemoryMaps) <- Vec.split -< memoryMaps
+
+  [bootJtag, otherJtag] <- jtagChain -< jtag
+  otherJtagBittide <- unsafeJtagSynchronizer refClk bittideClk -< otherJtag
+
+  transceiverWb <- withRefClockResetEnable arbiterMm -< [muTransceiverWb, bootTransceiverWb]
+  muTransceiverWb <-
+    fmapC (extendAddressWidthWb <| xpmCdcHandshakeWb bittideClk bittideRst refClk refRst)
+      -< muTransceiverWbBittide
+
+  -- Start boot PE
+  (bootUartBytes, _spiDone, spi, bootTransceiverWb) <-
+    withRefClockResetEnable
+      $ bootPe bootPeConfig
+      -< (bootMm, bootJtag)
+  -- Stop boot PE
+
+  -- Start UART multiplexing
+  uartTxBytes <-
+    withRefClockResetEnable
+      $ asciiDebugMux d1024 uartLabels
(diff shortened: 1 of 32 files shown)


bittide-instances/src/Bittide/Instances/Hitl/WireDemo/BringUp.hs
```diff
@@ -0,0 +1,188 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP #-}
+
+{- | Switch demo for a Bittide system. In concert with its driver file, this device under
+test should demonstrate the predictability of a Bittide system once it has achieved logical
+synchronicity.
+
+For more details, see [QBayLogic's presentation](https://docs.google.com/presentation/d/1AGbAJQ1zhTPtrekKnQcthd0TUPyQs-zowQpV1ux4k-Y)
+on the topic.
+-}
```
martijnbastiaan: This is outdated documentation

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/BringUp.hs
```diff
@@ -0,0 +1,188 @@
...
+  Fwd tOutputs <-
+    Transceiver.transceiverPrbsNWb @Bittide @GthRx @Ext200 @Basic125 @GthTxS @GthRxS
+      refClk
+      refRst
+      Transceiver.defConfig
+      -< (transceiverWb, gths, Fwd (bundle switchDataOutOrLocalCounters))
+
+  -- Send local counter the very first "user" cycle for UGN computation. Should
+  -- this be handled in 'core'?
+  Fwd switchDataOutOrLocalCounters <-
```
martijnbastiaan, line 132: Probably :). But I assume all the other demos do this here too?

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -0,0 +1,322 @@
...
+type StartDelay = 10 -- seconds
+
+{- | The delay in clock cycles between 2 PEs which is not accounted for by the
+`captureUgn` component.
+
+XXX: I have no idea why this would be anything other than 0. I cannot find the register
+which would be present on the path between 2 PEs, but not on the path between 2
+`captureUgn` components.
+-}
+type InternalDelay = 1
```
martijnbastiaan, line 65: As discussed offline, this happens because of:

```haskell
Fwd (dflipflop bitClk <$> (unbundle txs))
```

Which happens to be _before_ `sendUgnC`.

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Core.hs
```diff
@@ -0,0 +1,324 @@
...
+              <*> rxs0
+           )
+        <| repeatC (fmapC $ withBittideClockResetEnable delayWishbone)
+        -< ebWbs
+
+    -- Use of `dflipflop` to add pipelining should be replaced by
+    -- https://github.com/bittide/bittide-hardware/pull/1134
+    Fwd rxs2 <-
+      withBittideClockResetEnable
+        $ Vec.vecCircuits ((captureUgn localCounter . dflipflop bitClk) <$> rxs1)
```
martijnbastiaan, line 212: It does feel a little bit weird to me that `captureUgn` and `sendUgn` aren't instantiated at the same level. What do you think?

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -0,0 +1,322 @@
...
+  }
+  deriving (Show)
+
+{- | Generate a schedule for the wire demo, which consists of a configuration for the PE
+and the programmable mux for each node.
+
+First generates an 'absolute' schedule with only a few rules:
+  - 'firstBCycle' is always 1 cycle after the previous node's 'firstBCycle'
+  - 'readLink' is always the previous node (except for the first node)
+  - 'writeLink' is always the next node (except for the last node)
```
martijnbastiaan:
```suggestion
  - 'readLink' is always the previous node, except for the first node
  - 'writeLink' is always the next node, except for the last node
```

Pretty crucial info :-)!

docs/sections/demos/wire-demo.md
```diff
@@ -0,0 +1,121 @@
+<!--
+SPDX-FileCopyrightText: 2026 Google LLC
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Wire Demo
+Last year we added a [switch demo](switch-demo-asic.md) to showcase that we could use the bittide mechanism to thread a path through an 8-node network, all statically scheduled. To simplify things, we programmed a static calendar and used a calculator to produce the exact clock cycles at which the PE should send and receive data. For more details, see [this presentation](https://docs.google.com/presentation/d/1AGbAJQ1zhTPtrekKnQcthd0TUPyQs-zowQpV1ux4k-Y/edit?usp=sharing).
```
martijnbastiaan: I don't think qualifiers like "last year" do a lot in documentation.

martijnbastiaan (approved):
Nice work!

1. It is what it is. Could you open an issue on deduplicating the demos? It doesn't have to be fully fleshed out, I'll see if I can write up a plan in the comments.
2. See comment / as discussed.
3. Yes!
4. Local counter is good. Could you add that as a comment to the issue that that's what we decided to do?
5. Yes.

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/BringUp.hs
hiddemoll: I've removed the module documentation in each demo's `BringUp`, and added module documentation to each demo's `Core`. This makes more sense, since the `BringUp` is intentionally (almost) identical between demos to prepare for deduplication.

hiddemoll:
1. I've created an issue on deduplicating demos https://github.com/bittide/bittide-hardware/issues/1252
4. I've added a comment to original issue (https://github.com/bittide/bittide-hardware/issues/1109#issuecomment-4260055409)

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/BringUp.hs
hiddemoll: Yup, this behaviour is shared between all demos. I've opened issue https://github.com/bittide/bittide-hardware/issues/1253

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Core.hs
hiddemoll: I agree, I've opened issue https://github.com/bittide/bittide-hardware/issues/1253

PR #1251, opened 2026-04-16
nix: Update flake

rowanG077:
I updated the `flake.nix` to bring it into 2026. Consequently the cabal.freeze also had to be updated.

GHC 9.10.1 has a linker bug causing GHC to panic. This is fixed with GHC 9.10.3 and later. But since the newest GHC on the current nix env is 9.10.1 I had to update.


rowanG077:
All source code changes are simply that fourmolu decided haddock needs to be formatted differently.

rowanG077:
Depends on: https://github.com/QBayLogic/mdbook-drawio/pull/3/changes

bittide/src/Bittide/Ethernet/Mac.hs
```diff
@@ -339,9 +341,10 @@ ethMac1GFifoBb ::
   SNat txFifoDepth ->
   -- | RX FIFO depth
   SNat rxFifoDepth ->
+  -- | Logic clock
+
```
martijnbastiaan, line 345: This changes makes no sense 🙈. Broken `fourmolu`?

martijnbastiaan: Nvm, this is weirdly formatted anyway.

martijnbastiaan (approved):

martijnbastiaan:
Given that this PR will break all existing caches I've taken the opportunity to remove a Cabal hack too :).

PR #1259, opened 2026-04-27
Remove all `unsafePerformIO` usages from `bittide-hardware`

### LINKED ISSUE #1256: Remove `unsafePerformIO` usage
It's unsafe. And such.

More seriously though, I've had a couple times now where I was misled about whether a test would work because of `unsafePerformIO` shenanigans, and it would be nice not to have to worry about that in the future. So let's get rid of all of the usages of this that it's possible to, I think (:

rslawson:
_What (what did you do)_
Got rid of all uses of `unsafePerformIO` from `bittide-hardware`.

_Why (context, issues, etc.)_
Kept getting compile-time errors for what should be runtime dependencies and it really annoyed me.

The thing I'm most unsure of is honestly how I removed `unsafePerformIO` from `uartIO`. Previously it was used in a map over `[IO (Maybe Char)]` to produce a `[Maybe (BitVector 8)]`. I sort of read this as a `sequence`, but I'm not certain this is actually the correct way to do it. It does appear to work, but I'm not super happy about how I did it.

Nope.

# TODO


martijnbastiaan (approved):
Nice. Begone `unsafePerformIO`!

Edit: I only checked a few, I'm assuming you did the exact same thing for different files :-).

Issue #1254, opened 2026-04-17
Remove manual `memory.x` + peripheral addressing


rslawson:
I believe this is a comprehensive list of the tests that still have some manual stuff happening without use of the new memory map infrastructure. This is likely for legacy reasons, but it should still be updated.
- [ ] AXI Stream self test
- [x] Capture UGN test
- [x] Clock Control WB
- [x] DNA port E2 test
- [ ] Switch demo PE test
- [x] Time self test
- [ ] Watchdog test

Each of these tests should get a file in `bittide-instances/src/Bittide/Instances/Tests/` which exports its memory map, which should then be generated in `bittide-instances/src/Bittide/Instances/MemoryMaps.hs`. The tests should then have their `memory.x` file deleted and instead generated through their `build.rs` file, and the instantiation of peripherals with manually created addresses should also be removed in their binaries.

PR #1263, opened 2026-04-28
Use `deviceWb` and `registerWb` functions in SDPE, update firmware

rslawson:
_What (what did you do)_
`Wishbone.SwitchDemoProcessingElement.switchDemoPeWb` did not use the new `deviceWb` and `registerWb*` functions, instead relying on the old method of manually unpacking Wishbone reads/writes and building read/write vecs. These have been brought up to date with the new library functions available. Additionally, the switch demo processing element test was made to export its memory map, since previously it did not. The relevant firmware libraries and binaries have been updated to use the new exported memory map and the new generated code for the peripheral.

_Why (context, issues, etc.)_
Code cleanup as per #965 and #1254.

Hello!

No.

# TODO

- [x] Link to existing issue


bittide/src/Bittide/SwitchDemoProcessingElement.hs
```diff
@@ -171,18 +169,19 @@ switchDemoPeWb ::
...
-   where
-    readVec :: Vec (8 + bufferSize * 3 * 2 + 2) (Signal dom (BitVector 32))
-    readVec =
-      dflipflop
-        <$> ( unbundle (bitCoerce . map swapWords <$> writableRegs)
-                ++ unbundle (bitCoerce . map swapWords . bitCoerce <$> localCounter)
-                ++ unbundle (bitCoerce . map swapWords <$> buffer)
-            )
-
+-- switchDemoPeWb SNat localCounter maybeDna linkIn = withMemoryMap mm $ Circuit go
```
hiddemoll: Please remove commented code

bittide/src/Bittide/SwitchDemoProcessingElement.hs
```diff
@@ -171,18 +169,19 @@ switchDemoPeWb ::
...
+switchDemoPeWb SNat localCounter maybeDna linkIn = circuit $ \bbMm -> do
+  [wb0, wb1, wb2, wb3, wb4, wb5] <-
+    deviceWbI (deviceConfig "SwitchDemoPE"){registered = False} -< bbMm
```
hiddemoll: Is there a reason why this device should _not_ be registered?

bittide/src/Bittide/SwitchDemoProcessingElement.hs
```diff
@@ -171,18 +169,19 @@ switchDemoPeWb ::
...
+  Fwd (readStart, _) <- registerWbI readStartCfg (unpack 0) -< (wb0, Fwd (pure Nothing))
+  Fwd (readCycles, _) <- registerWbI readCyclesCfg (unpack 0) -< (wb1, Fwd (pure Nothing))
+  Fwd (writeStart, _) <- registerWbI writeStartCfg (unpack 0) -< (wb2, Fwd (pure Nothing))
+  Fwd (writeCycles, _) <- registerWbI writeCyclesCfg (unpack 0) -< (wb3, Fwd (pure Nothing))
+  registerWbI_ localClockCycleCounterCfg (unpack 0) -< (wb4, Fwd (Just <$> localCounter))
+  registerWbI_ bufferCfg (repeat 0 :: Vec (bufferSize * 3) (BitVector 64))
+    -< (wb5, Fwd (Just <$> buffer))
```
hiddemoll: You can remove all the `unpack`s here.
The type for the `repeat 0` can also be removed, as it is inferred by `buffer`

firmware-binaries/sim-tests/switch_demo_pe_test/src/main.rs
```diff
@@ -31,18 +25,15 @@ const BUFFER_SIZE: usize = 2;
...
+    let mut uart = DEVICES.uart;
+    let timer = DEVICES.timer;
+    let mut switch_pe_a = DEVICES.switch_demo_pe_0;
+    let mut switch_pe_b = DEVICES.switch_demo_pe_1;
 
     unsafe {
         let logger = &mut (*LOGGER.get());
         logger.set_logger(uart.clone());
-        let log_timer = Timer::new(TIMER_ADDR);
+        let log_timer = DEVICES.timer;
```
hiddemoll, line 36: Can this be written as such?
```suggestion
        let log_timer = timer;
```

firmware-support/bittide-sys/src/switch_demo_pe.rs
```diff
@@ -11,93 +11,75 @@ pub struct NodeData {
     pub dna: u128,
 }
 
-pub struct SwitchDemoProcessingElement<const BUFFER_SIZE: usize> {
-    base_addr: *const u64,
+pub trait SdpeUtils {
```
hiddemoll: Why do we now need a macro?

hiddemoll (changes requested):
This is a nice cleanup. I love `deviceWb`.

bittide/src/Bittide/SwitchDemoProcessingElement.hs
rslawson: From the `clash-protocols-memmap` docs:
```
Whether to insert a component breaking the combinatorial path
between the Wishbone bus and the registers. This improves timing
(place and route) characteristics, at the cost of consuming
hardware resources. Its default is set to `True`.

Note that both the forwards and the backwards path are registered.
This means that any transaction takes at least two clock cycles,
with memory backed transactions taking at least three. In the
future we might opt to make this a more flexible option.
```
I don't think that breaking the path is necessary here, plus looking at the old code the Wishbone bus lines don't appear to be buffered, so I didn't introduce it.

firmware-binaries/sim-tests/switch_demo_pe_test/src/main.rs
rslawson: Unfortunately no, since the `Timer` type isn't `Copy` (or `Clone`, so you can't `let log_timer = timer.clone()`).

firmware-support/bittide-sys/src/switch_demo_pe.rs
rslawson: I moved the `impl`'d functions/methods to a trait which needs to be implemented on multiple types, but those trait implementations are virtually identical so I macro'd it.

firmware-binaries/sim-tests/switch_demo_pe_test/src/main.rs
hydrolarus: I don't know if we'd even *want* for a timer to be copy/clone-able, but it would be easy to add a `impl Clone for Timer` in the `manual_additions`. Or some other custom method for this, because I'm not sure if access to the same timer from multiple references is going to work so well?

rslawson: Agreed, I also don't particularly like the idea of hardware resources being `Copy` or `Clone`.

bittide/src/Bittide/SwitchDemoProcessingElement.hs
martijnbastiaan: Yeah but this should be reasoned the other way around. The bus has, by its nature, an incredibly fanout. So the default is to limit that somewhat, to make PNR easy in general. If you truly need high performance, you can turn it off. I don't think high performance is needed in this case though.

hiddemoll (approved):

PR #1264, opened 2026-04-29
Add `Protocols.MemoryMap.Mask`

martijnbastiaan:
In Clash designs we often use `Vec n Bool` to track per-bit enables, valid bits, and similar. The default `BitPackC` representation of `Vec n Bool` allocates a full byte per bit, which is wasteful when these end up in memory-mapped registers. Switching to a `BitVector` is denser, but inverts the index correspondence: `Vec` index `0` maps to the MSB of the bitvector via Clash's standard `pack`/`unpack`, so a loop walking `0..n` on the Haskell side walks the wrong direction relative to firmware that pokes individual bits.

`Mask` fixes both problems: it is packed compactly (like `BitVector`), and its `toVec` / `fromVec` / `toBitVector` / `fromBitVector` conversions all preserve the `Vec`-index <-> bit-position correspondence: index `0` is bit `0` (the least significant bit).

Lastly, it introduces a mask-focused interface on the Rust side.

**Dear reviewer**
I'm currently extending the native types `clash-protocols-memmap` and friends know about. This is because in the future I'd like to have `Mask`s operate more efficiently for tasks common to them, e.g., testing a single bit. As far as I understand, the complete structure is currently fetched, after which the bit is tested. IMO that's out of scope for this PR though.

~~Note that the type is backed by `Unsigned`. I had originally implemented backed by `BitVector`, but this requires modulo operations in software.~~ @hydrolarus pointed out that this isn't true because you can translate `%8` to `& 0b111`. 

**AI disclaimer**
You bet!

# TODO
- [X] Write (regression) test
- [X] ~~Link to existing issue~~
- [x] Use `BitVector`-like backend in Rust
- [x] Add `Mask![n]` marcos

### DIFF
--- a/clash-protocols-memmap/tests/Tests/Protocols/MemoryMap/Mask.hs
+++ b/clash-protocols-memmap/tests/Tests/Protocols/MemoryMap/Mask.hs
@@ -1,0 +1,76 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE OverloadedStrings #-}
+
+module Tests.Protocols.MemoryMap.Mask where
+
+import Clash.Explicit.Prelude
+
+import Clash.Class.BitPackC (maybeUnpackC, packC)
+import Hedgehog (Property, forAll, property, (===))
+import Test.Tasty
+import Test.Tasty.Hedgehog (testPropertyNamed)
+
+import Protocols.MemoryMap.Mask (Mask, fromBitVector, fromVec, toBitVector, toVec)
+
+import qualified Clash.Hedgehog.Sized.Vector as HV
+import qualified Hedgehog.Gen as Gen
+
+prop_vecRoundTrip :: Property
+prop_vecRoundTrip = property $ do
+  bs <- forAll (HV.genVec @17 Gen.bool)
+  toVec (fromVec bs) === bs
+
+prop_bitVectorRoundTrip :: Property
+prop_bitVectorRoundTrip = property $ do
+  bv <- forAll (Gen.enumBounded @_ @(BitVector 17))
+  toBitVector (fromBitVector bv) === bv
+
+-- | Index 0 of the input 'Vec' becomes bit 0 (LSB) of the resulting 'BitVector'.
+prop_indexConvention :: Property
+prop_indexConvention = property $ do
+  let
+    v0 = True :> repeat False :: Vec 8 Bool
+    bv = toBitVector (fromVec v0)
+  bv === 0b0000_0001
+
+  let
+    v1 = repeat False :< True :: Vec 8 Bool
+    bv1 = toBitVector (fromVec v1)
+  bv1 === 0b1000_0000
+
+prop_packCRoundTrip :: Property
+prop_packCRoundTrip = property $ do
+  bv <- forAll (Gen.enumBounded @_ @(BitVector 17))
+  -- Note that byte ordering shouldn't matter as 'BitVector' isn't a numeric primitive,
+  -- but we test it anyway to snuff out issues.
+  byteOrder <- forAll Gen.enumBounded
+  let
+    m = fromBitVector bv :: Mask 17
+    packed = packC byteOrder m
+  case maybeUnpackC byteOrder packed of
+    Nothing -> fail "maybeUnpackC returned Nothing"
+    Just (m' :: Mask 17) -> toBitVector m' === bv
+
+tests :: TestTree
+tests =
+  testGroup
+    "Mask"
+    [ testPropertyNamed
+        "fromVec . toVec ≡ id (Vec 17 Bool)"
+        "prop_vecRoundTrip"
+        prop_vecRoundTrip
+    , testPropertyNamed
+        "fromBitVector . toBitVector ≡ id (BitVector 17)"
+        "prop_bitVectorRoundTrip"
+        prop_bitVectorRoundTrip
+    , testPropertyNamed
+        "Index 0 maps to LSB"
+        "prop_indexConvention"
+        prop_indexConvention
+    , testPropertyNamed
+        "BitPackC round-trip via packC/maybeUnpackC"
+        "prop_packCRoundTrip"
+        prop_packCRoundTrip
+    ]
--- a/firmware-support/bittide-hal/src/manual_additions/mask.rs
+++ b/firmware-support/bittide-hal/src/manual_additions/mask.rs
@@ -1,0 +1,290 @@
+// SPDX-FileCopyrightText: 2026 Google LLC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//! A `Mask<M, N>` type with bool-array semantics.
+//!
+//! `Mask<M, N>` is a transparent newtype around [`BitVector<M, N>`]: it has
+//! identical storage and identical bit ordering (bit `i` of the mask is bit
+//! `i` of the underlying byte array — LSB at byte 0, bit 0). The wrapper
+//! exists to provide a bool-indexed API (`get`, `set`, `from_iter`, `iter`)
+//! without re-implementing storage or size-checking. Conversion to/from
+//! `BitVector` is zero-cost.
+//!
+//! This is the Rust counterpart of the Haskell `Mask n` type from
+//! `Protocols.MemoryMap.Mask`. The Haskell side enforces the
+//! "index `i` of a `Vec n Bool` corresponds to bit `i`" convention, which is
+//! the same convention this type uses on the Rust side.
+#![deny(missing_docs)]
+
+use core::ops::Index;
+
+use crate::manual_additions::bitvector::{BitVector, BitVectorSizeCheck};
+
+/// Bit-mask of `M` bits, backed by a [`BitVector<M, N>`].
+///
+/// Bit `i` (for `i in 0..M`) is bit `i` of the underlying byte array (byte
+/// `i / 8`, bit `i % 8`). This matches the storage convention of
+/// [`BitVector<M, N>`], so converting between `Mask<M, N>` and
+/// `BitVector<M, N>` is zero-cost.
+#[derive(Clone, Copy, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct Mask<const M: usize, const N: usize>(pub(crate) BitVector<M, N>);
+
+impl<const M: usize, const N: usize> Mask<M, N>
+where
+    BitVector<M, N>: BitVectorSizeCheck<Inner = [u8; N]>,
+{
+    /// Create a `Mask` with all bits cleared.
+    #[inline]
+    pub fn zero() -> Self {
+        let _: () = <BitVector<M, N> as BitVectorSizeCheck>::SIZE_CHECK;
+        // SAFETY: the all-zero byte array represents the value `0`, which is
+        // in range for any `BitVector<M, N>`.
+        Mask(unsafe { BitVector::<M, N>::new_unchecked([0u8; N]) })
+    }
+
+    /// Read bit `i`. Returns `None` if `i >= M`.
+    #[inline]
+    pub fn get(&self, i: usize) -> Option<bool> {
+        if i >= M {
+            None
+        } else {
+            // SAFETY: just bounds-checked.
+            Some(unsafe { self.get_unchecked(i) })
+        }
+    }
+
+    /// Read bit `i` without bounds-checking.
+    ///
+    /// # Safety
+    /// `i` must be `< M`.
+    #[inline]
+    pub unsafe fn get_unchecked(&self, i: usize) -> bool {
+        let _: () = <BitVector<M, N> as BitVectorSizeCheck>::SIZE_CHECK;
+        // PERFORMANCE: The `/8` and `%8` compile to `i >> 3` and `i & 7` respectively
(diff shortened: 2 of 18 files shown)


firmware-support/bittide-hal/src/manual_additions/mask.rs
```diff
@@ -0,0 +1,272 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn from_iter_round_trip() {
+        let bs = [true, false, true, false, false, true, true, false];
+        let m: Mask<8, u8> = Mask::from_iter(bs);
+        let collected: [bool; 8] = core::array::from_fn(|i| m.get(i as u8).unwrap());
```
hydrolarus: Would it make sense to also add an `Index` instance that does an implicit unwrap?

hydrolarus (approved):
Overall looking good! Nice work. Is the idea to eventually have this type be in `clash-prelude`?

martijnbastiaan:
> Is the idea to eventually have this type be in `clash-prelude`?

Oh, I didn't even think of that! I'll open a Discussion.

firmware-support/bittide-hal/src/manual_additions/mask.rs
martijnbastiaan: We're talking about the trait `Index`, right? How would it prevent the `unwrap()`? You can still give an illegal index, right?

hydrolarus: Yes, the trait that lets you use the `val[index]` syntax.

Slices/Vecs have a `.get()` method that gives an `Option<Item>` but if you do `my_vec[i]` you get `Item` or `&Item` directly, it will panic if you give it an invalid index. This is a common enough convention that I think it would be worth it to include here?

martijnbastiaan: Okay, understood! I hadn't considered the `my_vec[i]` notation.

martijnbastiaan: Is this what you had in mind @hydrolarus?

firmware-support/bittide-hal/src/manual_additions/mask.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    #[test]
+    fn index_matches_get() {
+        let bs = [true, false, true, false, false, true, true, false];
+        let m: Mask<8, u8> = Mask::from_iter(bs);
+        for i in 0..8u8 {
+            assert_eq!(m[i], bs[i as usize]);
+        }
+    }
```
hydrolarus, line 260: the name of this test sounds like it should test that `m.get(i) == Some(m[i])` but that's not what it's doing.

firmware-support/bittide-hal/src/manual_additions/mask.rs
```diff
@@ -0,0 +1,306 @@
...
+    fn iter_matches_get() {
+        let bs = [true, false, true, false, false, true, true, false];
+        let m: Mask<8, u8> = Mask::from_iter(bs);
+        let mut collected = [false; 8];
+        for (i, b) in m.iter().enumerate() {
+            collected[i] = b;
+        }
+        assert_eq!(collected, bs);
+    }
+}
```
hydrolarus, line 290: Same here, the test is fine but the naming is confusing

hydrolarus (commented):
As discussed off-PR, there's still some changes to be made, but here's still a naming thing I noticed!

firmware-support/bittide-hal/src/manual_additions/mask.rs
```diff
@@ -0,0 +1,306 @@
...
+    #[inline]
+    pub fn into_unsigned(self) -> Unsigned<M, T> {
+        self.0
+    }
+
+    /// Construct a `Mask` from an [`Unsigned<M, T>`] without copying.
+    #[inline]
+    pub fn from_unsigned(u: Unsigned<M, T>) -> Self {
+        Mask(u)
+    }
```
rslawson: Since you have these, it would be nice to see the `From` impls below using them. Otherwise there's not really much point to having them.

firmware-support/bittide-hal/src/manual_additions/mask.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+impl<const M: u8, T> From<Mask<M, T>> for Unsigned<M, T> {
+    #[inline]
+    fn from(m: Mask<M, T>) -> Self {
+        m.0
+    }
+}
```
rslawson: I'd prefer to see these be a little more flexible on the size of the type of `other`. Maybe something like this?
```rs
impl<const M: u8, const N: u8, T, U> From<Unsigned<M, T>> for Mask<N, U>
where
    Unsigned<M, T>: UnsignedSizeCheck,
    Unsigned<N, U>: UnsignedSizeCheck,
    U: FromAs<T>,
{
    fn from(other: Unsigned<M, T>) -> Mask<N, U> {
        let _ = Unsigned::<M, T>::SIZE_CHECK;
        let _ = Unsigned::<N, U>::SIZE_CHECK;
        let _ = const {
            // ensure that M <= N here, else panic
        };
        if const { M == N } {
            Mask::<N, U>::from_unsigned(other)
        } else {
            Mask(Unsigned(other.0))
        }
    }
}
```
As well as `From<Mask<N, U>> for Unsigned<M, T>`.

rslawson (changes requested):
+1 to Lara's comments about test names, otherwise just some cleanup.

firmware-support/bittide-hal/src/manual_additions/mask.rs
martijnbastiaan:
```
        let _ = const {
            // ensure that M <= N here, else panic
        };
```

That's super interesting Rust lets you do that! I think it means that this will trigger a compile error, right?

martijnbastiaan: Could you check whether the current implementation is what you intended?

firmware-support/bittide-hal/src/manual_additions/mask.rs
martijnbastiaan: Like this?

martijnbastiaan: `index_matches_get` -> `index_round_trip`

`iter_matches_get` -> `iter_round_trip`

clash-protocols-memmap/tests/Tests/Protocols/MemoryMap/Mask.hs
```diff
@@ -0,0 +1,70 @@
...
+  bv1 === 0b1000_0000
+
+prop_packCRoundTrip :: Property
+prop_packCRoundTrip = property $ do
+  bv <- forAll (Gen.enumBounded @_ @(BitVector 17))
+  let m = fromBitVector bv :: Mask 17
+  let packed = packC LittleEndian m
+  case maybeUnpackC LittleEndian packed of
+    Nothing -> fail "maybeUnpackC returned Nothing"
+    Just (m' :: Mask 17) -> toBitVector m' === bv
```
hydrolarus, line 54: This could also test that the behaviour is the same no matter if it's LE or BE (since BitVectors are not numeric primitives)?

It's not super necessary but it could be nice to have this documented in the tests?

firmware-support/bittide-hal/src/manual_additions/mask.rs
```diff
@@ -0,0 +1,290 @@
...
+        (0..M).map(|i| unsafe { self.get_unchecked(i) })
+    }
+
+    /// Unwrap to the underlying [`BitVector<M, N>`].
+    #[inline]
+    pub fn into_bitvector(self) -> BitVector<M, N> {
+        self.0
+    }
+
+    /// Construct a `Mask` from a [`BitVector<M, N>`] without copying.
```
hydrolarus: I guess it depends on the definition of "copying" here, but it could be wrong! Values passed as arguments can be memcopied or involve copying in some other way, just because it's a moved value does not mean there's no copying.

I don't think the documentation needs to highlight this aspect in the first place, so maybe it can just be dropped?
```suggestion
    /// Construct a `Mask` from a [`BitVector<M, N>`].
```

firmware-support/bittide-hal/src/manual_additions/mask.rs
```diff
@@ -0,0 +1,290 @@
...
+    fn from_bitvector_zero_cost() {
+        // Round-trip Mask -> BitVector -> Mask preserves bits.
+        let bs = [true, false, true, true, false, true, false, false];
+        let m: Mask<8, 1> = Mask::from_iter(bs);
+        let bv: BitVector<8, 1> = m.into();
+        let m2: Mask<8, 1> = bv.into();
+        let collected: [bool; 8] = core::array::from_fn(|i| m2[i]);
+        assert_eq!(collected, bs);
+    }
+}
```
hydrolarus, line 290: This name also doesn't make sense. Usually "zero cost" means something in terms of guarantees of code generation for certain language features, but here it seems like it's doing the same check as `from_iter_round_trip()` above but goes through BitVector as well?

Maybe a better name would be `bitvector_round_trip`?

firmware-support/bittide-macros/src/lib.rs
```diff
@@ -449,6 +449,37 @@ pub fn BitVector(toks: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+/// ```rust,ignore
+/// # use bittide_macros::Mask;
+/// fn test(m: Mask![32]) {
+///     // ...
+/// #   _ = m;
+/// }
+/// ```
+#[proc_macro]
+#[allow(non_snake_case)]
+pub fn Mask(toks: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
hydrolarus, line 465: Nice! It would be cool if there was also a value-level macro `mask!(0b...., n = 32)`, but if you don't want to add that in this PR I think that would be fine too!

hydrolarus (changes requested):
Much better with the BitVector backing! Still some naming discrepancies and small suggesetions, but it's look good!

firmware-support/bittide-hal/src/manual_additions/mask.rs
martijnbastiaan: Yeah let's drop it. It is a little bit surprising to me that moving involves copying, but the Rust docs definitely say so.

clash-protocols-memmap/tests/Tests/Protocols/MemoryMap/Mask.hs
martijnbastiaan: Note + generator added :)

firmware-support/bittide-hal/src/manual_additions/mask.rs
martijnbastiaan: Done

firmware-support/bittide-hal/src/manual_additions/mask.rs
```diff
@@ -0,0 +1,290 @@
...
+    #[test]
+    fn from_iter_round_trip() {
+        let bs = [true, false, true, false, false, true, true, false];
+        let m: Mask<8, 1> = Mask::from_iter(bs);
+        let collected: [bool; 8] = core::array::from_fn(|i| m[i]);
+        assert_eq!(collected, bs);
+    }
+
+    #[test]
+    fn from_iter_round_trip_non_byte_aligned() {
```
hydrolarus: Sorry, one more tiiiny nitpick. This doesn't really test anything related to alignment (the address at which an object starts) but seems to check that a non-multiple-of-8 size works? The `get_and_set` below uses non-byte-multiple size as well.

If you want to keep that test, maybe name it `from_iter_round_trip_non_byte_multiple` or with some better name if you can come up with something better 😅

hydrolarus (approved):
Looking good to me! One more tiny naming nit-pick but it's not a blocker to get this merged IMO

firmware-support/bittide-macros/src/lib.rs
martijnbastiaan: I'll open another PR :)

firmware-support/bittide-hal/src/manual_additions/mask.rs
martijnbastiaan: Fixed!

hydrolarus (approved):
Nice!

rslawson (approved):

firmware-support/bittide-hal/src/manual_additions/mask.rs
rslawson: > ```
>         let _ = const {
>             // ensure that M <= N here, else panic
>         };
> ```
> 
> That's super interesting Rust lets you do that! I think it means that this will trigger a compile error, right?

Exactly, yep! It'll fail to evaluate the `const { ... }` block so it'll produce an error at comptime. You can just write something like
```rs
let _ = const {
    if M > N {
        panic!("Cannot infallibly convert if the `Unsigned` is larger than the `Mask`!");
    }
};
```
And then it'll fail to compile at call sites that are invalid.



> Could you check whether the current implementation is what you intended?

This is still missing from the current implementation.

martijnbastiaan: Sorry I messed up when reverting to the BV implementation, it's monomorphic again..

PR #1275, opened 2026-05-01
Export memmap from watchdog test, use memmap in binary

rslawson:
_What (what did you do)_
Made the watchdog test export its memory map, and used that memory map in the test binary (and build file).

_Why (context, issues, etc.)_
Code cleanup as per #1254.

I'm confused why it was necessary to use a `Bool` register type. I wanted it to be `()`, but doing that made the unit test fail.

None.

# TODO

- [x] Link to existing issue


bittide-instances/src/Bittide/Instances/Tests/Watchdog.hs
```diff
@@ -0,0 +1,107 @@
...
+    (uartTx, _uartStatus) <-
+      (uartInterfaceWb @_ @_ @4) d2 d2 uartBytes -< (uartBus, uartRx)
+    idC -< uartTx
+ where
+  idleMm devName =
+    MemoryMap
+      { tree = DeviceInstance locCaller devName
+      , deviceDefs = deviceSingleton $ idleMmDevDef devName
+      }
+  idleMmDevDef devName =
```
martijnbastiaan, line 64: Could you please add that these are fake registers? I.e., it is not backed by an actual register, it's just that everything in a (much lager) memory space gets ignored.

rslawson: Yep, can do.

bittide-instances/src/Bittide/Instances/Tests/Watchdog.hs
```diff
@@ -0,0 +1,109 @@
...
+-- Other
+import Bittide.Cpus.Riscv32imc (vexRiscv0)
+import Protocols
+import Protocols.Idle
+import Protocols.MemoryMap
+import VexRiscv (DumpVcd (NoDumpVcd))
+
+-- Qualified
+
+import GHC.Stack (HasCallStack)
```
hydrolarus: Hmmmm

martijnbastiaan (approved):

PR #1245, opened 2026-04-13
Change Rust debug profile for smaller build artifacts

rslawson:
_What (what did you do)_
Changed the debug profile for Rust binaries to decrease resource usage, and adjusted the ELF limits stage of CI to check all debug and release builds of binaries.

_Why (context, issues, etc.)_
This is the replacement PR for #1238.

_Dear reviewer (anything you'd like the reviewer to pay close attention to?)_
Not sure how to check it exactly, but I'm a little worried about putting the `fd`, `sd`, `rg`, and `tq` deps in just an install step in CI. Might be better to put them in the Nix flake?

_AI disclaimer (heads-up for more than inline autocomplete)_

None.

# TODO
_~~Cross-out~~ any that do not apply_

- [ ] ~~Write (regression) test~~
- [ ] ~~Update documentation, including `docs/`~~
- [x] Link to existing issue

PR #1279, opened 2026-05-06
Fix race condition in Cargo when building both debug and release

hiddemoll:
_What (what did you do)_
See PR title

_Why (context, issues, etc.)_
We've observed intermittent failing unittests for bittide-instances (last 3 days: [1](https://github.com/bittide/bittide-hardware/actions/runs/25397930944/job/74489878896), [2]()https://github.com/bittide/bittide-hardware/actions/runs/25339182288/job/74292323390, [3](https://github.com/bittide/bittide-hardware/actions/runs/25288467353/job/74136610665)). Note that 3 did not even fail, whereas 1 and 2 show a different but similar error. This seems to look like a race condition in Cargo when build both debug and release builds in parallel. 

We're not entirely sure why this did not show up in unscheduled CI runs (on pushes). Our guess on why it is happening _now_ is that debug builds now take longer than before PR https://github.com/bittide/bittide-hardware/pull/1245.

Nothing

Nothing

# TODO

- [ ] ~Write (regression) test~
- [ ] ~Update documentation, including `docs/`~
- [ ] Link to existing issue

### DIFF
--- a/bittide-instances/tests/unittests.hs
+++ b/bittide-instances/tests/unittests.hs
@@ -65,8 +65,11 @@ prepareTests = do
     dependentTestGroup
       "bittide-instances"
       AllSucceed
-      [ testGroup
+      [ -- These 2 cargo build commands are not actually dependent on eachother, but they
+        -- cannot happen in parallel as it could cause a race condition in Cargo.
+        dependentTestGroup
           "Build Rust crates"
+          AllSucceed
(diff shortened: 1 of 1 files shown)


martijnbastiaan (approved):
Yeah I also observed this locally, but wasn't motivated enough to chase this given that this happened only 1/20 (or so) tries..

PR #1284, opened 2026-05-07
Fix debug profile options for Rust builds

rslawson:
_What (what did you do)_
Removed the `strip` option from the debug profile for Rust builds. As noted [here](https://users.rust-lang.org/t/how-to-gdb-with-split-debug-files/102989/2), adding the strip option strips out the headers that the split debuginfo files use, which broke things that relied on debug headers.

_Why (context, issues, etc.)_
`VexRiscvTest` was failing.

Should I consider writing up something like what user p-kraszewski has in that thread? I don't know if it would be worth the effort, but it should certainly be doable at least.

None.

# TODO

- [x] Refer to existing issue (:

### DIFF
--- a/firmware-binaries/Cargo.toml
+++ b/firmware-binaries/Cargo.toml
@@ -8,8 +8,7 @@ incremental = false
 lto = true
 opt-level = "z"
 panic = "abort"
-split-debuginfo = "packed"
-strip = "symbols"
+split-debuginfo = "unpacked"
 
 [profile.release]
 codegen-units = 1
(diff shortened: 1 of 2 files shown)


hiddemoll (approved):
How does this influence the size of the binaries? 

Regarding this question: 
> Should I consider writing up something like what user p-kraszewski has in that thread? I don't know if it would be worth the effort, but it should certainly be doable at least.

I don't think this is necessary, at least not right now.

PR #1291, opened 2026-05-13
Use `PreferCircuit` for `UART` data register

martijnbastiaan:
_What (what did you do)_
The `data` register now feeds results from the FIFO to the CPU, instead of data from its own register. Ideally we would have a `pipeDf`, but this fix will do too.

_Why_
It's borked.

_Dear reviewer_
No.

_AI disclaimer_
Debugged by Claude, fixed by Claude, tested by Claude.

# TODO
- [x] Write (regression) test
- [X] Link to existing issue

### DIFF
--- a/bittide/tests/Tests/Wishbone.hs
+++ b/bittide/tests/Tests/Wishbone.hs
@@ -66,26 +66,28 @@ genStalls' = do
   numStalls <- smallInt
   genVec (PH.genStalls smallInt numStalls PH.Stall)
 
+maxInFlight :: SNat 16
+maxInFlight = SNat
+
 uartInterfaceWbTest :: Property
 uartInterfaceWbTest = property $ do
   stallsBefore <- forAll $ genStalls'
   stallsAfter <- forAll $ genStalls'
+
   let
     deviceName = "Uart"
     defs = (((getMMAny dutMm).deviceDefs) Map.! deviceName)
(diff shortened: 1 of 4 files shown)


hiddemoll (approved):

bittide/tests/Tests/Wishbone.hs
```diff
@@ -111,6 +115,129 @@ uartInterfaceWbTest = property $ do
...
+
+It relies on the bus stalling @uart.data()@ until the rxFifo has a byte and
+returning that byte. This test directly mimics that loop as a Wishbone master
+state machine and checks that bytes injected on the UART rx side come out on
+the UART tx side in the correct order.
+
+The pre-existing 'uartInterfaceWbTest' did not catch this because
+'wishbonePropWithModel' interposes a 'validatorCircuit' which adds a cycle of
+register delay between master and slave, masking the off-by-one. This test
+drives the bus directly without that delay.
```
martijnbastiaan: I'm very suspicious of this (Claude) reasoning

lmbollen: This is not true as the validatorCircuit does not insert latency

lmbollen: So it's still interesting to figure out why the existing test did not catch it.

PR #1308, opened 2026-05-27
Emit register descriptions for zero-width registers

### LINKED ISSUE #1182: Register descriptions don't make their way to `WriteOnly` registers
E.g., I added:

```haskell
    (_autoCenterReset, Fwd autoCenterResetActivity) <-
      registerWbI
        (registerConfig "auto_center_reset")
          { access = WriteOnly
          , description =
              "Clear total adjustments. You must disable the state machine and wait for it to be idle before resetting it."
          }
        ()
        -< (wbAutoCenterReset, Fwd (pure Nothing))
```

And the Rust I get is:

```rust
    pub fn set_auto_center_reset(&self, _val: ()) {
        unsafe { self.0.add(20usize).cast::<u8>().write_volatile(0) }
    }
```

I.e., no `doc`.

martijnbastiaan:
The Rust backend interpolated the `#desc` doc comment into the scalar and vector register accessor branches but not the zero-width branch, so `WriteOnly`/`ReadOnly` `()`-typed registers lost their descriptions in the generated code.

Fixes #1182

_Why (context, issues, etc.)_

The test is quite large, is there something more sensible?

Yes

# TODO

- [X] Write (regression) test
- [X] Link to existing issue
- [x] ~~Investigate whether the test can be written more compactly~~ It adds a little bit of test infra that can be reused for other tests. It is what it is?

### DIFF
--- a/firmware-support/memmap-generate/src/backends/rust/mod.rs
+++ b/firmware-support/memmap-generate/src/backends/rust/mod.rs
@@ -25,6 +25,8 @@ use crate::{
     storage::Handle,
 };
 
+pub mod testing_utils;
+
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum IdentType {
     Type,
@@ -576,6 +578,7 @@ fn generate_reg_get_method(
     } else if ctx.tags[reg.tags].iter().any(|tag| tag == "zero-width") {
         let ty = generate_type_ref(ctx, varis, None, refs, reg.type_ref);
         quote! {
+            #desc
             pub fn #name(&self) -> #ty {
                 let _ = unsafe {
                     self.0.add(#offset).cast::<u8>().read_volatile()
@@ -652,6 +655,7 @@ fn generate_reg_set_method(
     } else if ctx.tags[reg.tags].iter().any(|tag| tag == "zero-width") {
         let ty = generate_type_ref(ctx, varis, None, refs, reg.type_ref);
         quote! {
+            #desc
             pub fn #name(&self, _val: #ty) {
                 unsafe {
                     self.0.add(#offset).cast::<u8>().write_volatile(0)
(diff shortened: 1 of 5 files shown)


firmware-support/memmap-generate/src/backends/rust/mod.rs
```diff
@@ -479,6 +479,49 @@ pub fn generate_device_desc<'ir>(
...
+///
+/// This runs the full parse-to-codegen pipeline for one HAL: it builds the IR,
+/// monomorphises every register's type (nats only, as in the production build
+/// scripts), and generates the Rust code for each device. Unlike the build
+/// scripts in `bittide-hal`, it does *not* deduplicate types or devices across
+/// multiple HALs, which makes it convenient for tests that exercise a single
+/// memory map.
+///
+/// Returns the device name and generated code for each device.
+pub fn generate_device_descs(memmap: &MemoryMapDesc) -> Vec<(String, TokenStream)> {
```
hydrolarus: This seems nice for tests, but then maybe it should be in a separate module `backends::rust::testing_utils` maybe?

firmware-support/memmap-generate/tests/fixtures/zero_width_with_desc.json
```diff
@@ -0,0 +1,90 @@
...
+    "src_locations": [
+        {
+            "end_col": 1,
+            "end_line": 1,
+            "file": "test_fixture",
+            "module": "TestFixture",
+            "package": "memmap-generate-tests",
+            "start_col": 1,
+            "start_line": 1
+        }
```
hydrolarus, line 59: Could we find a way to mark this as "obviously a stub" without breaking expectations for what a source location looks like? Because there is no `memmap-generate-tests` Haskell package...

At the moment nothing really uses the source locations, so there's nothing to break yet, but I think it would be good if some basic intuitions like "a package is a Haskell package identifier" would hold. So maybe
```suggestion
        {
            "end_col": 1,
            "end_line": 1,
            "file": "test_fixture",
            "module": "TestFixture",
            "package": "stub-for-memmap-tests",
            "start_col": 1,
            "start_line": 1
        }
```

hydrolarus (approved):

PR #1318, opened 2026-06-03
Ugn grooming

lmbollen:
_What (what did you do)_
Implements UGN grooming for the wire demo: restoring a freshly-booted bittide system to a stored, known-good latency profile (λ^safe) so the application's fixed schedule lands identically every boot, regardless of boot-time counter offsets.

The work is built bottom-up:

* Generic weighted directed-graph library (Bittide.Graph.Weighted) with Bellman-Ford potentials / negative-cycle detection.
* UGN grooming algorithm (Bittide.ClockControl.Ugn.Grooming) on top of the graph — computes the per-node relabel q and per-link frame corrections that map an observed boot onto λ^safe.
* Two Wishbone peripherals: TimedReset (releases the application reset at a software-chosen local-counter cycle — the relabel point) and UgnCorrections (host→MU channel for per-link frame corrections).
* Host/MU integration: the wire-demo driver reads the hardware UGNs over GDB, grooms this boot onto the stored λ^safe (Bellman-Ford), writes the corrections + reset-release timing; the management-unit firmware applies each correction to its elastic buffer as a single atomic Wishbone write. An app_counter readback lets the host verify on hardware that localCounter − app_counter == tReset per node.
* Hedgehog property test for the relabel-restoration invariant, plus unit tests for the graph and TimedReset.

_Why (context, issues, etc.)_
After a boot, each node's UGNs carry arbitrary boot-time counter offsets and elastic buffers might settle at different occupancies, so a fixed application schedule can't be reused across boots. Grooming relabels every boot onto a single stored reference profile, absorbing the offsets via the timed-reset release and trimming the residual per-link drift (a few frames) inside the elastic buffer's safe range. This is the host-side groundwork for reproducible wire-demo scheduling.

Commit-by-commit review is intended — the history is structured so each commit introduces one coherent piece and compiles on its own (graph → algorithm → peripherals → MU wiring → HITL utils → driver → firmware → readback → buffer tuning).

λ^safe is stored in the raw (non-canonicalized) gauge in the Driver. 

I guided claude through the implementation of this PR based on the presentation offered by Sanjay regarding the method.

# TODO

- [x] Write (regression) test
- [ ] Link to existing issue

### DIFF
--- a/bittide-instances/src/Bittide/Instances/Hitl/GenericDemo/Core.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/GenericDemo/Core.hs
@@ -59,6 +59,7 @@ import Protocols
 import Bittide.CaptureUgn (captureUgns, sendUgn)
 import Bittide.ClockControl (SpeedChange)
 import Bittide.ClockControl.CallistoSw (SwcccInternalBusses, callistoSwClockControlC)
+import Bittide.ClockControl.Ugn.Corrections (correctionsWb)
 import Bittide.DoubleBufferedRam (wbStorage)
 import Bittide.ElasticBuffer (fromData, xilinxElasticBufferWb)
 import Bittide.Extra.Maybe (toMaybe)
@@ -76,6 +77,7 @@ import Bittide.ProcessingElement (
 import Bittide.RingBuffer (receiveRingBuffer, transmitRingBuffer)
 import Bittide.SharedTypes (Bitbone, BitboneMm)
 import Bittide.Sync (Sync)
+import Bittide.TimedReset (timedResetWb)
 import Bittide.Wishbone (readDnaPortE2WbWorker, timeWb, uartBytes, uartInterfaceWb)
 import Clash.Class.BitPackC (ByteOrder)
 import Clash.Cores.Xilinx (withXilinx)
@@ -105,8 +107,10 @@ type FifoSize = 5 -- = 2^5 = 32
     - `timeWb`
     - DNA
     - UART
+    - UGN grooming corrections
+    - application reset (TimedReset)
 -}
-type NmuInternalBusses = 3 + PeInternalBusses
+type NmuInternalBusses = 5 + PeInternalBusses
 
 {- Busses per link:
     - Elastic buffer
@@ -144,6 +148,10 @@ type UserCoreCircuit userCoreBusses muRemBusWidth =
     Enable Bittide ->
     Signal Bittide (Unsigned 64) ->
     Signal Bittide (Maybe (BitVector 96)) ->
+    -- Application reset, released by the management unit's 'timedResetWb' at a
+    -- chosen local-counter cycle (the UGN-grooming relabel). Demos that have an
+    -- application counter gate it on this reset; others may ignore it.
+    Reset Bittide ->
     Circuit
       ( Vec userCoreBusses (ToConstBwd Mm, Bitbone Bittide muRemBusWidth)
       , "RXS2_RAW" ::: CSignal Bittide (Vec LinkCount (BitVector 64))
@@ -201,6 +209,7 @@ managementUnit ::
   Circuit
     (ToConstBwd Mm.Mm, Jtag dom)
     ( Df dom (BitVector 8)
+    , "APP_RESET" ::: Reset dom
     , Vec
         (NmuExternalBusses userCoreBusses)
         ( ToConstBwd Mm.Mm
@@ -211,16 +220,24 @@ managementUnit externalCounter maybeDna =
   circuit $ \(mm, jtag) -> do
     -- Core and interconnect
     allBusses <- processingElement NoDumpVcd muConfig -< (mm, jtag)
-    ([timeBus, uartBus, dnaBus], restBusses) <- Vec.split -< allBusses
+    ([timeBus, uartBus, dnaBus, correctionsBus, timedResetBus], restBusses) <-
+      Vec.split -< allBusses
 
     -- Peripherals
     _localCounter <- timeWb (Just externalCounter) -< timeBus
     (uartOut, _uartStatus) <-
       uartInterfaceWb d16 d16 uartBytes -< (uartBus, Fwd (pure Nothing))
     readDnaPortE2WbWorker maybeDna -< dnaBus
+    -- Host-written UGN grooming corrections, polled and applied by the MU CPU.
+    correctionsWb (SNat @LinkCount) -< correctionsBus
+    -- Application reset for the demo's user core: the management unit chooses the
+    -- local-counter cycle at which the application leaves reset (the UGN-grooming
+    -- relabel). The reset lives here, in the layer around the user core; the
+    -- application counter it gates lives in the user core itself.
+    appReset <- timedResetWb externalCounter -< timedResetBus
 
     -- Output
-    idC -< (uartOut, restBusses)
+    idC -< (uartOut, appReset, restBusses)
 
 core ::
   forall userCoreBusses ringBufferDepth.
@@ -262,7 +279,7 @@ core bufferDepth mkUserCore (refClk, refRst) (bitClk, bitRst, bitEna) rxClocks r
       localCounter = register bitClk bitRst bitEna 0 (localCounter + 1)
 
     -- Start management unit
-    (muUartBytesBittide, muWbAll) <-
+    (muUartBytesBittide, Fwd appReset, muWbAll) <-
       withBittideClockResetEnable managementUnit localCounter maybeDna -< (muMm, muJtag)
     (ebWbs, muWbs1) <- Vec.split -< muWbAll
     (rxBufferBusses, muWbs2) <- Vec.split -< muWbs1
@@ -329,7 +346,7 @@ core bufferDepth mkUserCore (refClk, refRst) (bitClk, bitRst, bitEna) rxClocks r
 
     -- Start user core: post-handshake stage drives the GTH-TX wire.
     Fwd txsOut <-
-      mkUserCore bitClk bitRst bitEna localCounter maybeDna
+      mkUserCore bitClk bitRst bitEna localCounter maybeDna appReset
         -< ( extraMuBusses
            , Fwd (bundle rxs2Raw)
            , Fwd (bundle handshakesOut.toNeighbors)
--- a/bittide-instances/src/Bittide/Instances/Hitl/GenericDemo/Core.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/GenericDemo/Core.hs
@@ -99,7 +99,7 @@ import qualified Protocols.Vec as Vec
 -- | The number of CPUs in 'core'
 type InternalCpuCount = 2
 
-type FifoSize = 5 -- = 2^5 = 32
+type FifoSize = 6 -- = 2^6 = 64
 
 {- Internal busses:
     - Instruction memory
--- a/bittide-instances/src/Bittide/Instances/Hitl/Utils/Gdb.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/Utils/Gdb.hs
@@ -6,11 +6,20 @@ module Bittide.Instances.Hitl.Utils.Gdb where
 
 import Prelude
 
+import Clash.Prelude (Signed, Unsigned, Vec, checkedFromIntegral, toList)
+
 import Bittide.Hitl (DeviceInfo (..))
+import Bittide.Instances.Hitl.Setup (LinkCount)
 import Bittide.Instances.Hitl.Utils.Driver (getTargetIndex)
+import Bittide.Instances.Hitl.Utils.MemoryMap (getPathAddress)
 import Bittide.Instances.Hitl.Utils.OpenOcd (TapInfo (..))
+import Bittide.Wishbone (TimeCmd (Capture))
+import Control.Monad (forM_)
+import GHC.Stack (HasCallStack)
 import Gdb (Gdb)
 import Project.FilePath
+import Project.Handle (expectRight)
+import Protocols.MemoryMap (MemoryMap)
 import System.FilePath ((</>))
 import Vivado.Tcl (HwTarget)
 
@@ -26,3 +35,75 @@ initGdb hitlDir binName gdb tapInfo (hwT, _d) = do
   Gdb.setTimeout gdb Nothing
   Gdb.runCommand gdb "echo connected to target device"
   pure ()
+
+-- * GDB-based component HAL accessors
+
+--
+-- Read/write a component's registers over GDB, addressed through a 'MemoryMap'. Generic
+-- across demos (the memory map is a parameter). All require the CPU to be halted.
+
+-- | Read the current local-counter value from a device's @Timer@ peripheral.
+readCurrentTime :: (HasCallStack) => MemoryMap -> (HwTarget, DeviceInfo) -> Gdb -> IO (Unsigned 64)
+readCurrentTime mm (_, d) gdb = do
+  putStrLn $ "Getting current time from device " <> d.deviceId
+  commandAddress <- expectRight $ getPathAddress mm ["0", "Timer", "command"]
+  scratchAddress <- expectRight $ getPathAddress mm ["0", "Timer", "scratchpad"]
+  Gdb.writeLe gdb commandAddress Capture
+  Gdb.readLe gdb scratchAddress
+
+{- | Write a node's reset-release cycle to its @TimedReset@ peripheral. Once the local
+counter exceeds this value the node's application (and its application counter) leave
+reset.
+-}
+writeReleaseCycle :: (HasCallStack) => MemoryMap -> Gdb -> Unsigned 64 -> IO ()
+writeReleaseCycle mm gdb releaseCycle = do
+  addr <- expectRight $ getPathAddress mm ["0", "TimedReset", "release_cycle"]
+  Gdb.writeLe gdb addr releaseCycle
+
+{- | Write the per-link frame corrections and set the @valid@ flag on the
+@UgnCorrections@ peripheral. Resuming the CPU makes the management unit apply the
+corrections to its elastic buffers.
+-}
+writeCorrections :: (HasCallStack) => MemoryMap -> Gdb -> Vec LinkCount (Signed 64) -> IO ()
+writeCorrections mm gdb corrections = do
+  correctionsAddr <- expectRight $ getPathAddress mm ["0", "UgnCorrections", "corrections"]
+  validAddr <- expectRight $ getPathAddress mm ["0", "UgnCorrections", "valid"]
+  Gdb.writeLe gdb correctionsAddr corrections
+  Gdb.writeLe gdb validAddr True
+
+{- | Read the per-link hardware UGN counter captures from the @CaptureUgns@ peripheral:
+the @(localCounter, remoteCounter)@ pair per link, with @localCounter@ already adjusted
+for the elastic-buffer frames added/removed since capture (@elastic_buffer_delta@).
+-}
+readHardwareUgns ::
+  (HasCallStack) => MemoryMap -> (HwTarget, DeviceInfo) -> Gdb -> IO [(Unsigned 64, Unsigned 64)]
+readHardwareUgns mm (_, d) gdb = do
+  let
+    getUgnRegister reg = expectRight $ getPathAddress mm ["0", "CaptureUgns", reg]
+
+    -- Adjust the local counter for the frames added/removed from the elastic
+    -- buffer after capturing the UGN. Leaves the remote counter untouched.
+    adjustLocalCounter :: (Unsigned 64, Unsigned 64, Signed 32) -> (Unsigned 64, Unsigned 64)
+    adjustLocalCounter (localCounter, remoteCounter, delta) =
+      (addSigned localCounter delta, remoteCounter)
+     where
+      addSigned :: Unsigned 64 -> Signed 32 -> Unsigned 64
+      addSigned u s = checkedFromIntegral (toInteger u + toInteger s)
+
+  putStrLn $ "Getting UGNs for device " <> d.deviceId
+  localCounters <- Gdb.readLe @(Vec LinkCount (Unsigned 64)) gdb =<< getUgnRegister "local_counter"
+  remoteCounters <- Gdb.readLe @(Vec LinkCount (Unsigned 64)) gdb =<< getUgnRegister "remote_counter"
+  deltas <- Gdb.readLe @(Vec LinkCount (Signed 32)) gdb =<< getUgnRegister "elastic_buffer_delta"
+  let ugnTriples = zip3 (toList localCounters) (toList remoteCounters) (toList deltas)
+  forM_ ugnTriples $ \triple -> putStrLn $ "Raw UGN triple: " <> show triple
+  pure $ adjustLocalCounter <$> ugnTriples
+
+{- | Read the relabeled application counter exposed by the user core's
+@AppCounterReadback@ peripheral. With the application running, @localCounter -
+app_counter@ equals the node's reset-release cycle, letting the host verify the relabel
+landed on hardware.
+-}
+readAppCounter :: (HasCallStack) => MemoryMap -> (HwTarget, DeviceInfo) -> Gdb -> IO (Unsigned 64)
+readAppCounter mm (_, _) gdb = do
+  addr <- expectRight $ getPathAddress mm ["0", "AppCounterReadback", "app_counter"]
+  Gdb.readLe gdb addr
--- a/bittide-instances/src/Bittide/Instances/Hitl/Utils/Ugn.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/Utils/Ugn.hs
@@ -211,6 +211,31 @@ timingOracleToUgnEdge oracle =
     , ugn = bitCoerce oracle.dstCounter - bitCoerce oracle.srcCounter
     }
 
+{- | Turn a per-node, per-link table of UGN values (@λ@) into 'UgnEdge's keyed by DNA
+node id. The outer vector is FPGA-indexed (the capturing / destination node); each
+inner vector is link-indexed (matching 'knownLinkConfigs').
+-}
+ugnEdges :: CP.Vec FpgaCount (CP.Vec LinkCount (Signed 64)) -> [UgnEdge]
+ugnEdges lams =
+  L.concat (CP.toList (CP.imap nodeEdges lams))
+ where
+  nodeEdges dstIndex links = CP.toList (CP.imap (edge dstIndex) links)
+  edge dstIndex port lam =
+    UgnEdge
+      { srcNode = indexToNodeId ((knownLinkConfigs CP.!! dstIndex) CP.!! port)
+      , srcPort = port
+      , dstNode = indexToNodeId dstIndex
+      , dstPort = port
+      , ugn = lam
+      }
+
+{- | As 'ugnEdges', but from raw @(localCounter, remoteCounter)@ captures: the UGN is
+their signed difference @local - remote@.
+-}
+hardwareUgnEdges ::
+  CP.Vec FpgaCount (CP.Vec LinkCount (Unsigned 64, Unsigned 64)) -> [UgnEdge]
+hardwareUgnEdges = ugnEdges . CP.map (CP.map (\(l, r) -> bitCoerce l - bitCoerce r))
+
 {- | Compare hardware and software UGN edges with detailed difference reporting.
 
 Performs edge-by-edge comparison and reports:
--- a/bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
@@ -1,0 +1,140 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+{- |
+The application-facing UGN grooming layer for HITL: it adapts the generic grooming
+algorithm ("Bittide.ClockControl.Ugn.Grooming", on "Bittide.Graph.Weighted") to the HITL
+'UgnEdge' type and turns its result into a directly-applyable 'RelabelPlan'.
+
+Nodes are identified by their DNA-derived 'BitVector' 32 id (the @srcNode@ / @dstNode@
+fields of a 'UgnEdge'), so the graph is keyed directly on those ids. Edges are matched
+between the measured @λ@ and stored @λ^safe@ snapshots by their @(srcNode, dstNode)@ pair
+— unique because the demo topology is a complete graph (one link per ordered node pair).
+
+@Signed 64@ UGN values are widened to 'Integer' for the (overflow-prone) relaxation, and
+the resulting reset offsets / frame counts are narrowed back with a bounds check.
+-}
+module Bittide.Instances.Hitl.Utils.UgnGrooming (
+  ugnGraph,
+  safeMargin,
+  canonicalizeUgn,
+  RelabelPlan (..),
+  computeRelabel,
+) where
+
+import Prelude
+
+import Clash.Prelude (BitVector, Signed, Vec, checkedFromIntegral)
+
+import Bittide.ClockControl.Ugn.Grooming (GroomResult (..), groomCorrection)
+import Bittide.Graph.Weighted (Graph)
+import Bittide.Instances.Hitl.Setup (FpgaCount, LinkCount, knownLinkConfigs)
+import Bittide.Instances.Hitl.Utils.Ugn (UgnEdge (..), indexToNodeId)
+
+import qualified Bittide.Graph.Weighted as G
+import qualified Clash.Prelude as CP
+import qualified Data.Map.Strict as Map
+
+{- | Build a weighted graph from UGN edges, keyed on the DNA-derived node ids with
+the (signed) UGN as the 'Integer' edge weight.
+-}
+ugnGraph :: [UgnEdge] -> Graph (BitVector 32) Integer
+ugnGraph es = G.fromEdges [(e.srcNode, e.dstNode, toInteger e.ugn) | e <- es]
+
+{- | Add a safety margin @ε@ to every edge: @λ^safe = λ + ε@. Mirrors the
+record-update style of 'Bittide.Instances.Hitl.Utils.Ugn.addLatencyEdge'.
+-}
+safeMargin :: Signed 64 -> [UgnEdge] -> [UgnEdge]
+safeMargin eps = map (\e -> e{ugn = e.ugn + eps})
+
+{- | Relabel UGN edges to their minimal non-negative form: the Bellman-Ford reduced
+costs @λ_{i->j} + q_i - q_j@ (potentials @q@ from the super-source), which are all
+@>= 0@ and leave every round-trip / cycle sum unchanged.
+
+This is a pure gauge change (a relabeling), so it represents the /same/ physical
+system, but in a gauge where the UGNs are small and non-negative. Storing
+@λ^safe@ in this gauge keeps the application's per-link UGNs (and hence its fixed
+schedule) inside the working range a depth-bounded mux/elastic-buffer needs, while
+still honouring the true /asymmetric/ cycle constraints (unlike a symmetric
+midpoint, which is only a valid relabeling on an acyclic graph).
+
+If the edges are not physically allowed (a negative cycle), they are returned
+unchanged.
+-}
+canonicalizeUgn :: [UgnEdge] -> [UgnEdge]
+canonicalizeUgn es =
+  case G.potentials (ugnGraph es) of
+    Nothing -> es
+    Just q ->
+      [ e{ugn = checkedFromIntegral (toInteger e.ugn + qOf e.srcNode - qOf e.dstNode)}
+      | e <- es
+      ]
+     where
+      qOf n = Map.findWithDefault 0 n q
+
+{- | The result of grooming a measured boot onto a target: per-node reset offsets and
+per-node, per-link frame corrections, both FPGA-/link-indexed and ready to apply.
+-}
+data RelabelPlan = RelabelPlan
+  { resetOffsets :: Vec FpgaCount (Signed 64)
+  {- ^ Per-node offset to add to the shared base to get its reset-release cycle. This
+  realises the relabel @q@: every node's application counter leaves reset in the
+  relabeled domain. Gauged so node 0's offset is 0.
+  -}
+  , corrections :: Vec FpgaCount (Vec LinkCount (Signed 64))
+  {- ^ Per-node, per-link frame corrections: the frames the management unit inserts
+  (@> 0@) or removes (@< 0@) on each receiving link to reach @λ^safe@ after relabeling.
+  -}
+  }
+  deriving (Show)
+
+{- | Groom this boot's measured UGNs onto the target @λ^safe@ and produce an applyable
+'RelabelPlan': per-node reset offsets (the relabel @q@) and per-link frame corrections.
+Returns 'Left' (with the witnessing node ids) if the UGNs changed too much to fit under
+@λ^safe@ (a negative cycle in the slack graph).
+
+Calls 'groomCorrection' (Bellman-Ford, per the UGN grooming slides) directly on the
+measured / target 'ugnGraph's. The relabel @q@ is the per-node potential that fits this
+boot's @λ@ under the stored @λ^safe@ (@λ + B^T q <= λ^safe@); the residual
+@frames = λ^safe - (λ + B^T q)@ are the small, non-negative per-link corrections the
+elastic buffers insert. So:
+
+  * @resetOffsets_i = -(q_i - q_0)@ — the reset release that applies the relabel
+    (gauged so node 0 is 0); and
+  * @corrections@ — the per-receiving-link frames. Since 'groomCorrection' returns frames
+    keyed by @(srcNode, dstNode)@, the receiving @(fpga, port)@ is reconstructed via
+    'knownLinkConfigs' (the same link map 'Bittide.Instances.Hitl.Utils.Ugn.ugnEdges'
+    uses).
+
+When @λ^safe@ is stored in a small non-negative gauge (see 'canonicalizeUgn'), the relabel
+absorbs this boot's counter offsets so the groomed app-frame UGNs land back on @λ^safe@ —
+small, non-negative, and identical every run, which is what makes a single fixed schedule
+reusable.
+
+Edges are matched by @(srcNode, dstNode)@; pass the full measured + target graphs to
+groom every link.
+-}
+computeRelabel ::
+  -- | Measured UGNs @λ@
+  [UgnEdge] ->
+  -- | Target UGNs @λ^safe@
+  [UgnEdge] ->
+  Either [BitVector 32] RelabelPlan
+computeRelabel measured target =
+  case groomCorrection (ugnGraph measured) (ugnGraph target) of
+    Infeasible ns -> Left ns
+    Feasible{correction = q, framesToInsert} ->
+      Right
+        RelabelPlan
+          { resetOffsets = (\idx -> negate (qOf idx - qOf 0)) <$> CP.indicesI
+          , corrections = (\idx -> (\port -> frameOf idx port) <$> CP.indicesI) <$> CP.indicesI
+          }
+     where
+      qOf idx = checkedFromIntegral (Map.findWithDefault 0 (indexToNodeId idx) q)
+      frameByEdge = Map.fromList [((i, j), f) | (i, j, f) <- framesToInsert]
+      frameOf dstIdx port =
+        checkedFromIntegral (Map.findWithDefault 0 (srcNode, dstNode) frameByEdge)
+       where
+        srcNode = indexToNodeId ((knownLinkConfigs CP.!! dstIdx) CP.!! port)
+        dstNode = indexToNodeId dstIdx
--- a/bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
+++ b/bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
@@ -17,6 +17,7 @@ import Bittide.Instances.Hitl.Setup (FpgaCount, demoRigInfo, fpgaSetup)
 import Bittide.Instances.Hitl.Utils.Driver
 import Bittide.Instances.Hitl.Utils.Gdb (
   initGdb,
+  readAppCounter,
   readCurrentTime,
   readHardwareUgns,
   writeCorrections,
@@ -525,6 +526,21 @@ driver testName targets = do
             newCurrentTime <-
               readCurrentTime MemoryMaps.managementUnit (L.head targets) (L.head managementUnitGdbs)
             putStrLn [i|Clock is now: #{newCurrentTime}|]
+
+            -- Relabel-landing check: with the app running, read each node's app_counter
+            -- and localCounter. If the timed-reset relabel landed, localCounter -
+            -- app_counter == tReset for that node. A divergence pinpoints a node whose
+            -- reset did not release at the intended (relabeled) cycle.
+            putStrLn "\n=== Relabel landing (localCounter - app_counter should == tReset) ==="
+            forM_ (L.zip3 targets managementUnitGdbs (toList tResets)) $ \(tgt@(_, dev), gdb, tReset) -> do
+              appCtr <- readAppCounter MemoryMaps.managementUnit tgt gdb
+              locCtr <- readCurrentTime MemoryMaps.managementUnit tgt gdb
+              let
+                observed = toInteger locCtr - toInteger appCtr
+                did = dev.deviceId
+                diff = observed - toInteger tReset
+              putStrLn
+                [i|  #{did}: app_counter=#{appCtr} loc-app=#{observed} tReset=#{tReset} diff=#{diff}|]
 
           let
             dnas = L.map (.dna) demoRigInfo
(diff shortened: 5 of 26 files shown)


bittide/src/Bittide/TimedReset.hs
```diff
@@ -0,0 +1,74 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+{- |
+A small Wishbone peripheral that releases a reset at a software-chosen cycle.
```
martijnbastiaan, line 6: This is a duplicate of `ProgrammableMux`. (Not **exactly**, but you know..)

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+{- |
+UGN grooming, expressed on top of "Bittide.Graph.Weighted".
+
+UGNs (@\lambda_{i->j}@) are signed per-link clock/frame offsets, i.e. edge weights
```
martijnbastiaan, line 8: The LaTeX style documentation doesn't do it for me: it neither looks good in the source code, nor does it render to anything with Haddock. I'd prefer it if it either:

1. Renders properly in Haddock
2. Looks good in the source through use of unicode (unicode in documentation is a-ok).

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
...
+on a directed graph of nodes. They cannot be chosen independently: the sum around
+every directed cycle must be nonnegative (the physical round-trip constraint).
+
+This module provides 'groomCorrection': given a freshly measured @\lambda@ and a
+previously stored @\lambda^{safe} = \lambda^{obs} + \varepsilon@, it finds a relabeling
+@q@ and the frames to insert so the system is restored to @\lambda^{safe}@, or reports
+that the UGNs changed too much. (Whether a set of UGNs is physically realisable at all —
+no negative cycle — is 'Bittide.Graph.Weighted.negativeCycle'.)
+
+The math (see the UGN grooming slides): with slack
```
martijnbastiaan, line 18: What slides? IMO we should just drop the PDF in the repo, _but_ we should ask @sanjaylall for permission.

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
```
martijnbastiaan, line 3: The commit message's body is weird: it doesn't assume the reader is familiar with UGNs, but it _does_ just drop jargon like "difference-constraint system is feasible". Who is your target audience?

The last paragraph can be removed, it doesn't belong in the commit message body.

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
...
+
+UGNs (@\lambda_{i->j}@) are signed per-link clock/frame offsets, i.e. edge weights
+on a directed graph of nodes. They cannot be chosen independently: the sum around
+every directed cycle must be nonnegative (the physical round-trip constraint).
+
+This module provides 'groomCorrection': given a freshly measured @\lambda@ and a
+previously stored @\lambda^{safe} = \lambda^{obs} + \varepsilon@, it finds a relabeling
+@q@ and the frames to insert so the system is restored to @\lambda^{safe}@, or reports
+that the UGNs changed too much. (Whether a set of UGNs is physically realisable at all —
+no negative cycle — is 'Bittide.Graph.Weighted.negativeCycle'.)
```
martijnbastiaan, line 16: This is _way_ too short/jargony to be useful to the average digital designer. I think you should pull in more context on what UGN grooming is and why we're doing all this. Preferably write it yourself, not Claude. In my mind a good outline would be:

1. What is UGN / UGN pair
2. What differs between boots
3. Why is that bad
4. Why can't you _just_ relabel
5. Explain that we found a trick (+\varepsilon) which happens to get a set of equations we can easily solve using well-known algorithms.
6. Mention that this is just _a_ solution, but perhaps we'll find a "better" one in the future (e.g., one that can also remove frames).

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
...
+import Data.Map.Strict (Map)
+
+import Bittide.Graph.Weighted
+
+import qualified Data.Map.Strict as Map
+
+-- | Outcome of 'groomCorrection'.
+data GroomResult n w
+  = {- | The slack graph has a negative cycle (UGNs changed too much); the
+    witnessed cycle is included for diagnostics.
```
martijnbastiaan, line 41: Or it was a broken graph to begin with, right? I.e., it contained negative cycles.

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+  isAllowed (bidirTriangle 1 1 1 (-3) 1 1) @?= False
+
+-- | Complete digraph over @0 .. k-1@ with the given per-edge weights.
+completeFrom :: Int -> [Integer] -> Graph Int Integer
+completeFrom k ws = fromEdges (zipWith (\(i, j) w -> (i, j, w)) (pairs k) ws)
+
+pairs :: Int -> [(Int, Int)]
+pairs k = [(i, j) | i <- [0 .. k - 1], j <- [0 .. k - 1], i /= j]
+
+prop_groom_reaches_safe :: Property
```
martijnbastiaan, line 66: This test is narrower than what we would expect the algorithm to groom in practice. I think what it should do is:

1. Generate a graph (any shape where each link is bidirectional will do)
2. Pick a safety margin and calculate your "safe graph".
3. Generate a random "error" for each edge in `[0, margin]`
4. Generate a random relabel "error" for each node and apply that to all its edges.
5. Run grooming, check that you get the original errors out.

What should be documented:

1. "Any shape where each link is bidirectional" could be loosened further, but difficult to generate that each link is part of a cycle and all nodes are reachable from every other node.
2. The error for each edge COULD also be negative, but it's difficult to generate while making sure no negative cycles are introduced.
3. Safety margin could be picked per link.

bittide/src/Bittide/ClockControl/Ugn/Corrections.hs
```diff
@@ -0,0 +1,72 @@
...
+{- | UGN grooming corrections peripheral. Holds @n@ signed corrections and a
+validity flag, all writable and readable over Wishbone. See the module header
+for the intended host/CPU handshake.
+-}
+correctionsWb ::
+  forall n dom addrW.
+  ( HasCallStack
+  , C.HiddenClockResetEnable dom
+  , KnownNat addrW
+  , KnownNat n
```
martijnbastiaan, line 60: Make this an `SNat` argument please, drop `AllowAmbiguousTypes`.

bittide/src/Bittide/ClockControl/Ugn/Corrections.hs
```diff
@@ -0,0 +1,72 @@
...
+
+  1. The CPU finishes its normal work and spins, polling @valid@ over MMIO.
+  2. The host computes the corrections, halts the CPU via GDB, writes
+     @corrections@, sets @valid@ to 'True', and resumes the CPU.
+  3. On its next poll the CPU sees @valid == True@, reads @corrections@, and
+     applies them.
+
+Both registers are plain bus-accessible storage (no fabric write port): the CPU
+and the host both reach them as Wishbone masters. Nothing reads them from the
+fabric, so the peripheral has no outputs.
```
martijnbastiaan, line 31: Yeah, but this is a bit superfluous. The host accesses the component _through_ the CPU, but this has nothing to do with this component.

bittide/src/Bittide/ClockControl/Ugn/Corrections.hs
```diff
@@ -0,0 +1,72 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE AllowAmbiguousTypes #-}
+
+-- The number of corrections @n@ is fixed by the caller via a type application
+-- (e.g. @correctionsWb \@LinkCount@); it no longer appears in the circuit's
+-- input/output, hence 'AllowAmbiguousTypes'.
+
+{- |
+A small Wishbone scratch peripheral holding the UGN grooming corrections that the
+host computes and the management-unit CPU applies.
```
martijnbastiaan, line 12: I'd say the module documentation is just something like "tooling to communicate corrections from a host to a management-unit CPU". The rest of the documentation is really documentation for `correctionsWb`.

bittide/src/Bittide/ClockControl/Ugn/Corrections.hs
```diff
@@ -0,0 +1,72 @@
...
+  forall n dom addrW.
+  ( HasCallStack
+  , C.HiddenClockResetEnable dom
+  , KnownNat addrW
+  , KnownNat n
+  , ?byteOrder :: ByteOrder
+  ) =>
+  Circuit (BitboneMm dom addrW) ()
+correctionsWb = circuit $ \bus -> do
+  [wbCorrections, wbValid] <- deviceWbI (deviceConfig "UgnCorrections") -< bus
```
martijnbastiaan, line 65:
```suggestion
  [wbCorrections, wbValid] <- deviceWbI (deviceConfig "ugn_corrections") -< bus
```

(Though I think the Rust layer already applies this convention.)

bittide/src/Bittide/Graph/Weighted.hs
```diff
@@ -0,0 +1,235 @@
+-- SPDX-FileCopyrightText: 2026 Google LLC
+--
+-- SPDX-License-Identifier: Apache-2.0
+
```
martijnbastiaan, line 4: Is it really impossible to use a package from Hackage? I see multiple packages providing an implementation for Bellman-Ford. I'm asking because you're asking me to review it, but that seems like a bunch of busy work if there are pre-tested libraries out there.

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
...
+
+import Prelude
+
+import Data.Map.Strict (Map)
+
+import Bittide.Graph.Weighted
+
+import qualified Data.Map.Strict as Map
+
+-- | Outcome of 'groomCorrection'.
```
martijnbastiaan, line 38:
```suggestion
```

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
...
+data GroomResult n w
+  = {- | The slack graph has a negative cycle (UGNs changed too much); the
```
martijnbastiaan, line 40: Why is it called a slack graph? Its definition doesn't seem to have to do with "slack".

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
...
+    witnessed cycle is included for diagnostics.
+    -}
+    Infeasible [n]
+  | {- | A feasible restore: relabel by 'correction', then insert
+    'framesToInsert' frames per edge to reach @\lambda^{safe}@.
+    -}
+    Feasible
+      { correction :: Map n w
```
martijnbastiaan, line 48: `correction` sounds like "frames to insert". Maybe "relabelDeltas"?

bittide/src/Bittide/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,85 @@
...
+{- | Find the correction restoring measured UGNs @\lambda@ to the stored safe UGNs
+@\lambda^{safe}@.
+
+The slack graph has edge weights @\lambda^{safe}_{i->j} - \lambda_{i->j}@; its
+Bellman-Ford potentials give the relabeling @q@. The frames to insert on each edge
+are @\lambda^{safe}_{i->j} - (\lambda_{i->j} + q_j - q_i)@, which are @>= 0@ by the
+shortest-path relaxation invariant.
+
+Edges are matched between the two graphs by 'zipEdgesWith'; only edges present in
+both are considered.
```
martijnbastiaan, line 62:
```suggestion
```

These are implementation details, not relevant for users of this function.

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+import Test.Tasty.Hedgehog (testProperty)
+import Test.Tasty.TH (testGroupGenerator)
+
+import qualified Data.Map.Strict as Map
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+-- | Are these UGNs physically realisable? (No negative cycle on weights = @λ@.)
+isAllowed :: (Ord n, Ord w, Num w) => Graph n w -> Bool
+isAllowed = isNothing . negativeCycle
```
martijnbastiaan, line 25: Either

```
noNegativeCycles :: (Ord n, Ord w, Num w) => Graph n w -> Bool
noNegativeCycles = isNothing . negativeCycle
```

or

```
hasNegativeCycles :: (Ord n, Ord w, Num w) => Graph n w -> Bool
hasNegativeCycles = isJust . negativeCycle
```

and invert at call site.

Document its context at use site.

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+import Test.Tasty
+import Test.Tasty.HUnit (Assertion, testCase, (@?=))
+import Test.Tasty.Hedgehog (testProperty)
+import Test.Tasty.TH (testGroupGenerator)
+
+import qualified Data.Map.Strict as Map
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+-- | Are these UGNs physically realisable? (No negative cycle on weights = @λ@.)
```
martijnbastiaan, line 23: What does

> No negative cycle on weights = ...

mean? Where does the lambda come from? How is it relevant to this function?

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+isAllowed :: (Ord n, Ord w, Num w) => Graph n w -> Bool
+isAllowed = isNothing . negativeCycle
+
+-- | A directed 3-cycle (slide-5 style): allowed iff the cycle sum is nonnegative.
```
martijnbastiaan, line 27: References slide the reader can't access.

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+triangle :: Integer -> Integer -> Integer -> Graph Int Integer
+triangle a b c = fromEdges [(0, 1, a), (1, 2, b), (2, 0, c)]
+
+{- | A bidirectional 3-node graph (slide-5 style): edges @0->1@, @1->2@, @2->0@ and
```
martijnbastiaan, line 31: References slide the reader can't access.

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+        )
+        frames
+ where
+  reaches measured safe q =
+    [ (i, j, safeW - (lam + qf j - qf i))
+    | (i, j, lam) <- edges measured
+    , Just safeW <- [weight i j safe]
+    ]
+   where
+    qf m = Map.findWithDefault 0 m q
```
martijnbastiaan, line 98: This looks like the helper function calculating `framesToInsert`. Can you deduplicate?

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+-- | Complete digraph over @0 .. k-1@ with the given per-edge weights.
+completeFrom :: Int -> [Integer] -> Graph Int Integer
+completeFrom k ws = fromEdges (zipWith (\(i, j) w -> (i, j, w)) (pairs k) ws)
+
+pairs :: Int -> [(Int, Int)]
+pairs k = [(i, j) | i <- [0 .. k - 1], j <- [0 .. k - 1], i /= j]
+
+prop_groom_reaches_safe :: Property
+prop_groom_reaches_safe = property $ do
+  k <- forAll $ Gen.int (Range.linear 2 5)
```
martijnbastiaan, line 68:
```suggestion
  nNodes <- forAll $ Gen.int (Range.linear 2 5)
```

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+  let ps = pairs k
+      n = length ps
+  lams <- forAll $ Gen.list (Range.singleton n) (Gen.integral (Range.linearFrom 0 (-50) 50))
+  eps <- forAll $ Gen.list (Range.singleton n) (Gen.integral (Range.linear 0 20))
+  let
+    measured = completeFrom k lams
+    safe = completeFrom k (zipWith (+) lams eps)
```
martijnbastiaan, line 75:
```suggestion
measured = fromWeightedPairs $ zip ps lams
safe = fromWeightedPairs $ zip ps (zipWith (+) lams eps)
```

`completeFrom` doesn't really make sense as a top-level function, there are way too many invariants your input has to conform to in order to use it sensibly.

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+
+pairs :: Int -> [(Int, Int)]
+pairs k = [(i, j) | i <- [0 .. k - 1], j <- [0 .. k - 1], i /= j]
+
+prop_groom_reaches_safe :: Property
+prop_groom_reaches_safe = property $ do
+  k <- forAll $ Gen.int (Range.linear 2 5)
+  let ps = pairs k
+      n = length ps
+  lams <- forAll $ Gen.list (Range.singleton n) (Gen.integral (Range.linearFrom 0 (-50) 50))
```
martijnbastiaan, line 71: I don't understand how this prevents generating graphs that have negative cycles. And consequently, I don't see how this test doesn't encounter the `Infeasible` arm. Could you explain?

For example, for `k=2` you get two pairs: `(0, 1)` and `(1, 0)`. Why would `lams` never be `[-1, -1]`?

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
martijnbastiaan: Also, could you bump the maximum number of nodes to a sillier number like 10? Or just pick a number that doesn't make it unreasonably slow.

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+  -- lambda == lambda^safe: feasible with all-zero frames (no relabeling needed).
+  k <- forAll $ Gen.int (Range.linear 2 5)
+  let ps = pairs k
+      n = length ps
+  lams <- forAll $ Gen.list (Range.singleton n) (Gen.integral (Range.linearFrom 0 (-50) 50))
+  let g = completeFrom k lams
+  case groomCorrection g g of
+    Infeasible c -> annotateShow c >> failure
+    Feasible{framesToInsert = frames} ->
+      assert (all (\(_, _, f) -> f == 0) frames)
```
martijnbastiaan, line 111: This test tests a subset of what I describe for the other test, so I think it can be removed.

bittide/tests/Tests/ClockControl/Ugn/Grooming.hs
```diff
@@ -0,0 +1,124 @@
...
+
+case_infeasible_when_changed_too_much :: Assertion
+case_infeasible_when_changed_too_much = do
```
martijnbastiaan, line 114: Slop!

bittide/src/Bittide/TimedReset.hs
```diff
@@ -0,0 +1,74 @@
...
+          "release_cycle"
+          "Local-counter value at which the output reset is released. While the\
+          \ local counter is at or below this value the reset stays asserted; once\
+          \ it exceeds it the reset deasserts and (being a monotonic counter) stays\
+          \ deasserted. The management unit writes this to choose when the\
+          \ application leaves reset (the UGN-grooming relabel). Defaults to\
+          \ maxBound, i.e. held until written."
+      )
+      (maxBound :: Unsigned 64)
+      -< (wbReleaseCycle, Fwd (pure Nothing))
```
martijnbastiaan, line 70: This write can't happen atomically on our CPU, so it might trigger early. This can happen if the local counter is >34 seconds.

bittide/src/Bittide/ClockControl/Ugn/Corrections.hs
```diff
@@ -0,0 +1,72 @@
...
+import Clash.Class.BitPackC (ByteOrder)
+
+import qualified Clash.Prelude as C
+
+{- | UGN grooming corrections peripheral. Holds @n@ signed corrections and a
+validity flag, all writable and readable over Wishbone. See the module header
+for the intended host/CPU handshake.
+-}
+correctionsWb ::
+  forall n dom addrW.
```
martijnbastiaan, line 56: Could this be moved into `CaptureUgn`? It already has "debug"/"test" registers.

bittide-instances/src/Bittide/Instances/Hitl/GenericDemo/Core.hs
```diff
@@ -59,6 +59,7 @@ import Protocols
 import Bittide.CaptureUgn (captureUgns, sendUgn)
 import Bittide.ClockControl (SpeedChange)
 import Bittide.ClockControl.CallistoSw (SwcccInternalBusses, callistoSwClockControlC)
```
martijnbastiaan, line 61: Could you have a look at the commit message? The general take is that it doesn't really convey information that's important.

bittide-instances/src/Bittide/Instances/Hitl/GenericDemo/Core.hs
```diff
@@ -144,6 +148,10 @@ type UserCoreCircuit userCoreBusses muRemBusWidth =
     Enable Bittide ->
     Signal Bittide (Unsigned 64) ->
     Signal Bittide (Maybe (BitVector 96)) ->
+    -- Application reset, released by the management unit's 'timedResetWb' at a
```
martijnbastiaan, line 151: Update documentation after switching to programmable mux.

bittide-instances/src/Bittide/Instances/Hitl/GenericDemo/Core.hs
```diff
@@ -211,16 +220,24 @@ managementUnit externalCounter maybeDna =
...
     _localCounter <- timeWb (Just externalCounter) -< timeBus
     (uartOut, _uartStatus) <-
       uartInterfaceWb d16 d16 uartBytes -< (uartBus, Fwd (pure Nothing))
     readDnaPortE2WbWorker maybeDna -< dnaBus
+    -- Host-written UGN grooming corrections, polled and applied by the MU CPU.
+    correctionsWb (SNat @LinkCount) -< correctionsBus
+    -- Application reset for the demo's user core: the management unit chooses the
+    -- local-counter cycle at which the application leaves reset (the UGN-grooming
+    -- relabel). The reset lives here, in the layer around the user core; the
+    -- application counter it gates lives in the user core itself.
```
martijnbastiaan, line 236: > the UGN-grooming relabel

The what now?

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+algorithm ("Bittide.ClockControl.Ugn.Grooming", on "Bittide.Graph.Weighted") to the HITL
+'UgnEdge' type and turns its result into a directly-applyable 'RelabelPlan'.
+
+Nodes are identified by their DNA-derived 'BitVector' 32 id (the @srcNode@ / @dstNode@
+fields of a 'UgnEdge'), so the graph is keyed directly on those ids. Edges are matched
+between the measured @λ@ and stored @λ^safe@ snapshots by their @(srcNode, dstNode)@ pair
+— unique because the demo topology is a complete graph (one link per ordered node pair).
+
+@Signed 64@ UGN values are widened to 'Integer' for the (overflow-prone) relaxation, and
+the resulting reset offsets / frame counts are narrowed back with a bounds check.
```
martijnbastiaan, line 16: He talked a lot, but said nothing.

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+import Bittide.Graph.Weighted (Graph)
+import Bittide.Instances.Hitl.Setup (FpgaCount, LinkCount, knownLinkConfigs)
+import Bittide.Instances.Hitl.Utils.Ugn (UgnEdge (..), indexToNodeId)
+
+import qualified Bittide.Graph.Weighted as G
+import qualified Clash.Prelude as CP
+import qualified Data.Map.Strict as Map
+
+{- | Build a weighted graph from UGN edges, keyed on the DNA-derived node ids with
+the (signed) UGN as the 'Integer' edge weight.
```
martijnbastiaan, line 40:
```suggestion
{- | Build a weighted graph from UGN edges, where nodes are represented
based on their DNA (@BitVector 32@).
```

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+-}
+ugnGraph :: [UgnEdge] -> Graph (BitVector 32) Integer
+ugnGraph es = G.fromEdges [(e.srcNode, e.dstNode, toInteger e.ugn) | e <- es]
```
martijnbastiaan, line 43: Can this function move to `Bittide.Instances.Hitl.Utils.Ugn.edgesToGraph`?

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+
+{- | Add a safety margin @ε@ to every edge: @λ^safe = λ + ε@. Mirrors the
+record-update style of 'Bittide.Instances.Hitl.Utils.Ugn.addLatencyEdge'.
+-}
+safeMargin :: Signed 64 -> [UgnEdge] -> [UgnEdge]
+safeMargin eps = map (\e -> e{ugn = e.ugn + eps})
+
```
martijnbastiaan, line 50: 1. The documentation isn't great. (_Lekker boeien_ that you use a record-update style that's similar to some other random function.)
2. `safeMargin` isn't a great name. Perhaps `addToUgn` / `addToUgns`.
3. It feels like this should be moved to `Ugn.hs`.

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+{- | Relabel UGN edges to their minimal non-negative form: the Bellman-Ford reduced
+costs @λ_{i->j} + q_i - q_j@ (potentials @q@ from the super-source), which are all
+@>= 0@ and leave every round-trip / cycle sum unchanged.
+
+This is a pure gauge change (a relabeling), so it represents the /same/ physical
+system, but in a gauge where the UGNs are small and non-negative. Storing
+@λ^safe@ in this gauge keeps the application's per-link UGNs (and hence its fixed
+schedule) inside the working range a depth-bounded mux/elastic-buffer needs, while
+still honouring the true /asymmetric/ cycle constraints (unlike a symmetric
+midpoint, which is only a valid relabeling on an acyclic graph).
```
martijnbastiaan, line 60:
```suggestion
{- | Relabel UGN edges to their minimal non-negative form.
```

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+
+If the edges are not physically allowed (a negative cycle), they are returned
+unchanged.
```
martijnbastiaan, line 63: Make it return an `Either String [UgnEdge]`.

```console
martijn@volthe:~$ python3 -c "import this"  | head -n 12 | tail -n 1
Errors should never pass silently.
```

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+      | e <- es
+      ]
+     where
+      qOf n = Map.findWithDefault 0 n q
+
+{- | The result of grooming a measured boot onto a target: per-node reset offsets and
+per-node, per-link frame corrections, both FPGA-/link-indexed and ready to apply.
+-}
+data RelabelPlan = RelabelPlan
+  { resetOffsets :: Vec FpgaCount (Signed 64)
```
martijnbastiaan, line 80: Why isn't this just an absolute clock cycle? What's this all about a shared base all of a sudden?

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+  {- ^ Per-node offset to add to the shared base to get its reset-release cycle. This
+  realises the relabel @q@: every node's application counter leaves reset in the
+  relabeled domain. Gauged so node 0's offset is 0.
```
martijnbastiaan, line 83: This _seems_ to relate to the canonicalization step, but that's not clear to me.

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+  -}
+  , corrections :: Vec FpgaCount (Vec LinkCount (Signed 64))
```
martijnbastiaan, line 85: This is derived from `framesToInsert`. The record that houses that field also has a field called `corrrection`, which -- confusingly -- doesn't correspond to this `corrections`.

martijnbastiaan, line 85: Why is this a `Signed 64`? None of the underlying algorithms can produce negative numbers.

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+  {- ^ Per-node, per-link frame corrections: the frames the management unit inserts
```
martijnbastiaan, line 86: Overly complex.

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+    'knownLinkConfigs' (the same link map 'Bittide.Instances.Hitl.Utils.Ugn.ugnEdges'
+    uses).
+
+When @λ^safe@ is stored in a small non-negative gauge (see 'canonicalizeUgn'), the relabel
+absorbs this boot's counter offsets so the groomed app-frame UGNs land back on @λ^safe@ —
+small, non-negative, and identical every run, which is what makes a single fixed schedule
+reusable.
+
+Edges are matched by @(srcNode, dstNode)@; pass the full measured + target graphs to
+groom every link.
```
martijnbastiaan, line 116: Okay...

1. Don't mention Bellman-Ford
2. Don't mention the slides
3. Don't explain the fields of a different data type.
4. It's not clear why the word "gauge" keeps cropping up everywhere.
5. Documentation is not self-contained, points to a lot of concepts which the reader is just supposed to find on their own.
6. The last sentence doesn't add anything.

In general it's a lot of prose which only achieves confusion. Just call it what it is: a call to `groomCorrection` and a group by FPGA index.

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/UserCore.hs
```diff
@@ -31,10 +31,16 @@ ringBufferDepth :: SNat RingBufferDepth
...
-mkUserCore bitClk bitRst bitEna localCounter maybeDna =
+mkUserCore bitClk bitRst bitEna localCounter maybeDna appReset =
   circuit $ \(userCoreBusses, Fwd rxs2Raw, handshakeOut) -> do
     [muProgrammableMuxBus, peConfigBus] <- idC -< userCoreBusses
 
+    -- The application counter is the relabeled time base: it starts counting when
+    -- the management unit releases 'appReset' (see 'timedResetWb' in the MU), so the
+    -- per-node release timing realizes the UGN-grooming relabel. The reset itself
+    -- lives outside the user core (in the layer with the management unit).
+    let appCounter = register bitClk appReset bitEna 0 (appCounter + 1)
```
martijnbastiaan, line 42: `programmableMux` should operate on `localCounter`.

Related-but-not-really: https://github.com/bittide/bittide-hardware/issues/1335

bittide-instances/tests/Tests/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,195 @@
...
+
+import qualified Data.Map.Strict as Map
+import qualified Hedgehog as H
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+-- * Helpers
+
+mkEdge :: BitVector 32 -> Index LinkCount -> BitVector 32 -> Index LinkCount -> Signed 64 -> UgnEdge
+mkEdge = UgnEdge
```
martijnbastiaan, line 36: This doesn't do anything, please remove. Also, it probably makes sense to construct with record field names (`Foo{a = ..., b = ...}`).

bittide-instances/tests/Tests/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,195 @@
...
+  map (.ugn) (safeMargin 5 es) @?= [15, 1]
+
+case_ugnGraph_roundtrip :: Assertion
+case_ugnGraph_roundtrip = do
+  let
+    es = [mkEdge 0 0 1 0 7, mkEdge 1 0 0 0 (-3)]
+    g = ugnGraph es
+  sort (edges g) @?= sort [(0, 1, 7), (1, 0, -3)]
+  weight 0 1 g @?= Just 7
+  weight 1 0 g @?= Just (-3)
```
martijnbastiaan, line 52: This test tests something, but it feels pretty much superfluous.

bittide-instances/tests/Tests/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,195 @@
...
+
+-- * Property: grooming restores a freshly-booted mesh to a prior state
+
+-- | The four nodes of the modelled mesh, identified by their (DNA-derived) node ids.
```
martijnbastiaan, line 56:
```suggestion
-- | The four nodes of the modelled mesh
```

bittide-instances/tests/Tests/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,195 @@
...
+fixed reference. Asymmetric on purpose (forward/return links differ) to exercise the
+directed-graph handling.
+-}
+referenceFibers :: Map (BitVector 32, BitVector 32) Integer
+referenceFibers =
+  Map.fromList $
+    zip
+      nodePairs
+      -- (0,1)(0,2)(0,3) (1,0)(1,2)(1,3) (2,0)(2,1)(2,3) (3,0)(3,1)(3,2)
+      [120, 340, 275, 118, 260, 410, 345, 262, 190, 280, 405, 188]
```
martijnbastiaan, line 75:
```suggestion
      -- 0-1  0-2  0-3    1-0  1-2  1-3    2-0  2-1  2-3    3-0  3-1  3-2
      [  120, 340, 275,   118, 260, 410,   345, 262, 190,   280, 405, 188]
```

bittide-instances/tests/Tests/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,195 @@
...
+nodePairs :: [(BitVector 32, BitVector 32)]
+nodePairs = [(a, b) | a <- nodeIds, b <- nodeIds, a /= b]
+
+{- | Reference fiber/transceiver latency per directed link, in cycles. Fixed hardware
+characteristic of the topology — hard-coded here just as the implementation stores a
+fixed reference. Asymmetric on purpose (forward/return links differ) to exercise the
+directed-graph handling.
+-}
+referenceFibers :: Map (BitVector 32, BitVector 32) Integer
+referenceFibers =
```
martijnbastiaan, line 70: The comment and name make me think these are based on "real" hardware, but that's not the case. It's okay to have it not be based on real hardware btw, it's really a doc issue.

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+
+Nodes are identified by their DNA-derived 'BitVector' 32 id (the @srcNode@ / @dstNode@
+fields of a 'UgnEdge'), so the graph is keyed directly on those ids. Edges are matched
+between the measured @λ@ and stored @λ^safe@ snapshots by their @(srcNode, dstNode)@ pair
+— unique because the demo topology is a complete graph (one link per ordered node pair).
+
+@Signed 64@ UGN values are widened to 'Integer' for the (overflow-prone) relaxation, and
+the resulting reset offsets / frame counts are narrowed back with a bounds check.
+-}
+module Bittide.Instances.Hitl.Utils.UgnGrooming (
```
martijnbastiaan, line 18: Alright. After reading through this whole commit, I feel that `UgnGrooming.hs` should move to `bittide`. The same goes for `Utils/Ugn.hs`. The only reason they now exist in `bittide-instances`, is a dependency on `LinkCount` and `FpgaCount`, which can just as well be solved by polymorphism.

There is also a bunch of duplication, though I feel that that will solve itself if you just move everything to `bittide`. This is especially obvious in the tests for this module, as they largely correspond to the ones for `Bittide.ClockControl.Ugn.Grooming`.

bittide-instances/src/Bittide/Instances/Hitl/Utils/UgnGrooming.hs
```diff
@@ -0,0 +1,140 @@
...
+@λ^safe@ in this gauge keeps the application's per-link UGNs (and hence its fixed
+schedule) inside the working range a depth-bounded mux/elastic-buffer needs, while
+still honouring the true /asymmetric/ cycle constraints (unlike a symmetric
+midpoint, which is only a valid relabeling on an acyclic graph).
+
+If the edges are not physically allowed (a negative cycle), they are returned
+unchanged.
+-}
+canonicalizeUgn :: [UgnEdge] -> [UgnEdge]
+canonicalizeUgn es =
```
martijnbastiaan, line 66: It's not clear to me whether this is here because of your work on Manticore, or whether it's fundamentally needed for the demo in this PR. Could you clarify?

bittide-instances/src/Bittide/Instances/Hitl/Utils/Gdb.hs
```diff
@@ -26,3 +35,75 @@ initGdb hitlDir binName gdb tapInfo (hwT, _d) = do
...
+-- across demos (the memory map is a parameter). All require the CPU to be halted.
+
+-- | Read the current local-counter value from a device's @Timer@ peripheral.
+readCurrentTime :: (HasCallStack) => MemoryMap -> (HwTarget, DeviceInfo) -> Gdb -> IO (Unsigned 64)
+readCurrentTime mm (_, d) gdb = do
+  putStrLn $ "Getting current time from device " <> d.deviceId
+  commandAddress <- expectRight $ getPathAddress mm ["0", "Timer", "command"]
+  scratchAddress <- expectRight $ getPathAddress mm ["0", "Timer", "scratchpad"]
+  Gdb.writeLe gdb commandAddress Capture
+  Gdb.readLe gdb scratchAddress
```
martijnbastiaan, line 52: This function already exists in `WireDemo`. Could you move them to their own module per memory mapped device? Also see issue https://github.com/bittide/bittide-hardware/issues/1336.

bittide-instances/src/Bittide/Instances/Hitl/Utils/Gdb.hs
```diff
@@ -6,11 +6,20 @@ module Bittide.Instances.Hitl.Utils.Gdb where
 
 import Prelude
 
+import Clash.Prelude (Signed, Unsigned, Vec, checkedFromIntegral, toList)
+
 import Bittide.Hitl (DeviceInfo (..))
+import Bittide.Instances.Hitl.Setup (LinkCount)
```
martijnbastiaan, line 12: See other comment in this module: this module shouldn't contain device-specific functions.

bittide-instances/src/Bittide/Instances/Hitl/Utils/Ugn.hs
```diff
@@ -211,6 +211,31 @@ timingOracleToUgnEdge oracle =
     , ugn = bitCoerce oracle.dstCounter - bitCoerce oracle.srcCounter
     }
 
+{- | Turn a per-node, per-link table of UGN values (@λ@) into 'UgnEdge's keyed by DNA
+node id. The outer vector is FPGA-indexed (the capturing / destination node); each
+inner vector is link-indexed (matching 'knownLinkConfigs').
+-}
+ugnEdges :: CP.Vec FpgaCount (CP.Vec LinkCount (Signed 64)) -> [UgnEdge]
+ugnEdges lams =
```
martijnbastiaan, line 219: This should probably take the map as an argument and move to `bittide`.

bittide-instances/src/Bittide/Instances/Hitl/Utils/Ugn.hs
```diff
@@ -211,6 +211,31 @@ timingOracleToUgnEdge oracle =
...
+      , dstPort = port
+      , ugn = lam
+      }
+
+{- | As 'ugnEdges', but from raw @(localCounter, remoteCounter)@ captures: the UGN is
+their signed difference @local - remote@.
+-}
+hardwareUgnEdges ::
+  CP.Vec FpgaCount (CP.Vec LinkCount (Unsigned 64, Unsigned 64)) -> [UgnEdge]
+hardwareUgnEdges = ugnEdges . CP.map (CP.map (\(l, r) -> bitCoerce l - bitCoerce r))
```
martijnbastiaan, line 237: Could you use `numConvert`?

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -47,11 +59,16 @@ import qualified Bittide.Instances.Hitl.Utils.OpenOcd as Ocd
...
 import qualified System.Timeout.Extra as T
 
-type StartDelay = 10 -- seconds
+{- | Seconds of headroom between reading the current time and the reset-release /
+schedule point. This must comfortably exceed the wall-clock time it takes to
+write all per-node timestamps + configs over GDB (~10s for 8 nodes), otherwise
+a node's reset can release before it is fully configured (see the post-write
+check in 'driver').
+-}
+type StartDelay = 30 -- seconds
```
martijnbastiaan, line 71:
```suggestion
type StartDelaySeconds = 30
```

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -62,6 +79,117 @@ to the PE-to-PE delay.
...
+appScheduleBase :: Unsigned 64
+appScheduleBase = 1000
+
+{- | Frames of safety margin added to the golden UGNs to form @λ^safe@ (@λ^safe =
+golden + ε@). Each boot is groomed up to @λ^safe@, so @ε@ is the number of frames the
+elastic buffer inserts on top of matching the golden latency; it must cover the small
+boot-to-boot latency drift yet stay within the buffer's safe range. The elastic buffer
+depth is 'Bittide.Instances.Hitl.GenericDemo.Core.FifoSize' bits (2^6 = 64 entries,
+occupancy range ±32), so margin 5 plus per-link drift stays comfortably within range.
+-}
```
martijnbastiaan, line 95:
```suggestion
{- | Frames of safety margin added to the golden UGNs to form @λ^safe@ (@λ^safe =
golden + ε@). Each boot is groomed up to @λ^safe@, so @ε@ is the number of frames the
elastic buffer inserts on top of matching the golden latency.
-}
```

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -62,6 +79,117 @@ to the PE-to-PE delay.
...
+{- | Golden UGN graph: a full set of per-link UGNs (@λ = local - remote@) captured from a
+passing CI boot of this rig, stored as a 'UgnEdge' list keyed by DNA-derived node id and
+port. This is the stored reference every boot is groomed onto.
+
+The UGNs carry their boot-time counter offsets; 'lambdaSafe' relabels them into a
+small, non-negative gauge before use.
+-}
+goldenUgns :: [UgnEdge]
+{- FOURMOLU_DISABLE -}
+goldenUgns = canonicalizeUgn
```
martijnbastiaan, line 108: Not really a fan of this approach, given that bittide is still under development. Can it be replaced by a (nightly) test that just does two wire demos -- where the second one reuses the schedule from the first one?

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -337,20 +414,110 @@ driver testName targets = do
...
-          -- Calculate schedule and write configurations
+          -- Relabel this boot onto the stored target latencies, using the grooming
+          -- utilities ("Bittide.Instances.Hitl.Utils.UgnGrooming"): build UGN
+          -- edges from the measured counters, groom them (Bellman-Ford) onto
+          -- λ^safe = the target latency on both directions of every chain hop, and turn
+          -- the result into per-node reset releases (which remove the boot-time counter
+          -- offsets) and per-link frame corrections (which raise each link's latency to
+          -- the target). The application then runs one FIXED schedule built from the
+          -- target latencies -- identical every run.
           currentTime <-
```
martijnbastiaan, line 425:
```suggestion
          -- Relabel this boot onto the stored target latencies, using the grooming
          -- utilities ("Bittide.Instances.Hitl.Utils.UgnGrooming").
          currentTime <-
```

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -337,20 +414,110 @@ driver testName targets = do
...
             liftIO $ readCurrentTime MemoryMaps.managementUnit (L.head targets) (L.head managementUnitGdbs)
           let
-            startOffset = currentTime + natToNum @(PeriodToCycles GthTx (Seconds StartDelay))
-            schedule = generateSchedule fpgaSetup ugnPairsTableV startOffset
+            -- Shared reference instant for the relabel: node 0's current time plus a
+            -- large constant. 'StartDelay' must exceed the GDB bookkeeping time, else a
+            -- node's reset could release before it is fully configured.
```
martijnbastiaan, line 430:
```suggestion
            -- large constant.
```

This is already documented at `StartDelay`.

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -337,20 +414,110 @@ driver testName targets = do
...
+            sharedBase = currentTime + natToNum @(PeriodToCycles GthTx (Seconds StartDelay))
+
+            -- This boot's measured UGN edges for the FULL graph (every node, every
+            -- link) -- the grooming is independent of the application and adjusts all
+            -- elastic buffers.
+            measuredEdges = hardwareUgnEdges ugnPairsTableV
```
martijnbastiaan, line 436:
```suggestion
            measuredEdges = hardwareUgnEdges ugnPairsTableV
```

bittide-instances/src/Bittide/Instances/Hitl/WireDemo/Driver.hs
```diff
@@ -337,20 +414,110 @@ driver testName targets = do
...
+
+          -- Groom this boot's measured UGNs onto the stored 'lambdaSafe' (Bellman-Ford):
+          -- per-node reset offsets + per-link frame corrections for every edge.
+          -- 'appSchedule' is fixed and derived from the same 'lambdaSafe'.
+          relabelPlan <- case computeRelabel measuredEdges lambdaSafe of
```
martijnbastiaan, line 441:
```suggestion
```

This is already well-documented in the functions

martijnbastiaan (changes requested):
We've run out of time on this one unfortunately. But here's all the feedback we've collected so far.

PR #1323, opened 2026-06-04
Replace picocom with a native Haskell serial port

martijnbastiaan:
_What (what did you do)_
Replaced `picocom` with a native Haskell implementation.

_Why (context, issues, etc.)_
We've had troubles with picocom not cleanly exiting when playing around with multiple tests in a single driver, we apply workarounds, and dealing with subprocesses in Haskell (unix) is a pain anyway.

~~Is self-hosting this worth it? To me it seems like this implementation is simpler than what we had previously and it removes worries I have over proper subprocess management.~~ No. `sport` exists.

I let Claude implement this while I was working on client negotiations.

# TODO

- [ ] Write (regression) test
- [ ] Link to existing issue
- [x] Remove picocom specific workarounds (weird bytes we currently filter out)


bittide-instances/src/Bittide/Instances/Hitl/Utils/Picocom.hs
```diff
@@ -2,267 +2,146 @@
...
+Two styles are offered:
+
+  * A raw 'SerialHandle' ('start', 'withSerial') for drivers that read and write
+    the port directly.
+  * A 'Chan' 'ByteString' of received lines ('initPicocom',
+    'startWithLoggingChan') for drivers that only consume output. In this style,
+    every received line is also appended to a log file - this replaces
+    @picocom@'s @tee@ and is the primary HITL debug artifact.
+-}
 module Bittide.Instances.Hitl.Utils.Picocom where
```
martijnbastiaan: Rename to `Utils.Serial`.

lmbollen (approved):
I see the benefits.

Issue #1118, opened 2025-12-08
Remove hardware for clock board programming


hiddemoll:
In #1100 we moved programming the external clock board to software for all demo's. To speed up compilation we should remove the hardware SPI programming from the repo, including the slow template haskell for parsing clock configurations. To do this we should first add a boot CPU to `fincFdecTests`, which is the last remaining test which uses hardware SPI.

PR #1327, opened 2026-06-12
Move `fincFdecTests` to software

hiddemoll:
_What (what did you do)_
Move `fincFdecTests` to software.

_Why (context, issues, etc.)_
Prerequisite for #1118

Requesting @martijnbastiaan for everything except the Rust, and requesting @rslawson specifically for the Rust.

I'm not super happy with the rust firmware (AI generated). It has a lot of very similar functions, but the hardware and software versions differ ever so slightly; HW only needs to set the `SpeedChange` once, but also set it to `NoChange` at the end, while SW has to keep doing SPI write operations to the FINC/FDEC register. @rslawson are you happy with it as is? If not, do you have some tips on how to improve it?

Only for the rust code.

# TODO

- [x] Write (regression) test
- ~[ ] Update documentation, including `docs/`~
- [x] Link to existing issue -> (part of #1118)
- [x] Run test on CI -> see [this run ](https://github.com/bittide/bittide-hardware/actions/runs/28084066217)


bittide-instances/src/Bittide/Instances/Hitl/GenericDemo/Core.hs
```diff
@@ -89,6 +89,7 @@ import Protocols.MemoryMap (Mm)
 import Protocols.Wishbone.Extra (delayWishbone)
 import VexRiscv (DumpVcd (..), Jtag)
 
+import Bittide.ClockControl (SpeedChange (NoChange))
```
martijnbastiaan: One line up please

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
```diff
@@ -8,205 +8,190 @@ FINC and FDEC pins.
...
-  goFincFdec FInc n
-    | n > threshold = (FDec, (NoChange, Busy))
-    | n < -threshold = (FInc, (NoChange, Fail))
-    | otherwise = (FInc, (SpeedUp, Busy))
-  goFincFdec FDec n
-    | n > (3 * threshold) = (FDec, (NoChange, Fail))
-    | n < 0 = (FDec, (NoChange, Success))
-    | otherwise = (FDec, (SlowDown, Busy))
-  goFincFdec s _ = (s, (NoChange, Fail)) -- Illegal state
+  -- TODO: Check if depth can be smaller
```
martijnbastiaan: 8 KiB is fiiiiiine

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
```diff
@@ -8,205 +8,190 @@ FINC and FDEC pins.
...
+
+  Fwd _localCounter <- withClockResetEnable freeClk freeRst enableGen $ timeWb Nothing -< timeBus
+  (uartTx, _uartStatus) <-
+    withClockResetEnable freeClk freeRst enableGen
+      $ uartInterfaceWb d16 d16
+      $ uartDf baud
+      -< (uartBus, Fwd (pure low))
+  (Fwd spiDone, spiOut) <-
+    withClockResetEnable freeClk freeRst enableGen
+      $ si539xSpiWb (SNat @(Microseconds 10))
```
martijnbastiaan, line 122: What is the 10 microseconds again? Could you comment on why it is 10?

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
```diff
@@ -8,205 +8,190 @@ FINC and FDEC pins.
...
+    toSignals
+      (withXilinx $ withLittleEndian $ fincFdecPe freeClk testRst boardClk)
+      (((), jtagIn), ((), spiS2M, ()))
 
-  -- For debugging, add to separate VIO?
-  -- clkStable1 = unsafeToActiveLow clkStableRst
-  -- testRstBool = unsafeToActiveHigh testRst
-  -- (fInc, fDec) = fIncDec
-  -- (spiBusy, spiState, siClkLocked, counterActive, counter) = debugSignals
+  fincFdec = unbundle $ speedChangeToFincFdec freeClk testRst speedChange
```
martijnbastiaan, line 185: It's fine for now, but I wonder if we should make software responsible for this. Perhaps open an issue?

martijnbastiaan (approved):

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
hiddemoll: Should we document that here? The `si539xSpiWb` function already documents that input: `Minimum period of the SPI clock frequency for the SPI clock divider.`. It is used internally to calculate the clock divider for `spiMaster`.

As for why it is 10, I don't really know. All other uses of our specialized SPI core use 10 microseconds, but now that I look at it the previous implementation of `fincFdecTests` actually used 1 microsecond. I'll look into it.

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
hiddemoll: I had already reduced to to the smallest size, but forgot to remove the TODO.

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
hiddemoll: I don't agree. Software is too slow for these durations, so the pulses would be much longer and thus the frequency of FINC/FDEC pulses would be lower.

martijnbastiaan: Ok!

bittide-instances/src/Bittide/Instances/Hitl/Driver/FincFdec.hs
```diff
@@ -0,0 +1,94 @@
...
+
+      let picocomStarts = liftIO <$> L.zipWith (initPicocom hitlDir) targets [0 ..]
+      brackets picocomStarts (liftIO . snd) $ \(L.map fst -> picocoms) -> do
+        liftIO $ mapConcurrently_ Gdb.continue gdbs
+
+        liftIO
+          $ T.tryWithTimeout T.PrintActionTime "Waiting for test success" 60_000_000
+          $ forConcurrently_ picocoms
+          $ \pico ->
+            waitForLine pico "All tests passed"
```
martijnbastiaan: It's not a big issue, but this will just block for 60 seconds even if the test already indicated it failed. Not for this PR, but maybe good to make an issue out of it.

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
```diff
@@ -8,205 +8,189 @@ FINC and FDEC pins.
...
-{- | Counter threshold after which a test is considered passed/failed. In theory
-clocks can diverge at +-20 kHz (at 200 MHz), which gives the tests 500 ms to
-adjust their clocks - which should be plenty.
+type Baud = 921_600
+
+baud :: SNat Baud
+baud = SNat
+
+{- | Memory mapped component to control the FINC and FDEC pins. Continuously applies the
+requested speed change as fast as the clock board allows.
```
martijnbastiaan: Wait, does it? The "as fast as the clock board allows" is handled by an external component right?

bittide-instances/src/Bittide/Instances/Hitl/Driver/FincFdec.hs
hiddemoll: To be fair, this only really goes for 2 HITL tests: `si539xConfigTest` and `fincFdecTests`. None of the other tests explicitely write something like "Some tests failed". So I think the ~30 seconds you may save when one of these fails early is not worth the engineering effort.

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
hiddemoll: Good catch, I indeed moved that out of the function. ~But since the firmware for this test still depends on that behaviour it should probably be moved back into this function. Do you agree? Or shall I just remove the second sentence from the function documention?~

I just removed the second sentence.

bittide-instances/src/Bittide/Instances/Hitl/Driver/FincFdec.hs
martijnbastiaan: Hmmyeah, well similar things happen elsewhere: https://github.com/bittide/bittide-hardware/pull/1214#discussion_r3402338341. Testing with software in the loop is ad-hoc throughout the repo.

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
martijnbastiaan: Makes sense!

bittide-instances/src/Bittide/Instances/Hitl/Driver/FincFdec.hs
martijnbastiaan: Lucas has reported this as an issue: https://github.com/bittide/bittide-hardware/issues/1331.

firmware-binaries/hitl-tests/finc-fdec/src/main.rs
rslawson, line 1: I'd like test files like these not to return `bool` as the success measure, but rather have
```rust
use core::panic::Location;

enum TestFailKind {
    SpiWriteNotConfirmed(ConfigEntry, u8),
    CounterTest { limit: i32, found: i32 },
    // any other failure kinds
}

struct TestFail {
    loc: &'static Location<'static>,
    kind: TestFailKind,
}

impl TestFail {
    #[track_caller]
    fn here(kind: TestFailKind) -> Self {
        TestFail {
            loc: Location::caller(),
            kind,
        }
    }

    fn with_loc(loc: &'static Location<'static>, kind: TestFailKind) -> Self {
        TestFail { loc, kind }
    }
}

type TestResult = Result<(), TestFail>;

trait UpdateCaller {
    fn update_caller(self) -> Self;
}

impl UpdateCaller for TestResult {
    #[track_caller]
    fn update_caller(self) -> Self {
        match self {
            Ok(_) => self,
            Err(err) => Err(TestFail {
                loc: Location::caller(),
                kind: err.kind,
            }),
        }
    }
}
```
And then tests should return a `TestResult<'static>` or so instead of a `bool`. This should also allow for tests to be written in terms of others more easily, without the `if other_test() { return true; }` stuff:
```rust
fn counter_test<Pass, Fail>(
    dc: &DomainDiffCounters,
    (mut pass_test, pass): (Pass, i32),
    (mut fail_test, fail): (Fail, i32),
) -> TestResult
where
    Pass: FnMut(&i32, &i32) -> bool,
    Fail: FnMut(&i32, &i32) -> bool,
{
    loop {
        let count = read_counter(dc);
        if pass_test(&count, &pass) {
            break Ok(());
        }
        if fail_test(&count, &fail) {
            break Err(TestFail::here(TestFailKind::CounterTest {
                limit: fail,
                found: count,
            }));
        }
    }
}

fn do_fdec_hw(sc: &HardwareSpeedChange, dc: &DomainDiffCounters) -> TestResult {
    sc.set_speed_change(SpeedChange::SlowDown);
    let result = counter_test(dc, (i32::lt, -THRESHOLD), (i32::gt, THRESHOLD)).update_caller();
    sc.set_speed_change(SpeedChange::NoChange);
    result
}

fn do_finc_hw(sc: &HardwareSpeedChange, dc: &DomainDiffCounters) -> TestResult {
    sc.set_speed_change(SpeedChange::SpeedUp);
    let result = counter_test(dc, (i32::gt, THRESHOLD), (i32::lt, -THRESHOLD)).update_caller();
    sc.set_speed_change(SpeedChange::NoChange);
    result
}  

// code the rest of the owl
```
I don't think this is a necessary change for this PR, but I would like to note this would be a nice change to make.

firmware-binaries/hitl-tests/finc-fdec/src/main.rs
```diff
@@ -0,0 +1,339 @@
...
+///
+/// Returns whether the test passed.
+fn run_test(
+    name: &str,
+    si539x_spi: &Si539xSpi,
+    timer: &Timer,
+    uart: &mut Uart,
+    dc: &DomainDiffCounters,
+    test_body: &dyn Fn(&DomainDiffCounters) -> bool,
+) -> bool {
```
rslawson: Unnecessary `&dyn`. Make this
```rust
fn run_test<Test: FnOnce(&DomainDiffCounters) -> bool>(
    name: &str,
    si539x_spi: &Si539xSpi,
    timer: &Timer,
    uart: &mut Uart,
    dc: &DomainDiffCounters,
    test_body: Test,
) -> bool {
    // ...
}

// or this, if you prefer explicit `where`s:

fn run_test<Test>(
    name: &str,
    si539x_spi: &Si539xSpi,
    timer: &Timer,
    uart: &mut Uart,
    dc: &DomainDiffCounters,
    test_body: Test,
) -> bool
where
    Test: FnOnce(&DomainDiffCounters) -> bool,
{
    // ...
}
```
And then in each of the `run_test` invocations below turn `&|dc| ...` into `|dc| ...`.

firmware-binaries/hitl-tests/finc-fdec/src/main.rs
```diff
@@ -0,0 +1,339 @@
...
+    let mut all_passed = true;
+
+    all_passed &= run_test(
+        "FDec (hardware)",
+        &si539x_spi,
+        &timer,
+        &mut uart,
+        &dc,
+        &|dc| do_fdec_hw(&sc, dc),
+    );
```
rslawson: Now I know I'm going to get a sigh for suggesting this, but this extremely repetitive structure is perfect for a macro. Something like
```rust
macro_rules! run_tests {
    (
        spi: $spi:ident,
        timer: $timer:ident,
        uart: $uart:ident,
        dc: $dc:ident,
        tests: [
            $([name: $name:literal, test: $test:expr]),+
            $(,)?
        ],
    ) => {
        $(
            run_test(
                $name,
                &$spi,
                &timer,
                &mut uart,
                &dc,
                $test,
            )
        )&&+
    };
}

fn main() -> ! {
    let si539x_spi = INSTANCES.si539x_spi;
    let timer = INSTANCES.timer;
    let mut uart = INSTANCES.uart;
    let dc = INSTANCES.domain_diff_counters;
    let sc = INSTANCES.hardware_speed_change;

    let all_passed = run_tests!(
        spi: si539x_spi, 
        timer: timer,
        uart: uart,
        dc: dc,
        sc: sc,
        tests: [
            [name: "FDec (hardware)", test: |dc| do_fdec_hw(&sc, dc)],
            [name: "FInc (hardware)", test: |dc| do_finc_hw(&sc, dc)],
            // ...
        ],
    );

    if all_passed {
        uwriteln!(uart, "All tests passed").unwrap();
    } else {
        uwriteln!(uart, "Some tests failed").unwrap();
    }

    loop {
        continue;
    }
}

rslawson (changes requested):
Only change I actually would like done for this PR is the one about the unnecessary `&dyn`, the other ones can be left until a later date.

bittide-instances/src/Bittide/Instances/Hitl/FincFdec.hs
hiddemoll: I fixed this by adding a retry mechanism in https://github.com/bittide/bittide-hardware/pull/1344. 
In this PR I added `write_configuration_with_retry` which I use for `fincFdecTests`.

firmware-binaries/hitl-tests/finc-fdec/src/main.rs
hiddemoll: I did use your suggestion as a baseline, but it was overkill for my purposes. For now I've added this change in a separate commit: "Generalize counter tests". I intend to squash it if it passes your second round of review.

firmware-binaries/hitl-tests/finc-fdec/src/main.rs
hiddemoll: IMO this is actually a good reason to use a macro, but I'd like a second opinion aside from Mr. Macro.

martijnbastiaan: Sure, makes sense

rslawson (approved):

PR #1333, opened 2026-06-16
AXI stream self test plus sidequests

rslawson:
_What (what did you do)_
In list form:
- Updated `clash-protocols-memmap`
- Moved the DUT for the AXI self test to `Bittide.Instances.Tests.Axi` so that it can produce a memmap that can be generated in `Bittide.Instances.MemoryMaps`
- Defined wrapper traits for AXI transmit/receive types and made implementations for them
- Rewrote the relevant firmware binaries/library code to use the new AXI code

_Why (context, issues, etc.)_
Issue #1254 

This is still marked as WIP because the AXI self test is failing on `send_multiple_packets_receive_all`. As far as I can tell, this is because at some point the packet length is being read as 129 instead of the maximum expected 128, and I don't know what's causing that. That needs to be chased down before this can be reviewed properly.

None.

# TODO

- [x] Link to existing issue


martijnbastiaan (approved):
LGTM, but I'd like @hydrolarus to take a look at the Rust code :).

firmware-binaries/sim-tests/axi_stream_self_test/src/main.rs
```diff
@@ -7,30 +7,48 @@
...
+    use bittide_hal::shared_devices::Uart;
+    use core::panic::Location;
+    use ufmt::uWrite;
+    fn write_loc_to_uart(uart: &mut Uart, loc: &Location<'_>) {
+        let file = loc.file();
+        let line = loc.line();
+        let col = loc.column();
+        uart.write_str(file).unwrap();
+        ufmt::uwrite!(uart, ":{}:{}", line, col).unwrap();
+    }
```
hydrolarus: does it make sense to create a function for those 5 lines when it's just used once?

firmware-support/bittide-hal/src/manual_additions/axi/self_test.rs
```diff
@@ -93,44 +89,60 @@ fn read_rx_status<const BUF_SIZE: usize>(
     false
 }
 
-fn clear_rx_status<const BUF_SIZE: usize>(
-    _tx: AxiTx,
-    rx: AxiRx<BUF_SIZE>,
+fn clear_rx_status<Tx: AxiTx, Rx: AxiRx>(
+    _tx: &Tx,
+    rx: &Rx,
```
hydrolarus, line 94: Maybe there should be documentation what exactly the returned bool means?

hydrolarus (commented):
Just style/doc remarks, looking good otherwise!

firmware-binaries/sim-tests/axi_stream_self_test/src/main.rs
rslawson: No, I just misread the docs for `Location::caller` and thought I might need to set up a recursive thing, and ended up forgetting to remove this before pushing.

firmware-support/bittide-hal/src/manual_additions/axi/self_test.rs
rslawson: Probably! This is something that was in the tests to begin with before this PR and I'd like to do a rewrite so that it doesn't return these `bool`s as a success measure (since they can't be `?`'d like `Option<_>` or `Result<_, _>`).

hydrolarus (approved):
Looking good from the Rust side!

